Issue #9, opened 2018-07-10
Support non-semver releases


toolmantim:
Currently if you try to use Release Drafter on a repo that has non-semver compatible release names, it throws `Error: Invalid argument not valid semver` when [compare-versions is used](https://github.com/toolmantim/release-drafter/blob/master/lib/releases.js#L11).

We should perhaps just compare the dates of the two releases in this case, and assume the more recent release is greater?

toolmantim:
@balloob this is why it's currently not working on https://github.com/home-assistant/home-assistant-polymer/releases

balloob:
Maybe it could be an option ? `semver: false`. I can see how for projects with semver that have multiple active releases (like Node, example `6.X` and `8.X`), just picking the last one will not work.

toolmantim:
Ah, multiple release tracks might be something else altogether though, but still a neat feature. For Node, right now it works but it'd only track the v10 releases, because it just asks "what's merged to master since the release with the highest semver version number?"

I'll get non-semver releases working to start. Multiple release paths can come later when people need.

toolmantim:
Fixed in #15

Issue #8, opened 2018-07-10
Squash & Merge PRs


balloob:
I always use squash & merge when merging PRs. Just realized that this doesn't work with release-drafter. My change log looks like this:

![image](https://user-images.githubusercontent.com/1444314/42502626-e28d249e-8436-11e8-8a95-69268f24d38a.png)

Looking at the code, merge PRs are matched on "Merge pull request #" https://github.com/toolmantim/release-drafter/blob/e24d69800ca69614e821cf642097dd82e5206cd3/lib/commits.js#L23

Could this be updated to also match commits generated by squash & merge?

For example, `(\(#\d+\))$|Merge pull request #\d+` ([link](https://regex101.com/r/jc5wh2/2)):

![image](https://user-images.githubusercontent.com/1444314/42503134-2f188186-8438-11e8-997d-facfabcd555f.png)

toolmantim:
Ah, good catch. Sorry you were hit by this bug, and thanks for the detailed report (with regex!). I'll get this fixed.

toolmantim:
Fixed in #15

Issue #7, opened 2018-07-10
Attribute changes to the original author


jeanregisser:
Hi,

Great tool!
I tested it on one of my repo:

<img width="669" alt="screen shot 2018-07-10 at 09 03 09" src="https://user-images.githubusercontent.com/57791/42494320-3ce15418-8420-11e8-82a3-28b948ecf6fe.png">

Would it be possible to attribute the changes to the original author instead of the person who merged the PR?

Thanks

toolmantim:
It's a good question, and one I raised myself with the GitHub peeps here:
https://twitter.com/toolmantim/status/1015762564955795456

Unfortunately there's no GitHub API to do it at the moment. But we'll leave this here, in case we find a solution at some point.

jeanregisser:
Ah there's been a slight misunderstanding, my original question was about the fact that the release notes contain the github handle of the person who merged the PR:

> - Fix warnings when used in react-native-web (#122) @jeanregisser`

I wanted to know if the notes could instead mention the original author of the PR, in this case it should be:

> - Fix warnings when used in react-native-web (#122) @msand`

toolmantim:
Ah! I understand. Yep, that’s a bug that needs fixing.

toolmantim:
Fixed in #15

PR #15, opened 2018-07-13
Non-semver support, squashed PRs, and fixed attribution

toolmantim:
- [x] Fix non-semver sorting (#9)
- [x] Match the squash PR commit message format (#8)
- [x] Fetch the PRs rather than using commit string matching (#7)
- [x] Tests


balloob:
Thanks!

PR #56, opened 2018-08-29
Add a $CONTRIBUTORS template variable

### LINKED ISSUE #53: Environment variable for all contributors
Thanks for this awesome app! We just installed it and are going to start using it on the @probot org.

It'd be cool if the template included a string of all the people who contributed changes for a release so we could give them a special shoutout. I'm imagining adding this to our template:

> Special thanks to everyone who contributed to this release: $CONTRIBUTORS

toolmantim:
This adds a `$CONTRIBUTORS` template variable, which is a comma separated list of GitHub usernames from the commit authors, commit commiters, and pull request authors.

For example:

```yml
template: |
  # Contributors
  
  ❤️ A big thanks to $CONTRIBUTORS
```

which would produce:

```md
# Contributors

❤️ A big thanks to @bkeepers, @salmanulfarzy and @toolmantim.
```

Fixes #53

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -41,7 +41,24 @@ module.exports.findReleases = async ({ app, context }) => {
   return { draftRelease, lastRelease }
 }
 
-module.exports.generateReleaseBody = ({ config, lastRelease, mergedPullRequests }) => {
+const contributorsSentence = ({ commits, pullRequests }) => {
+  const logins = new Set()
+
+  commits.forEach((commit) => {
+    logins.add(commit.committer.login)
+    logins.add(commit.author.login)
+  })
+
+  pullRequests.forEach((pullRequest) => {
+    logins.add(pullRequest.user.login)
+  })
+
+  const loginStrings = Array.from(logins).map((login) => `@${login}`).sort()
+
+  return loginStrings.slice(0, loginStrings.length - 1).join(', ') + ' and ' + loginStrings.slice(-1)
+}
+
+module.exports.generateReleaseBody = ({ commits, config, lastRelease, mergedPullRequests }) => {
   let body = config.template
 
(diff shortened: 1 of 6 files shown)


toolmantim:
@bkeepers could you give this a review please?

lib/releases.js
```diff
@@ -41,7 +41,24 @@ module.exports.findReleases = async ({ app, context }) => {
   return { draftRelease, lastRelease }
 }
 
-module.exports.generateReleaseBody = ({ config, lastRelease, mergedPullRequests }) => {
+const contributorsSentence = ({ commits, pullRequests }) => {
+  const logins = new Set()
+
+  commits.forEach((commit) => {
+    logins.add(commit.committer.login)
+    logins.add(commit.author.login)
```
bkeepers: `login` could be undefined if the email address can't be matched to a valid GitHub user.

toolmantim: Ah, good catch 👌🏼

toolmantim:
Oh, I forgot about co-authors! I’ll add support for that too.

toolmantim:
I was excited about adding co-author support, but after testing 60d9fd2 with the GitHub API (`curl 'https://api.github.com/repos/toolmantim/release-drafter/commits/60d9fd2c83c2ed7d6a6718fa66a59870ce4acf6c'`) it looks like the API doesn't have any support for co-authors, and I don't think there's a way to obtain their GitHub usernames from the co-author information 😞

toolmantim:
Unless there's a GitHub API I don't know about, I think this is as good as it can get. Time to 🚀

For the review and rubber-ducking @bkeepers 🙏🏼

toolmantim:
👋🏼 @bkeepers

PR #61, opened 2018-08-31
Correct default `change-template` in the README

ticky:
The `change-template` specified in the README was missing the `#` which is present in the code. Oops!


toolmantim:
Woo thanks @ticky!

Issue #91, opened 2018-11-29
Automatically name new draft releases according to semver


rnorth:
Firstly thanks for creating this - I've used it in a couple of places and found it very helpful!

I'm wondering if you would accept a contribution for an opt-in feature to automatically name new draft releases. What I would like is for each new release to be `Last Release + 0.0.1`. My goal is to make it really easy to make a new patch release (as our most common release type), so that 'working out the next patch version number' becomes a mental no-op, as performing the release can be just a case of hitting the publish button.

Obviously there will be times when it's more appropriate to increment the major or minor version, but this would be easily done manually with no additional mental overhead from now.

How does this sound as a potential new feature?

toolmantim:
Sounds great, I’ve pondered something similar. Would happily accept this as an option!

Over time I imagine we could add some smarts ala https://github.com/semantic-release/semantic-release for figuring out the new version number based on commits and/or labels.

toolmantim:
Let me know if you need any help any way @rnorth 😊

rnorth:
Thinking about this, I'm inclined to:

* introduce some new template variables (e.g. `NEXT_PATCH_VERSION`, `NEXT_MINOR_VERSION`, `NEXT_MAJOR_VERSION`)

* add two templatable fields to `.github/release-drafter.yml`: `defaultReleaseName` and `defaultReleaseTag`

This way, for our simple use case I can add to our config:
```yaml
defaultReleaseName: "$NEXT_PATCH_VERSION"
defaultReleaseTag:  "$NEXT_PATCH_VERSION"
```

but people with more interesting naming conventions, like a `v` prefix on their tags, could do this:
```yaml
defaultReleaseName: "$NEXT_PATCH_VERSION"
defaultReleaseTag:  "v$NEXT_PATCH_VERSION"   # note the v prefix here
```

Additionally, this would mean that these are available in the template, allowing things like this (if desired):

```yaml
template: |
  ## How to install

  Add this maven dependency:
  ``
  <dependency>
    <groupId>blah</groupId>
    <artifactId>blahblah</artifactId>
    <version>$NEXT_PATCH_VERSION</version>
  </dependency>
  ``
  ...

  ## What’s Changed

  $CHANGES
```

toolmantim:
Nice! That's an even better idea. I can see myself personally using this is a default:

```yams
defaultReleaseName: "$NEXT_PATCH_VERSION"
defaultReleaseTag:  "v$NEXT_PATCH_VERSION"
``

PR #92, opened 2018-12-03
Automatic next release naming

rnorth:
Refs #91

### DIFF
--- a/lib/versions.js
+++ b/lib/versions.js
@@ -1,26 +1,16 @@
-const semverRegex = /(\d+)\.(\d+)\.(\d+)/
+const semver = require('semver')
 
 module.exports.getVersionInfo = lastRelease => {
-  const lastReleaseTag = lastRelease.tag_name
-  const lastReleaseName = lastRelease.name
+  let lastVersion =
+    semver.coerce(lastRelease.tag_name) || semver.coerce(lastRelease.name)
 
-  let lastReleaseMatch
-  lastReleaseMatch = lastReleaseTag.match(semverRegex)
-  if (!lastReleaseMatch) {
-    lastReleaseMatch = lastReleaseName.match(semverRegex)
-  }
-
-  if (!lastReleaseMatch) {
+  if (!lastVersion) {
     return undefined
   }
 
-  const major = lastReleaseMatch[1] - 0
-  const minor = lastReleaseMatch[2] - 0
-  const patch = lastReleaseMatch[3] - 0
-
   return {
-    incrementedMajor: `${major + 1}.0.0`,
-    incrementedMinor: `${major}.${minor + 1}.0`,
-    incrementedPatch: `${major}.${minor}.${patch + 1}`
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     ]
   },
   "jest": {
-    "collectCoverage": true,
+    "collectCoverage": false,
```
rnorth: Accidental commit - need to remove.
Coverage collection was preventing me from using debugging with jest in vs code, so I meant to temporarily disable.

lib/versions.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    const major = (lastReleaseMatch[1] - 0);
+    const minor = (lastReleaseMatch[2] - 0);
+    const patch = (lastReleaseMatch[3] - 0);
+
+    return {
+        major,
+        minor,
+        patch,
```
rnorth: TODO: Remove these - they're confusing.

rnorth:
Tested here: https://github.com/rnorth/rd-test2/releases

rnorth:
Hi @toolmantim, sorry for the delay, but I believe this is ready for review now.

toolmantim:
Thanks so much @rnorth! Awesome stuff. I just merged in #105 and I'll fix the merge conflicts, review and merge when it's ready.

rnorth:
Should I/you perhaps run prettier on this PR first, so that conflicts are limited to just genuine changes?

toolmantim:
Hi @rnorth! I've updated the PR already with prettier, and started making a few changes. I want to bring the config names and docs in line with much of the other ones, and just make sure that as a new user it's not all too overwhelming as we add more stuff.

I'm pondering if I can remove "## Automatic Versioning Variables" altogether (and roll the examples up into the tables above) but I might look at it again with fresh eyes, and then I'll update the code to match too.

Thanks again!

toolmantim:
👋🏼 @rnorth. Your readme additions were really well documented, but I felt the readme was starting to get really complicated with all the options Release Drafter is now supporting. So I've just pushed my take on simplifying it all (which means slightly less accurate/descriptive), whilst still hopefully providing lots of good examples so you grok what you need fairly quickly.

Could you check out the new readme param names, and changes, and let me know what you think?

rnorth:
Hi @toolmantim, of course, that absolutely makes sense! 

I've had a quick look and I think this is much better - hits the right balance between accuracy and conciseness. Also, the placeholder names look good too.

Totally 👍 from me. Thanks for helping improve the PR so readily!

toolmantim:
Okay, thanks @rnorth! I did some digging and made one final change… switching to [semver](https://www.npmjs.com/package/semver) for the version parsing and incrementing. I figured it's going to be the Wild West of version numbering schemes out there. `semver` supports more fuzzy matching, and we don't have to worry about maintaining/fixing that regex. I reckon this is ready to rumble!

PR #113, opened 2018-12-24
Fix glitch in example config

rnorth:


toolmantim:
Whoops! Good catch. Thanks @rnorth.

PR #120, opened 2019-01-18
Fix example change template format

zhigang1992:
image


toolmantim:
Oh whoops! Thanks again @zhigang1992 😊

Issue #106, opened 2018-12-20
Support for rebase merging


TimonVS:
Currently rebase merging is not supported because Release Drafter checks for merge commits which don't exist when using rebase merging. At our company we're unable to use Release Drafter because we use rebase merging. Adding support for rebase merging requires a bit of a rewrite since it shouldn't rely on commits anymore, instead we can use the search endpoint (or search field in the GraphQL API) to fetch pull requests.

Here's an example query:

```gql
{
  search(type: ISSUE, query: "type:pr merged:>=2018-11-10 repo:toolmantim/release-drafter", first: 10) {
    nodes {
      ... on PullRequest {
        title
        number
        author {
          login
        }
        mergedAt
      }
    }
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```

Would you be interested in this @toolmantim? I could start working on it tomorrow and next week :)

toolmantim:
Hi @TimonVS! We're actually hitting API limits with the current method (if there's lots of PRs to fetch), so switching to GraphQL would be better. Would be happy to switch to this.

If we could ensure that #98 is still possible though (adding directly-pushed-to-master commits), it'd be much appreciated.

TimonVS:
It should drastically reduce the amounts of requests made. Currently I'm not sure if #98 will be possible with this approach, but I think it'd be a very useful feature to have. I've asked help for this in the Probot Slack :)

toolmantim:
For #98, perhaps you might need to query both the commits and the PRs? Either separately, or in the same query?

TimonVS:
That could work for merge commits and squash merging, but for rebase merging I will not be able to discern commits to master and commits that belong to a rebase merge. GitHub does however have a way to link the commits of a rebase merge to the PR as can be seen in the interface:
<img width="186" alt="pr-number" src="https://user-images.githubusercontent.com/876666/50375745-dda83380-0602-11e9-99b9-633fb2f17190.png">
But I'm unsure if they expose it in the API in some way (I couldn't find it). I haven't received a response on the Probot Slack yet, I will try to reach out to the API team directly.

toolmantim:
Ah… I see! I've spotted `associatedPullRequests` on `Ref` in the GraphQL API before, but I'm guessing it doesn't do what we need.

TimonVS:
[...]

PR #146, opened 2019-03-20
Convert tests to use Nock

TimonVS:
I've been converting tests to use Nock in preparation for #106. Using Nock will improve the trustworthiness of the tests, especially with regards to updating to Probot 9 that deprecated a lot of methods. Previously those methods would be mocked so they wouldn't cause errors in the tests. With Nock we purely mock the API endpoints so the Octokit methods will be called and will fail when deprecated/removed.


toolmantim:
I always felt deeply uncomfortable with the default recommended approach of mocking out the functions… I'm sorry for all that mess. I endorse switching to Nock 110%!

TimonVS:
No need for sorry :). The application is tested well and I could reuse almost all expectations.

I just finished up converted the remaining tests to use Nock. Let me know what you think!

toolmantim:
Amazing! Now, to figure out why Travis isn’t testing these PRs 🤔

toolmantim:
Well, I can’t see to unstick Travis. So here it goes!

PR #158, opened 2019-04-01
Add support for non-semver next version

### LINKED ISSUE #156: non-semver next version?
Would you be interested in supporting non-semver next version?

I know it is not much to change the name but selecting the right tag can be a little cumbersome.

Would love to contribute the work, if you'd be interested.

casz:
fixes #156 

The code might be too safe around ensuring that it can never be false unless intended 😄
I welcome suggestion this was just a quick and dirty implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 npm-debug.log
+package-lock.json
 *.pem
 .env
 coverage
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -184,8 +184,9 @@ module.exports.generateReleaseInfo = ({
   if (lastRelease) {
     let name = config['name-template'] || ''
     let tag = config['tag-template'] || ''
+    let patchVersions = config['patchVersions']
 
-    const versionInfo = getVersionInfo(lastRelease)
+    const versionInfo = getVersionInfo(lastRelease, patchVersions)
     if (versionInfo) {
       body = templateNextVersion(body, versionInfo)
       name = templateNextVersion(name, versionInfo)
--- a/lib/versions.js
+++ b/lib/versions.js
@@ -1,6 +1,6 @@
 const semver = require('semver')
 
-module.exports.getVersionInfo = lastRelease => {
+module.exports.getVersionInfo = (lastRelease, patchVersions = true) => {
   let lastVersion =
     semver.coerce(lastRelease.tag_name) || semver.coerce(lastRelease.name)
 
@@ -8,9 +8,18 @@ module.exports.getVersionInfo = lastRelease => {
     return undefined
   }
 
-  return {
+  let versionInfo = {
     incrementedMajor: semver.inc(lastVersion, 'major', true),
     incrementedMinor: semver.inc(lastVersion, 'minor', true),
     incrementedPatch: semver.inc(lastVersion, 'patch', true)
   }
+
+  if (patchVersions === false) {
+    delete versionInfo.incrementedPatch
+    Object.keys(versionInfo).forEach(function(key) {
+      versionInfo[key] = versionInfo[key].replace(/\.\d+$/, '')
+    })
+  }
+
+  return versionInfo
 }
--- a/test/fixtures/config/config-with-no-patch-versioning.yml
+++ b/test/fixtures/config/config-with-no-patch-versioning.yml
@@ -1,0 +1,4 @@
+template: Placeholder with example. Automatically calculated values are next major=$NEXT_MAJOR_VERSION, minor=$NEXT_MINOR_VERSION, patch=$NEXT_PATCH_VERSION
+name-template: 'v$NEXT_MINOR_VERSION (Code name: Placeholder)'
+tag-template: v$NEXT_MINOR_VERSION
+patchVersions: false
--- a/test/versions.test.js
+++ b/test/versions.test.js
@@ -53,4 +53,18 @@ describe('versions', () => {
 
     expect(versionInfo).toEqual(undefined)
   })
+
+  it('extracts a non-semvar version when patchVersions is false', () => {
+    const versionInfo = getVersionInfo(
+      {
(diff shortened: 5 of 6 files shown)


lib/releases.js
```diff
@@ -184,6 +184,8 @@ module.exports.generateReleaseInfo = ({
   if (lastRelease) {
     let name = config['name-template'] || ''
     let tag = config['tag-template'] || ''
+    let patchVersions = config['patchVersions'] || true
```
casz: ?
```suggestion
    let patchVersions = config['patch-versions'] || true
```

lib/versions.js
```diff
@@ -8,9 +8,19 @@ module.exports.getVersionInfo = lastRelease => {
...
+
+  let versionInfo = {
     incrementedMajor: semver.inc(lastVersion, 'major', true),
     incrementedMinor: semver.inc(lastVersion, 'minor', true),
     incrementedPatch: semver.inc(lastVersion, 'patch', true)
   }
+
+  if (patchVersions == false) {
+    Object.keys(versionInfo).forEach(function(key) {
+      versionInfo[key] = versionInfo[key].replace(/\.\d+$/, '')
```
casz: regex seemed easier.
Should we delete the incrementedPatch key when `patchVersions` is false to make it obvious that the user has to use `NEXT_MINOR_VERSION` or `NEXT_MAJOR_VERSION`?

casz:
Argh, the test was working than I did some minor changes and it stopped. 😆

casz:
👍

casz:
I should add a full `index.test` to test that config gets through 🤕

casz:
@toolmantim fully tested, I wonder if it follows the code style or you have any suggestion on how to improve it :sweat_smile:

test/versions.test.js
```diff
@@ -53,4 +53,18 @@ describe('versions', () => {
 
     expect(versionInfo).toEqual(undefined)
   })
+
+  it('extracts a non-semvar version when patchVersions is false', () => {
```
TimonVS:
```suggestion
  it('extracts a non-semver version when patchVersions is false', () => {
```

Just a minor spelling mistake :)

lib/versions.js
```diff
@@ -1,16 +1,25 @@
...
-  return {
+  let versionInfo = {
     incrementedMajor: semver.inc(lastVersion, 'major', true),
     incrementedMinor: semver.inc(lastVersion, 'minor', true),
     incrementedPatch: semver.inc(lastVersion, 'patch', true)
   }
+
+  if (patchVersions === false) {
+    delete versionInfo.incrementedPatch
+    Object.keys(versionInfo).forEach(function(key) {
```
TimonVS: Very minor nitpick, but the rest of the codebase uses arrow functions, maybe you can change this to be an arrow function to match the rest of the codebase?

.gitignore
```diff
@@ -1,7 +1,8 @@
 node_modules
 npm-debug.log
+package-lock.json
```
TimonVS: This project uses yarn (there's a yarn.lock file in the repo), so package-lock.json shouldn't have to be ignored I think :)

test/fixtures/config/config-with-no-patch-versioning.yml
```diff
@@ -0,0 +1,4 @@
+template: Placeholder with example. Automatically calculated values are next major=$NEXT_MAJOR_VERSION, minor=$NEXT_MINOR_VERSION, patch=$NEXT_PATCH_VERSION
+name-template: 'v$NEXT_MINOR_VERSION (Code name: Placeholder)'
+tag-template: v$NEXT_MINOR_VERSION
+patchVersions: false
```
TimonVS: Should be `patch-versions`, no?

lib/versions.js
```diff
@@ -1,16 +1,25 @@
...
+      versionInfo[key] = versionInfo[key].replace(/\.\d+$/, '')
```
TimonVS: Do you think it'd be more clear if we extract the regex to a named variable or adding a comment would make it a bit more clear what's going on this line? Regexes are always difficult to parse in my brain :)

.gitignore
casz: well, I was about to accidentally add it cause I used `npm install` instead of `yarn` it's a minor addition I can send it in another PR.
As a yarn project, you _should_ ignore the package-lock.json so it does not get accidentally added by unsuspecting devs 😢

lib/versions.js
casz: I have arrow function locally, sorry for not catching it 😅

test/versions.test.js
casz: Nicely spotted, I'll fix that!

test/fixtures/config/config-with-no-patch-versioning.yml
casz: 👍

lib/versions.js
casz: My brain mostly thinks regex 😭 I can replace it with substring

```suggestion
      let index = versionInfo[key].lastIndexOf('.')
      versionInfo[key] = versionInfo[key].slice(0, index)
```

.gitignore
TimonVS: Ah okay, that makes sense. It's always a bit confusing to know when to use yarn and when to use npm honestly. Would be nice if npm and yarn would detect if a lock file from the other tool is present and warn the user about it 😅

lib/versions.js
TimonVS: Both is fine. But in the case of regex it'd be nice if the regex had a proper name like `const REMOVE_LAST_VERSION_SECTION` (not sure if this is a very good name) so that the code can be scanned. I had to put the regex in regexr.com to figure out what it was doing now :)

casz:
superseded by #161

PR #160, opened 2019-04-02
Add ESLint config

TimonVS:
I've added an ESLint config to catch common mistakes such as an undefined variable. It's also integrated with `lint-staged` so that code with linting errors can't be commited. I've tried to make sure no unnecessary rules have been turned on to prevent frustration. I added this purely to catch mistakes like missing imports and undefined variables :)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+coverage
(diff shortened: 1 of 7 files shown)


TimonVS:
Is it okay if I merge this PR @toolmantim? :)

PR #169, opened 2019-04-07
Add links to PR Labeler and Probot Auto Labeler

TimonVS:
I added links to PR Labeler and Probot Auto Labeler to automate the categorization of release notes. Full disclosure: I'm the author of PR Labeler, but I think it could be very useful for users of Release Drafter which is the reason I created it :)


toolmantim:
Cross promotion of GitHub Apps FTW! Next we’ll be creating bundles, and putting together a huge sale 😁

PR #179, opened 2019-04-13
fix open-releases

casz:
in addition to #121

Just helping you out @toolmantim when you made the release :package:


casz:
Just helping you out @toolmantim when you made the release :package:

toolmantim:
Thanks @casz!

PR #164, opened 2019-04-03
Add support for rebase merging

TimonVS:
This PR will add support for rebase merging and drastically reduces the amount of HTTP calls to GitHub by utilizing the GraphQL API :)
This PR is based off of #160 

Solves: #106 

## Todo

- [x] Refactor code in `commits.js`
- [x] Fix `exclude-labels` tests
- [x] Generate new mocks using the release-drafter-test repo
- [x] Check order of changelog entries
- [x] Fix `version-template` config tests
- [x] Clean up unused code and fixtures
- [x] Change master to default branch of repo in `getCommitsWithAssociatedPullRequestsQuery`
- [x] Add tests for testing merge commits
- [x] Add tests for testing squash merging
- [x] Add tests for testing rebase merging
- [x] Deploy test instance

PR #161, opened 2019-04-02
add support for version-template

casz:
fixes #156 
supersedes #158 

I don't know if makes sense to support version template replacement inside the body/tag/name :confused: 
Also in doubt about the readme :sweat:

Please take a look @toolmantim and @TimonVS

PR #181, opened 2019-04-17
Add beta GitHub action support

toolmantim:
Adds beta support for running Release Drafter as a GitHub Action, instead of a GitHub App. This means that people can run it entirely locally in their repository, manage version updates, and get failure output for misconfigured `.github/release-drafter.yml` configuration files

- [x] Spike the required plumbing
- [x] Use it on this repo
- [ ] Exit status neutral when branch match fails


TimonVS:
Nice 🙌. Looking forward to use it as an action. This will also make testing PRs like #161 and #164 so much easier!

josemarluedke:
Looking forward to use this. Does this change means that release-drafter won't need to have full write access to the repo?

toolmantim:
Yes and no @josemarluedke. Yes in that you don’t even need to authorize the GitHub App run by me at all, but no in that although the action will run in a Docker container attached to your own account, Actions run with full repo write access, and there’s no way to limit their scopes (yet).

Dockerfile
```diff
@@ -1,6 +1,16 @@
...
+LABEL com.github.actions.description="Drafts your next release notes as pull requests are merged into master."
+LABEL com.github.actions.icon="edit-2"
+LABEL com.github.actions.color="orange"
+
 WORKDIR /src
 ADD     yarn.lock package.json /src/
 RUN     yarn
 ADD     . /src
-CMD     ["npm", "start"]
+CMD     ["npm", "receive"]
```
TimonVS: I tried using the action in one of my repos, I got an error about `npm receive`. I think it should be:

```suggestion
CMD     ["npm", "run", "receive"]
```

toolmantim: Good catch! I made that change, let me know if it works.

TimonVS: `npm run receive` seems to work now :)
I'm now getting the following error:

```
docker run image=release-drafter:add-github-action-support entrypoint=[] cmd=[]
npm ERR! path /github/workspace/package.json
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall open
npm ERR! enoent ENOENT: no such file or directory, open '/github/workspace/package.json'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     /github/home/.npm/_logs/2019-04-26T13_12_48_443Z-debug.log
```

It's possible to debug locally using [act](https://github.com/nektos/act).

halkeye:
What does the recieve step need? Could I make it as a step in my jenkinsfile too? This is so cool, I've always meant to poke at the framework

Dockerfile
toolmantim: Ah, looks like we can’t use `WORKDIR`, which finally explains to me why all the GH actions focus on entrypoints.

I hadn’t quite gotten to the point of testing this on another repo with act! You’re one step ahead. Act is awesome tho. I used it for dev’ing https://github.com/buildkite/trigger-pipeline-action

halkeye: If you drop the workdir and use ./ In the add you should be fine

toolmantim:
@halkeye from memory, it takes the GitHub webhook payload via STDIN. You could use something like https://github.com/nektos/act to run it outside of GH infra, though you’d need to set up GH credentials? 🤷🏼‍♂️ At which point, you probably just want a release-drafter CLI?

Dockerfile
TimonVS: You guys know much more about Docker than me probably, but maybe we can take this as an example: https://github.com/JasonEtco/todo/blob/dockerfile/Dockerfile? :)

toolmantim: All fixed!

TimonVS: Nice! 😄

toolmantim: Strangely, the `branches` key in the `release-drafter` config doesn't seem to be working as expected. But it should give you a chance to test it @TimonVS!

TimonVS: Seems to work for me now! I didn't do very extensive testing though, but we can start using it on this repo before we publish it to the marketplace to see if everything works nicely? I will also start using the Action on the repos at work starting on Monday.

TimonVS:
If I understand correctly the Action will run for any push, also for branches that are not configured in the config. I was thinking that we could add a filter Action in the workflow that will do an early exit (neutral status) if the branch name doesn't match the ones in the config. This Action can be written as a shell script and thus use a very lightweight image (e.g. `debian:stable-slim`).

toolmantim:
Ace. This is close to ready for merging for testing, sounds like a good plan.

For branch filters, we could just make the branch match that already exists in the code exit with the correct neutral status, rather than having people having people configure another action in their workflow?

toolmantim:
I just realized a big advantage to running it as an action: new draft releases will finally be attributed to a person rather than release drafter.

TimonVS:
> For branch filters, we could just make the branch match that already exists in the code exit with the correct neutral status, rather than having people having people configure another action in their workflow?

Definitely! I was just thinking that initializing a Node.js container on every push is a bit wasteful as it has to install dependencies and such. But I think we can solve this by publishing the container to Docker Hub as well? I'm by no means a Docker expert though 😅

casz:
Docker push should be apart of the release :)

toolmantim:
Yeah, something to investigate once it’s actually working as expected. This way I can run it on branches and it sources from the branch. I wonder if GH are ever going to support their own registry

toolmantim:
Turns out, the answer is yes, yes they are: https://github.com/features/package-registry

Issue #168, opened 2019-04-06
Allow custom user templating?


casz:
Would be nice to give the user some templating power.

For instance, I need a regex capturing of `JENKINS-\d+` like `JENKINS-123` and turn it into a Markdown link of `[JENKINS-123](https://issues.jenkins-ci.org/browse/JENKINS-123)` [JENKINS-123](https://issues.jenkins-ci.org/browse/JENKINS-123)

Would be possible to add once #161 is merged.

The user templating would be run last.

halkeye:
are you tackling this? it seems #161 is merged. Otherwise I might be interested in trying

casz:
Go ahead

TimonVS:
Looks like this has been resolved in #185 😄

PR #185, opened 2019-04-24
Add supports for custom replacer

halkeye:
super related to #168 

Allows custom regexes to be applied.


index.js
```diff
@@ -14,12 +14,22 @@ module.exports = app => {
...
       categories: [],
-      'exclude-labels': []
+      'exclude-labels': [],
+      replacers: []
     }
     const config = Object.assign(
       defaults,
       (await getConfig(context, configName)) || {}
     )
+    config.replacers = config.replacers.filter(replacer => {
```
TimonVS: Would it be an idea to extract this to a separate function called `validateReplacers` as to keep the main function clean and it would be self-documenting instead of having to read the code? :)

lib/releases.js
```diff
@@ -144,11 +144,15 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
-              $AUTHOR: pullRequest.user.login
-            })
+            template(
+              config['change-template'],
+              {
+                $TITLE: pullRequest.title,
+                $NUMBER: pullRequest.number,
+                $AUTHOR: pullRequest.user.login
+              },
+              config['replacers']
```
TimonVS: Minor nitpick, but could we change this to:
```suggestion
              config.replacers
```

test/index.test.js
```diff
@@ -716,6 +716,53 @@ Previous tag: ''
           payload: require('./fixtures/push')
         })
 
+        expect.assertions(1)
+      })
+    })
+    describe('custom replacers', () => {
```
TimonVS: Could you add tests with multiple (maybe overlapping?) replacers as well?

lib/releases.js
halkeye: I would have totally thought eslint would have caught this. I kinda just copy and replaced it

lib/releases.js
```diff
@@ -168,24 +172,28 @@ module.exports.generateReleaseInfo = ({
...
     let name = config['name-template'] || ''
     let tag = config['tag-template'] || ''
 
     const versionInfo = getVersionInfo(lastRelease, config['version-template'])
     if (versionInfo) {
-      body = template(body, versionInfo)
-      name = template(name, versionInfo)
-      tag = template(tag, versionInfo)
+      body = template(body, versionInfo, config.replacers)
+      name = template(name, versionInfo, config.replacers)
```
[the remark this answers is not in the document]
halkeye: :) after my first attempt didn't work, i just put it everywhere
but you are right, i've removed it everywhere but body

lib/template.js
```diff
@@ -15,6 +15,12 @@ const template = (string, obj) => {
     }
     return result
   })
+  if (customReplacers) {
+    customReplacers.forEach(({ search, replace }) => {
+      str = str.replace(new RegExp(search, 'g'), replace)
```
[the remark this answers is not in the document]
halkeye: I agree with the flags, i'll add it as another property.
But the / syntax makes things super difficult because shouldn't get imported into `new RegExp`. So i'd have to process and delete the slashes, and it gets complicated.

```
new RegExp('/abc/')
//abc//
new RegExp('abc')
/abc/
```

lib/releases.js
```diff
@@ -168,24 +172,28 @@ module.exports.generateReleaseInfo = ({
...
+    body,
+    {
+      $PREVIOUS_TAG: lastRelease ? lastRelease.tag_name : '',
+      $CHANGES: generateChangeLog(mergedPullRequests, config),
+      $CONTRIBUTORS: contributorsSentence({
+        commits,
+        pullRequests: mergedPullRequests
+      })
+    },
+    config.replacers
```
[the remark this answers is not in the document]
halkeye: Good call, Its fixed now by only replacing once.

test/index.test.js
halkeye: Its easier to set them up in template.test, so I did it there

halkeye:
Okay, thanks for the review, I think i got everything brought up handled.

halkeye:
travis needs a retry

lib/template.js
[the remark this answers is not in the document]
casz: I'd prefer not adding another property and instead suggest you look at [regex-parser](https://www.npmjs.com/package/regex-parser)

Should handle parsing the regex in both formats only concern is to take the regex you get and add the default `g` flag :sweat_smile:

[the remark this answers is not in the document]
halkeye: Ooooo I didn't know it existed. I always assumed parsing regex was super complicated.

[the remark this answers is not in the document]
casz: Feel free to try out regex-parser: https://npm.runkit.com/regex-parser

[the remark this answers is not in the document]
casz:
```javascript
const regexParser = require("regex-parser")
let re = regexParser('/JENKINS-(\d+)/i')
if (!re.flags.includes('g')) { re = new RegExp(re.source, re.flags + 'g') }
if (re.flags.includes('g')) { console.log('yes') }
else { console.log('no') } 
console.log(re.source, re.flags)
```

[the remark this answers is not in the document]
casz: One could argue for `jenkins.com` regex you'd want to escape the `.` [escape-string-regexp](https://www.npmjs.com/package/escape-string-regexp)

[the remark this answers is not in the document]
halkeye: so that would force global with no way to disable it.  I can't decide my feeling on that.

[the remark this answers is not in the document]
halkeye: I wouldn't want to auto escape anything for the user.

Could make it so if it starts with a /, its a regex, otherwise its a string?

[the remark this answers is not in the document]
casz: If we support regex I'd be fine without auto global 😊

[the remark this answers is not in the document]
casz: The check should be does not startswith / and flags is empty for auto global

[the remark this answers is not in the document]
halkeye: /foo/bar/
flags is ""

Does that mean they don't want global? or want the default?
I think once you regex, you control all the flags

lib/template.js
```diff
@@ -28,26 +30,29 @@ const template = (string, obj, customReplacers) => {
...
-      if (replacer.flags === undefined || replacer.flags === null) {
-        replacer.flags = 'g'
-      }
-      return replacer
-    })
-    .filter(replacer => {
       try {
-        new RegExp(replacer.search, replacers.flags)
-        return true
+        if (replacer.search.startsWith('/')) {
```
casz: what if it is a part of a URL :laughing:

halkeye: it'll silently fail?

casz: > it'll silently fail?

exactly my point and they would have to do `/\/my-url/g` :)

halkeye: they sure will

halkeye: only solution i see is to make 2 new fields, string and regex, and be explicit

casz: For most simple regex which I imagine will be most, they can get away with stuff like `JENKINS-(\d+)` or similar hence my reluctance to add more explicit behavior

halkeye: okay, changed it to be explicit

halkeye: I appear to get notifications late.

I'm indifferent.

Just tell me the behavior you want

README.md
```diff
@@ -148,6 +149,20 @@ For example, if your project doesn't use patch version numbers, you can set `ver
...
 
+## Replacers
+
+With the `replacers` option you can provide string replaces for your template after all other template information is finished being processed.
+
+```yml
+replacers:
+  - regex: /JENKINS-(\d+)/g
+    replace: https://issues.jenkins-ci.org/browse/JENKINS-$1
+  - search: myname
```
toolmantim: I think we made a big mistake by not quoting the string values in the configuration examples, and many people get tripped up once they use a semi-colon. Perhaps we should do this for these? It's easy for the `search` option, but I wonder if we need it for `regex` too?

I wonder what a better example for this one is too? 🤔

README.md
```diff
@@ -148,6 +149,20 @@ For example, if your project doesn't use patch version numbers, you can set `ver
...
+    replace: My Name
+```
+
+Multiple replacers will be applied in order, so if your replacement string matches something in the second replacer, it'll get triggered again
```
toolmantim: Let's remove this, in favour of the paragraph above at the start of the section?

README.md
```diff
@@ -66,6 +66,7 @@ You can configure Release Drafter using the following key in your `.github/relea
 | `branches`            | Optional | The branches to listen for configuration updates to `.github/release-drafter.yml` and for merge commits. Useful if you want to test the app on a pull request branch. Default is the repository’s default branch. |
 | `categories`          | Optional | Categorize pull requests using labels. Refer to [Categorize Pull Requests](#categorize-pull-requests) to learn more about this option.                                                                            |
 | `exclude-lables`      | Optional | Exclude pull requests using labels. Refer to [Exclude Pull Requests](#exclude-pull-requests) to learn more about this option.                                                                                     |
+| `replacers`           | Optional | Manual search and replace strings. Applied in order. Refer to [Replacers](#replacers) to learn more about this option.                                                                                            |
```
toolmantim:
```suggestion
| `replacers`           | Optional | Search and replace content in the generated changelog body. Refer to [Replacers](#replacers) to learn more about this option.                                                                                            |
```

README.md
toolmantim: I just changed the string quoting in #189

README.md
```diff
@@ -148,6 +149,20 @@ For example, if your project doesn't use patch version numbers, you can set `ver
 
 Release Drafter requires full write, because GitHub does not offer a limited scope for only writing releases. **Don't install Release Drafter to your entire GitHub account — only add the repositories you want to draft releases on.**
 
+## Replacers
```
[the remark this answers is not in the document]
halkeye: Done!

halkeye:
Thanks for all the comments and suggestions @toolmantim 

I think i addressed them all, as well as fought with the master merge, so I think its ready to go again.

toolmantim:
Thanks for the updates, you’re awesome! 🎉🎉🎉🎉

halkeye:
:D SO EXCITED

How often do you release it to the actual github app? when can i start using it?

halkeye:
:D SO EXCITED

How often do you release it to the actual github app? when can i start using it?

toolmantim:
When the new npm package gets released, it gets deployed to match. I think the only thing we need to check is whether @TimonVS reckons we should add #188 beforehand?

TimonVS:
I'm leaning towards adding the sorting option first because we're changing the default sorting direction which might be confusing to users. But I don't have a strong opinion about it so I'll leave the decision to you @toolmantim, I'll probably have some time tomorrow to start implementing the sorting option :)

lib/template.js
```diff
@@ -15,6 +19,41 @@ const template = (string, obj) => {
...
+  return str
+}
+
+function validateReplacers({ app, context, replacers }) {
+  return replacers
+    .map(replacer => {
+      try {
+        if (replacer.regex) {
+          return {
+            search: regexParser(replacer.regex),
```
casz: cheeky *beep* :laughing: I spent an hour thinking my schema addition was borked when in reality you modified the schema. :rage: At myself for now reading the code my closely before deleting it :laughing:

PR #189, opened 2019-04-26
String quote example config string values

toolmantim:
This will help to avoid the most common misconfiguration mistake: special characters in some of the configuration values.


toolmantim:
I wish we didn't have to do this, but YAML is just too easy to mess up.

Issue #188, opened 2019-04-26
Support for changing sorting direction


TimonVS:
There's currently a small bug where the first draft with no prior releases will be sorted in reverse-chronological order but if the repo already has one or more prior releases it will be sorted in chronological order. I assume this happens because there's a difference in sorting order with `compareCommits` and `listCommits` (https://github.com/toolmantim/release-drafter/blob/master/lib/commits.js#L4). After #164 is merged, PRs will always be sorted in reverse-chronological order. I think it makes sense to add a `sort-direction` option that accepts either `chronological` or `reverse-chronological` so that users can decide for themselves what they like best.

PR #192, opened 2019-04-27
Add `sort-direction` option

TimonVS:
This PR adds the `sort-direction` option. Valid values are `ascending` or `descending`. I do notice that the list of configuration options is getting quite long, we may want to split up more advanced configuration options like this one into another document in the near future.

Fixes: #188

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,11 @@ const { isTriggerableBranch } = require('./lib/triggerable-branch')
 const { findReleases, generateReleaseInfo } = require('./lib/releases')
 const { findCommitsWithAssociatedPullRequests } = require('./lib/commits')
 const { validateReplacers } = require('./lib/template')
+const {
+  validateSortDirection,
+  sortPullRequests,
+  SORT_DIRECTIONS
+} = require('./lib/sort-pull-requests')
 const log = require('./lib/log')
 
 const configName = 'release-drafter.yml'
@@ -16,7 +21,8 @@ module.exports = app => {
       'version-template': `$MAJOR.$MINOR.$PATCH`,
       categories: [],
       'exclude-labels': [],
-      replacers: []
+      replacers: [],
+      'sort-direction': SORT_DIRECTIONS.descending
     }
     const config = Object.assign(
       defaults,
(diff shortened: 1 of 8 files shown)


TimonVS:
I can't select people who aren't a member of the repo as a reviewer 🤔. But I'd appreciate @casz review as well 🙇‍♂️.

casz:
I'd be happy to be considered a member of the repo :+1: :muscle:

index.js
```diff
@@ -27,6 +33,7 @@ module.exports = app => {
       context,
       replacers: config.replacers
     })
+    config['sort-direction'] = validateSortDirection(config['sort-direction'])
```
TimonVS: It's probably better to solve this with a Joi schema?

casz: I'll see if I can get around to it today :+1:

TimonVS: No need to hurry! If you don't get around to it we can simply create an issue to improve this later :)

lib/sort-pull-requests.js
```diff
@@ -0,0 +1,37 @@
+const SORT_DIRECTIONS = {
+  ascending: 'ascending',
+  descending: 'descending'
+}
+
+module.exports.SORT_DIRECTIONS = SORT_DIRECTIONS
+
+module.exports.validateSortDirection = sortDirection => {
+  console.log(Object.keys(SORT_DIRECTIONS).includes(sortDirection))
```
toolmantim: Did you mean to leave this in?

casz:
Should be easy for me to add into schema so perhaps my PR supersedes his? :laughing:

toolmantim:
So, after doing a quick test, I'm thinking we should reverse the default sort order @TimonVS, and make it `ascending`. What do you reckon from your testing?

TimonVS:
Your call, `ascending` kind of was the default before so I'm fine with changing it to `ascending`. Personally I think I prefer `descending` but I don't have a strong opinion on it :)

toolmantim:
Cool. I remember why I thought that maybe `ascending` was better again — I found while editing the latest release notes for Release Drafter, if you have PRs that build upon one another, they don't make much sense in reverse chronological order and takes more mental gymnastics.

But TBH, I'm not really fussed enough to change it either. Let's wait for feedback and see.

casz:
My 2 cents is I found ascending confusing, took me a while to get over 😆 however I much prefer descending.

PR #193, opened 2019-04-27
Improve config validation using a Joi schema

### LINKED ISSUE #353: Please don't break of give a nice error message when using label for multiple labels
This:

```
categories:
  - title: '🧰 Maintenance'
    label:
      - 'chore'
      - 'dependencies'
```

Will give the error

> Cannot read property 'title' of undefined

Proposal

- Don't care if `label` or `labels` is used, or
- Show a better error message

See also https://github.com/release-drafter/release-drafter/issues/320

casz:
The schema works. Tests are passing.

@toolmantim @TimonVS feel free to add your suggestions

- [x] add schema
- [x] improve schema with errors and descriptions
- [x] generate schema
- [x] add test/log for yaml exception
- [x] fixup JSON schema for publishing
- [x] add more schema tests

fixes #353

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,16 +14,13 @@ module.exports = app => {
       getConfig: require('probot-config')
     })
 
+    if (config === null) return
+
     // GitHub Actions merge payloads slightly differ, in that their ref points
     // to the PR branch instead of refs/heads/master
     const ref = process.env['GITHUB_REF'] || context.payload.ref
 
     const branch = ref.replace(/^refs\/heads\//, '')
-
-    if (!config.template) {
-      log({ app, context, message: 'No valid config found' })
-      return
-    }
 
     if (!isTriggerableBranch({ branch, app, context, config })) {
       return
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,47 +1,22 @@
-const {
-  SORT_DIRECTIONS,
-  validateSortDirection
-} = require('./sort-pull-requests')
-const { validateReplacers } = require('./template')
+const { validateSchema } = require('./schema')
+const { DEFAULT_CONFIG } = require('./default-config')
+const log = require('./log')
 
 const CONFIG_NAME = 'release-drafter.yml'
-
-const DEFAULT_CONFIG = Object.freeze({
-  'change-template': `* $TITLE (#$NUMBER) @$AUTHOR`,
-  'no-changes-template': `* No changes`,
-  'version-template': `$MAJOR.$MINOR.$PATCH`,
-  categories: [],
-  'exclude-labels': [],
-  replacers: [],
-  'sort-direction': SORT_DIRECTIONS.descending,
-  prerelease: false
-})
-
-module.exports.DEFAULT_CONFIG = DEFAULT_CONFIG
 
 module.exports.getConfig = async function getConfig({
   app,
   context,
   getConfig: fetchConfig
 }) {
-  const defaults = {
-    ...DEFAULT_CONFIG,
-    branches: context.payload.repository.default_branch
-  }
-
-  const config = await fetchConfig(context, CONFIG_NAME, defaults)
-
   try {
-    config.replacers = validateReplacers({
-      app,
-      context,
-      replacers: config.replacers
-    })
+    const repoConfig = await fetchConfig(context, CONFIG_NAME, DEFAULT_CONFIG)
+
+    const config = validateSchema(app, context, repoConfig)
+
+    return config
   } catch (error) {
-    config.replacers = []
+    log({ app, context, error, message: 'Invalid config file' })
+    return null
   }
-
-  config['sort-direction'] = validateSortDirection(config['sort-direction'])
-
-  return config
 }
--- a/lib/log.js
+++ b/lib/log.js
@@ -1,11 +1,11 @@
-module.exports = ({ app, context, message, info }) => {
+module.exports = ({ app, context, message, error }) => {
   const repo = context.payload.repository
   const prefix = repo ? `${repo.full_name}: ` : ''
   const logString = `${prefix}${message}`
 
-  if (info) {
-    app.log(logString, info)
+  if (error) {
+    app.log.warn(error, logString)
   } else {
-    app.log(logString)
+    app.log.info(logString)
   }
 }
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -188,8 +188,8 @@ module.exports.generateReleaseInfo = ({
   )
 
   if (lastRelease) {
-    let name = config['name-template'] || ''
-    let tag = config['tag-template'] || ''
+    let name = config['name-template']
+    let tag = config['tag-template']
 
     const versionInfo = getVersionInfo(lastRelease, config['version-template'])
     if (versionInfo) {
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1,0 +1,105 @@
+const _ = require('lodash')
+const Joi = require('@hapi/joi')
+const { SORT_BY, SORT_DIRECTIONS } = require('./sort-pull-requests')
+const { DEFAULT_CONFIG } = require('./default-config')
+const { validateReplacers } = require('./template')
+
+const schema = context => {
+  const defaultBranch = _.get(
+    context,
+    'payload.repository.default_branch',
+    'master'
+  )
+  return Joi.object().keys({
+    branches: Joi.array()
+      .items(Joi.string())
+      .default([defaultBranch]),
+
+    'change-template': Joi.string().default(DEFAULT_CONFIG['change-template']),
+
+    'no-changes-template': Joi.string().default(
+      DEFAULT_CONFIG['no-changes-template']
+    ),
+
+    'version-template': Joi.string().default(
+      DEFAULT_CONFIG['version-template']
+    ),
+
+    'name-template': Joi.string()
+      .allow('')
+      .default(DEFAULT_CONFIG['name-template']),
+
+    'tag-template': Joi.string()
+      .allow('')
+      .default(DEFAULT_CONFIG['tag-template']),
+
+    'exclude-labels': Joi.array()
+      .items(Joi.string())
+      .default(DEFAULT_CONFIG['exclude-labels']),
+
+    'sort-by': Joi.string()
+      .valid(SORT_BY.mergedAt, SORT_BY.title)
+      .default(DEFAULT_CONFIG['sort-by']),
+
+    'sort-direction': Joi.string()
+      .valid(SORT_DIRECTIONS.ascending, SORT_DIRECTIONS.descending)
+      .default(DEFAULT_CONFIG['sort-direction']),
+
+    prerelease: Joi.boolean().default(DEFAULT_CONFIG.prerelease),
+
+    replacers: Joi.array()
+      .items(
+        Joi.object().keys({
+          search: Joi.string()
+            .required()
+            .error(
+              new Error('"search" is required and must be a regexp or a string')
+            ),
+          replace: Joi.string()
+            .allow('')
+            .required()
+        })
+      )
+      .default(DEFAULT_CONFIG.replacers),
+
+    categories: Joi.array()
+      .items(
+        Joi.object()
+          .keys({
+            title: Joi.string().required(),
+            label: Joi.string(),
+            labels: Joi.array()
+              .items(Joi.string())
+              .single()
+          })
+          .xor('label', 'labels')
+      )
+      .default(DEFAULT_CONFIG.categories),
+
+    template: Joi.string().required(),
+
+    _extends: Joi.string()
+  })
+}
+
+module.exports.schema = schema
+
+const validateSchema = (app, context, repoConfig) => {
+  const { error, value } = schema(context).validate(repoConfig)
+  const config = value
+  if (error) throw error
+
+  try {
+    config.replacers = validateReplacers({
+      app,
+      context,
+      replacers: config.replacers
+    })
+  } catch (error) {
+    config.replacers = []
+  }
+
+  return config
+}
+
+module.exports.validateSchema = validateSchema
--- a/lib/sort-pull-requests.js
+++ b/lib/sort-pull-requests.js
@@ -11,16 +11,6 @@ const SORT_DIRECTIONS = {
 module.exports.SORT_BY = SORT_BY
 module.exports.SORT_DIRECTIONS = SORT_DIRECTIONS
 
-module.exports.validateSortBy = sortBy => {
-  return Object.values(SORT_BY).includes(sortBy) ? sortBy : SORT_BY.mergedAt
-}
-
-module.exports.validateSortDirection = sortDirection => {
-  return Object.keys(SORT_DIRECTIONS).includes(sortDirection)
-    ? sortDirection
-    : SORT_DIRECTIONS.descending
-}
-
 module.exports.sortPullRequests = (pullRequests, sortBy, sortDirection) => {
   const getSortFieldFn = sortBy === SORT_BY.title ? getTitle : getMergedAt
 
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "now-start": "PRIVATE_KEY=$(echo $PRIVATE_KEY_BASE64 | base64 -d) npm start",
     "test": "jest",
     "test:watch": "jest --watch --notify --notifyMode=change --coverage",
+    "generate-schema": "node ./bin/generate-schema.js print",
     "lint": "eslint '**/*.js'",
     "prettier": "prettier --write **/*.{js,md,json}",
     "postversion": "npm run test && git push && git push --tags && npm publish && npm run deploy && npm run open-releases",
@@ -19,6 +20,7 @@
   },
   "dependencies": {
     "@actions/core": "^1.2.0",
+    "@hapi/joi": "15.1.1",
     "compare-versions": "3.5.1",
     "escape-string-regexp": "2.0.0",
     "lodash": "4.17.15",
@@ -34,6 +36,7 @@
     "eslint-plugin-prettier": "3.1.2",
     "husky": "3.1.0",
     "jest": "24.9.0",
+    "joi-to-json-schema": "^5.1.0",
     "lint-staged": "9.5.0",
     "nock": "11.7.1",
     "nock-knock": "2.0.0",
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,131 @@
+{
+  "title": "JSON schema for Release Drafter yaml files",
+  "id": "https://github.com/release-drafter/release-drafter/blob/master/schema.json",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "branches": {
+      "default": ["master"],
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "change-template": {
+      "default": "* $TITLE (#$NUMBER) @$AUTHOR",
+      "type": "string"
+    },
+    "no-changes-template": {
+      "default": "* No changes",
+      "type": "string"
+    },
+    "version-template": {
+      "default": "$MAJOR.$MINOR.$PATCH",
+      "type": "string"
+    },
+    "name-template": {
+      "anyOf": [
+        {
+          "type": "string",
+          "enum": [""]
+        },
+        {
+          "default": "",
+          "type": "string"
+        }
+      ]
+    },
+    "tag-template": {
+      "anyOf": [
+        {
+          "type": "string",
+          "enum": [""]
+        },
+        {
+          "default": "",
+          "type": "string"
+        }
+      ]
+    },
+    "exclude-labels": {
+      "default": [],
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "sort-by": {
+      "default": "merged_at",
+      "enum": ["merged_at", "title"],
+      "type": "string"
+    },
+    "sort-direction": {
+      "default": "descending",
+      "enum": ["ascending", "descending"],
+      "type": "string"
+    },
+    "prerelease": {
+      "default": false,
+      "type": "boolean"
+    },
+    "replacers": {
+      "default": [],
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "search": {
+            "type": "string"
+          },
+          "replace": {
+            "anyOf": [
+              {
+                "type": "string",
+                "enum": [""]
+              },
+              {
+                "type": "string"
+              }
+            ]
+          }
+        },
+        "additionalProperties": false,
+        "patterns": [],
+        "required": ["search", "replace"]
+      }
+    },
+    "categories": {
+      "default": [],
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "title": {
+            "type": "string"
+          },
+          "label": {
+            "type": "string"
+          },
+          "labels": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          }
+        },
+        "additionalProperties": false,
+        "patterns": [],
+        "required": ["title"]
+      }
+    },
+    "template": {
+      "type": "string"
+    },
+    "_extends": {
+      "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "patterns": [],
+  "required": []
+}
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -24,10 +24,17 @@ Pc6zWtW2XuNIGHw9pDj7v1yDolm7feBXLg8/u9APwHDy
 
 describe('release-drafter', () => {
   let probot
+  let logger
 
   beforeEach(() => {
+    logger = jest.fn()
     probot = new Probot({ id: 179208, cert, Octokit })
     probot.load(releaseDrafter)
+    probot.logger.addStream({
+      level: 'trace',
+      stream: { write: logger },
+      type: 'raw'
+    })
 
     nock('https://api.github.com')
       .post('/app/installations/179208/access_tokens')
@@ -1220,6 +1227,50 @@ Previous tag: ''
       })
 
       expect.assertions(1)
+    })
+  })
+
+  describe('config error handling', () => {
+    it('schema error', async () => {
+      getConfigMock('config-with-schema-error.yml')
+
+      const payload = require('./fixtures/push')
+
+      await probot.receive({
+        name: 'push',
+        payload
+      })
+      expect(logger).toHaveBeenCalledWith(
+        expect.objectContaining({
+          msg: expect.stringContaining('Invalid config file'),
+          err: expect.objectContaining({
+            message: expect.stringContaining(
+              '"search" is required and must be a regexp or a string'
+            )
+          })
+        })
+      )
+    })
+
+    it('yaml exception', async () => {
+      getConfigMock('config-with-yaml-exception.yml')
+
+      const payload = require('./fixtures/push')
+
+      await probot.receive({
+        name: 'push',
+        payload
+      })
+      expect(logger).toHaveBeenCalledWith(
+        expect.objectContaining({
+          msg: expect.stringContaining('Invalid config file'),
+          err: expect.objectContaining({
+            message: expect.stringContaining(
+              'end of the stream or a document separator is expected at line 1, column 18:'
+            )
+          })
+        })
+      )
     })
   })
 })
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -1,0 +1,68 @@
+const { schema } = require('../lib/schema')
+const schemaJson = require('../schema.json')
+const { jsonSchema } = require('../bin/generate-schema')
+
+const template = '$CHANGES'
+
+const context = {
+  payload: {
+    repository: {
+      default_branch: 'foobar'
+    }
+  }
+}
+
+const validConfigs = [
+  [{ template }],
+  [{ template, replacers: [{ search: '123', replace: '' }] }],
+  [{ template, replacers: [{ search: '/123/gi', replace: '' }] }],
+  [{ template, replacers: [{ search: '/123/gi', replace: '123' }] }]
+]
+
+const invalidConfigs = [
+  [{ template: true }, 'must be a string'],
+  [{ template: 1 }, 'must be a string'],
+  [{ template: ['👶'] }, 'must be a string'],
+  [{ template: { '👶': 'a' } }, 'must be a string'],
+  [{ template: null }, 'must be a string'],
+  [{ template: '' }, 'is not allowed to be empty'],
+  [{ 'change-template': ['* $TITLE (#$NUMBER) @$AUTHOR'] }, 'must be a string'],
+  [{ 'change-template': null }, 'must be a string'],
+  [{ replacers: [{ search: 123 }] }, 'must be a regexp or a string'],
+  [{ replacers: [{ search: '123' }] }, 'is required'],
+  [
+    { replacers: [{ replace: 123 }] },
+    'is required and must be a regexp or a string'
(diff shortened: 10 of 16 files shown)


README.md
```diff
@@ -60,13 +60,13 @@ You can configure Release Drafter using the following key in your `.github/relea
...
 | `name-template`       | Optional | The template for the name of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                                                            |
 | `tag-template`        | Optional | The template for the tag of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                                                             |
-| `version-template`    | Optional | The template to use when calculating the next version number for the release. Useful for projects that don't use semantic versioning. Default: `"$MAJOR.$MINOR.$PATCH"`                                           |
+| `version-template`    | Optional | The template to use when calculating the next version number for the release. Useful for projects that don’t use semantic versioning. Default: `"$MAJOR.$MINOR.$PATCH"`                                           |
```
toolmantim: 🥰

README.md
```diff
@@ -139,27 +139,27 @@ exclude-labels:
...
 
 ## Replacers
 
-You can search and replace content in the generated changelog body using the `replacers` option. Replacers support searching via regular expressions with the `regex` option, and plain strings with the `search` option. Each replacer is applied in order.
+You can search and replace content in the generated changelog body using the `replacers` option. Replacers support searching via regular expressions. Each replacer is applied in order.
 
 ```yml
 replacers:
-  - regex: '/CVE-(\d{4})-(\d+)/g'
+  - search: '/CVE-(\d{4})-(\d+)/g'
```
toolmantim: What’s the story with changing this again?

lib/joi-extension.js
```diff
@@ -0,0 +1,25 @@
+const regexParser = require('regex-parser')
```
toolmantim: It looks like this file is specific to the `search` option but it’s named generically, so maybe it should stay inline?

But it’d be better if we could do without this altogether? Perhaps if we keep the `regex` option?

casz: Just preparing for if we needed more joi-extensions :sweat:

README.md
casz: I voiced my opinion in the #185 partly based on what I already knew would happen in the schema validation.

If you have a look at the code in #185 and my comment https://github.com/toolmantim/release-drafter/pull/185/files#r279162469

This is the very reason for keeping with the name `search`.
When the developer also referred to it as `search` inside the code.

As I said it increases the complexity for both users and developers.
Right now we have a simple function that can handle the use case for when a string is in regex representation and when it is a simple string.

With the regex kept in the schema, I would have to do something like this.
```
   customReplacers.forEach(({ search, regex, replace }) => {
      if (regex) search = regex
      str = str.replace(search, replace)
    })
```
Or add to the schema logic by using `xor` and/or `rename`

This is what had me confused in #185 changing the schema is not OK.
https://github.com/toolmantim/release-drafter/blob/5b573591f9055ce2280b41f5fb1b35beb6350201/lib/template.js#L34-L45

lib/joi-extension.js
casz: Keeping the regex option, I would still have to coerce the object into regex.
Same goes for the search.

I prefer to have Joi do the object conversion since it provides that option unless you favor otherwise.

Yes, I can validate against strings and convert them after validation.
But why not do it inside the schema validation?

casz:
I have no preference Joi vs. AJV
Joi can also be converted to JSON schema.
https://www.npmjs.com/package/joi-to-json-schema

README.md
casz: Makes copy-pasting into code harder :laughing:

README.md
toolmantim: I'm confused only because I thought changing it to be explicit (aka having two different config keys) was your idea? https://github.com/toolmantim/release-drafter/pull/185#discussion_r278704617 😁

toolmantim: I think there might have been a misunderstanding, in that both @halkeye and myself assumed you wanted an explicit parameter (and I just trusted you both had talked through the design enough) 😁

casz:
The main benefit of Joi as I see it, it does not mutate the original object it clones it and returns it after validation.

lib/joi-extension.js
toolmantim: Sorry, I wish GitHub reviews would let you comment on a whole file. I think I confused things by not making it clear I didn't mean to refer to this particular `require` statement, more just this whole file.

lib/schema.js
```diff
@@ -0,0 +1,92 @@
...
+  replace: Joi.string()
+    .allow('')
+    .required()
+})
+
+const fields = {
+  branches: Joi.array()
+    .items(Joi.string())
+    .single()
+    .description(
```
toolmantim: Are these descriptions used in the JOI error output anywhere? I'm a bit worried about the duplication between here and the readme, and if they're actually output then perhaps we can just leave them off?

README.md
casz: My earlier comment was against adding another property: 

https://github.com/toolmantim/release-drafter/pull/185#discussion_r278671766

> I'd prefer not adding another property and instead suggest you look at [regex-parser](https://www.npmjs.com/package/regex-parser)
> 
> Should handle parsing the regex in both formats only concern is to take the regex you get and add the default `g` flag

lib/schema.js
casz: Used in the schema output.

README.md
toolmantim: Ah I'm sorry, I don't think that came across very clearly to me. I don't think myself or @halkeye cared too much, and actually thought this was what you wanted. It's not a big deal to change it!

Want to open a separate PR just to change it to use `search`? And I'll update the readme too (I tried my best to simplify/explain the separate `regex`/`search` sub-properties.

lib/joi-extension.js
toolmantim: To make sense as a generic standalone file, I think you might need to move the `required()` and `error()` lines back into main schema definition, and leave this file to just handle the type coercion.

lib/joi-extension.js
```diff
@@ -0,0 +1,25 @@
...
+      .object()
+      .type(RegExp)
+      .required()
+      .error(() => '"search" is required and must be a regexp or a string'),
+    // eslint-disable-next-line no-unused-vars
+    coerce(value, state, options) {
+      let regex = value
+      if (typeof value !== 'string') {
+        return regex
+      } else if (value.match(/^\/.+\/[gmixXsuUAJ]*$/)) {
```
toolmantim: Did you try letting `regex-parser` do all of this? It's mostly the same, though it doesn't set `'g'` as a default flag in the plain string case:

https://github.com/IonicaBizau/regex-parser.js/blob/master/test/integration/node-regex-parser-test.js

casz: As mentioned in #185 the problem is `my.dev.com` will match to `my-dev-com` and most likely not what you want from a string search :laughing:

README.md
casz: Feel free to open a PR. I will rebase my changes on top.

lib/joi-extension.js
toolmantim: Ah, I see now! Such as shame to have to reimplement this regexp pattern, because it mostly exists inside `regexp-parser` already.

It'd be more useful if `regexParser("my.dev.com") === false` rather than turning it into `/my.dev.com/`. Ah well!

toolmantim:
Cool! The main priority from a UX POV is better config error messages (although they'll only be visible to Github Actions users, as the log for the GitHub App is only visible to me ☹️), and from a dev POV it's simplifying/centralising the config parsing. JOI sounds good over JSON Schema 👌🏼

lib/joi-extension.js
casz: We could say we only support `/regex/` syntax

toolmantim:
Speaking of giving people a nice error message when their config is broken, there's https://github.com/dessant/probot-messages but it doesn't seem to be able to close the issue once the config is fixed. It might be a easier just to reimplement something ourselves.

lib/schema.js
```diff
@@ -0,0 +1,92 @@
...
+  replacers: Joi.array()
+    .items(searchReplace)
+    .description('Search and replace content in the generated changelog body.'),
+
+  categories: Joi.array()
+    .items(category)
+    .description('Categorize pull requests using labels.'),
+
+  template: Joi.string()
+    .required()
```
casz: In the eyes of the one using the JSON schema outside the application, the template is not required. Since they could be using probot's `_extends` feature

lib/joi-extension.js
casz: After toying with schema generation, I decided not to let the schema do type coercion since the generator could not handle that case.

casz:
All that is left is to add more schema tests and add the missing bits to the schema so it can be submitted.

casz:
@TimonVS I merged your sorting/direction into this PR

README.md
casz: The readme in this PR seems fine.

README.md
```diff
@@ -139,27 +139,27 @@ exclude-labels:
...
+If your project doesn’t follow [Semantic Versioning](https://semver.org) you can still use Release Drafter, but you may want to set the `version-template` option to customize how the `$NEXT_{PATCH,MINOR,MAJOR}_VERSION` environment variables are generated.
 
-For example, if your project doesn't use patch version numbers, you can set `version-template` to `$MAJOR.$MINOR`. If the current release is version 1.0, then `$NEXT_MINOR_VERSION` will be `1.1`.
+For example, if your project doesn’t use patch version numbers, you can set `version-template` to `$MAJOR.$MINOR`. If the current release is version 1.0, then `$NEXT_MINOR_VERSION` will be `1.1`.
 
 ## Replacers
 
-You can search and replace content in the generated changelog body using the `replacers` option. Replacers support searching via regular expressions with the `regex` option, and plain strings with the `search` option. Each replacer is applied in order.
+You can search and replace content in the generated changelog body using the `replacers` option. Replacers support searching via regular expressions. Each replacer is applied in order.
```
toolmantim:
```suggestion
You can search and replace content in the generated changelog body, using regular expressions, with the `replacers` option. Each replacer is applied in order.
```

README.md
toolmantim: Not quite, it reads a bit strangely as it was structured for the old syntax. I’ve just added a suggested edit though 👍🏼

casz:
Hopefully, that cleans up the changes. I decided to keep validateReplacers as it will log if there is trouble with converting them to regex.

lib/schema.js
```diff
@@ -0,0 +1,109 @@
...
+const { SORT_DIRECTIONS } = require('./sort-pull-requests')
+
+const category = Joi.object().keys({
+  title: Joi.string().required(),
+  label: Joi.string()
+})
+
+const searchReplace = Joi.object().keys({
+  search: Joi.string()
+    .regex(/^\/.+\/[gmixXsuUAJ]*$|.+/)
```
toolmantim: Given this has `.+`, I wonder what benefit there is to having the first part of the pattern (or defining any `regex()` at all?)

casz: None, I was toying with the idea :laughing:

toolmantim: 😆

generate-schema.js
```diff
@@ -0,0 +1,8 @@
+const convert = require('joi-to-json-schema')
+const fs = require('fs')
+const { schema } = require('./lib/schema')
+
+const theSchema = convert(schema())
+// theSchema.$id = ''
```
TimonVS: This comment can be removed, I assume? :)

casz: I assume we have some fixes so we can publish the schema. That's why I left it in 😊

lib/log.js
```diff
@@ -1,10 +1,12 @@
-module.exports = ({ app, context, message, info }) => {
+module.exports = ({ app, context, message, info, error }) => {
   const repo = context.payload.repository
   const prefix = repo ? `${repo.full_name}: ` : ''
   const logString = `${prefix}${message}`
 
   if (info) {
     app.log(logString, info)
```
casz: This seems unused?

oleg-nenashev:
Would be great to get it landed. Thanks @casz !

casz:
@TimonVS I'd appreciate some help landing this.

Apart from the merge conflict and adding more tests (which can be done separately)

We need to add the missing new types.
This should be easy to land and publish the schema.

Do you see any other issues?

casz:
Should we use a different check to model/validate schema?

index.js
```diff
@@ -3,37 +3,26 @@ const { isTriggerableBranch } = require('./lib/triggerable-branch')
...
+      const repoConfig = (await getConfig(context, configName)) || {}
+      const { error, value } = schema(context).validate(repoConfig)
+      if (error) {
+        throw error
+      }
+      config = value
+    } catch (error) {
+      log({ app, context, error, message: 'Invalid config file' })
+      return
     }
```
TimonVS: I'd like to extract this to a separate module, just like https://github.com/release-drafter/release-drafter/pull/338. Though I can also do this after this is merged :)

TimonVS:
@casz I just took another look and all looks fine to me. How can I assist you further in getting this shipped?

> Should we use a different check to model/validate schema?

What do you mean exactly?

casz:
I meant different tech 😂 typo

TimonVS:
So like an alternative to Joi? Are we missing anything in Joi? If not I think it is fine as is.

casz:
I'll try and fix this PR up than.
Do you want to merge this first or #338 ?

casz:
you should have permissions to add any commits

TimonVS:
Then I will merge #338 first :)

casz:
Attempting to fix up the PR 😓 It's not easy 😆

index.js
```diff
@@ -8,23 +8,24 @@ const core = require('@actions/core')
...
     const config = await getConfig({
       app,
       context,
       getConfig: require('probot-config')
+    }).catch(error => {
+      log({ app, context, error, message: 'Invalid config file' })
+      exitEarly = true
     })
 
+    if (exitEarly) return
```
casz: not sure this is the best way to exit early.

Would you prefer try and catch block, problem is the config object is accessed below.

index.js
```diff
@@ -8,23 +8,24 @@ const core = require('@actions/core')
...
+
     // GitHub Actions merge payloads slightly differ, in that their ref points
     // to the PR branch instead of refs/heads/master
     const ref = process.env['GITHUB_REF'] || context.payload.ref
 
     const branch = ref.replace(/^refs\/heads\//, '')
 
-    if (!config.template) {
```
casz: added as a required object in the Joi schema so it should throw an error if deep merge does not contain `template`

lib/config.js
```diff
@@ -23,12 +26,18 @@ module.exports.getConfig = async function getConfig({
...
+  })
+
+  schema(context).validate(repoConfig)
 
-  const config = await fetchConfig(context, CONFIG_NAME, defaults)
+  const { error, value } = schema(context).validate(repoConfig)
+  const config = value
 
   try {
     config.replacers = validateReplacers({
```
casz: I'd like to explore Joi in a separate PR to see if it can handle the case for us.

lib/config.js
```diff
@@ -40,7 +49,128 @@ module.exports.getConfig = async function getConfig({
...
   return config
 }
+
+const category = Joi.object()
+  .keys({
+    title: Joi.string().required(),
+    label: Joi.array()
+      .items(Joi.string())
+      .single(),
+    labels: Joi.array()
```
casz: Not sure we want to keep both

lib/releases.js
```diff
@@ -188,8 +188,8 @@ module.exports.generateReleaseInfo = ({
   )
 
   if (lastRelease) {
-    let name = config['name-template'] || ''
-    let tag = config['tag-template'] || ''
+    let name = config['name-template']
```
casz: moved the default to the schema

lib/sort-pull-requests.js
```diff
@@ -11,16 +11,6 @@ const SORT_DIRECTIONS = {
 module.exports.SORT_BY = SORT_BY
 module.exports.SORT_DIRECTIONS = SORT_DIRECTIONS
 
-module.exports.validateSortBy = sortBy => {
```
casz: handled by joi

test/index.test.js
```diff
@@ -1222,4 +1222,46 @@ Previous tag: ''
...
+      getConfigMock('config-with-schema-error.yml')
+
+      const payload = require('./fixtures/push')
+
+      await probot.receive({
+        name: 'push',
+        payload
+      })
+      expect(output.length).toBe(3)
+      expect(output[1].msg).toMatch('Bad replacer regex')
```
casz: Seems like we need to return earlier inside config.js I just hate added try and catch and early returns 😭

lib/config.js
casz: perhaps worth making a breaking change when we also have published the the JSON schema?

lib/config.js
```diff
@@ -40,7 +49,128 @@ module.exports.getConfig = async function getConfig({
...
+  categories: Joi.array()
+    .items(category)
+    .description('Categorize pull requests using labels.'),
+
+  template: Joi.string()
+    .required()
+    .description('The template for the body of the draft release.')
+}
+
+const schema = context => {
```
TimonVS: It might be better to move the schema (and related objects) to a separate module to keep this module readable.

lib/config.js
```diff
@@ -23,12 +26,18 @@ module.exports.getConfig = async function getConfig({
...
-    ...DEFAULT_CONFIG,
-    branches: context.payload.repository.default_branch
-  }
+  const repoConfig = await fetchConfig(
+    context,
+    CONFIG_NAME,
+    DEFAULT_CONFIG
+  ).catch(error => {
+    throw error
+  })
```
TimonVS: I'm wondering, is it necessary to add this? In my mind a rejected promise will throw when using `await`, right? I could very well be wrong though.

lib/config.js
TimonVS: We keep can both in the schema.

lib/config.js
```diff
@@ -40,7 +49,128 @@ module.exports.getConfig = async function getConfig({
...
-  config['sort-direction'] = validateSortDirection(config['sort-direction'])
+  if (error) throw error
 
   return config
 }
+
+const category = Joi.object()
+  .keys({
+    title: Joi.string().required(),
+    label: Joi.array()
```
TimonVS: Currently, `label` can also be a `string`, does this account for that?

lib/config.js
```diff
@@ -40,7 +49,128 @@ module.exports.getConfig = async function getConfig({
...
+const schema = context => {
+  const default_branch = _.get(
+    context,
+    'payload.repository.default_branch',
+    'master'
+  )
+  return Joi.object().keys({
+    branches: fields.branches.default(default_branch),
+    'change-template': fields['change-template'].default(
+      '* $TITLE (#$NUMBER) @$AUTHOR'
```
TimonVS:
```suggestion
     DEFAULT_CONFIG['change-template']
```

index.js
casz: Sure thing

lib/config.js
casz: the `.single()` accounts for that

lib/config.js
casz: nicely spotted

lib/config.js
```diff
@@ -40,7 +49,128 @@ module.exports.getConfig = async function getConfig({
     config.replacers = []
   }
 
-  config['sort-direction'] = validateSortDirection(config['sort-direction'])
+  if (error) throw error
```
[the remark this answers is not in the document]
casz: Nicely spotted 👍

lib/config.js
casz: Fair enough

lib/config.js
casz: Not necessary

lib/config.js
casz: should they both allow single? perhaps labels should not?

lib/config.js
casz: Problem is actually with default config, you run into a circular dependency so. I'll most likely just move default_config to schema

lib/config.js
casz: actually not unnecessary without this tests fail :)

casz: NVM... 😖

lib/schema.js
```diff
@@ -1,7 +1,23 @@
 const _ = require('lodash')
 const Joi = require('@hapi/joi')
 const { SORT_BY, SORT_DIRECTIONS } = require('./sort-pull-requests')
-const { DEFAULT_CONFIG } = require('./config')
+
+const DEFAULT_CONFIG = Object.freeze({
```
casz: should it be moved to it's own module?

casz:
@TimonVS I'd like to know

> should both `label` and `labels` allow single? perhaps `labels` should not?

> should `DEFAULT_CONFIG` be moved to it's own module?

TimonVS:
> should both label and labels allow single? perhaps labels should not?

For now, let's just do what the code expects. That means that `label` is a single value and `labels` is a list.

> should DEFAULT_CONFIG be moved to it's own module?

That's probably a good idea because it doesn't really belong to the schema module.

lib/schema.js
```diff
@@ -0,0 +1,139 @@
...
+    .description(
+      'Replacement of text, it is possible to reference regex match groups by "$number" like $2 for second match group'
+    )
+})
+
+const fields = {
+  branches: Joi.array()
+    .items(Joi.string())
+    .single()
+    .description(
```
TimonVS: I've been thinking about the descriptions in the schema. While they do add value, I foresee them getting out of sync with the docs. For now I'd like to request that we remove all descriptions, the validation and JSON schema alone should be a big improvement :)

casz: sure thing, not a problem for me 👍

casz:
schema didn't make the decade release 😢
At least it gets to be one of the first PRs in the next decade!

Well big release of 2020 here we come!

schema.json
```diff
@@ -0,0 +1,131 @@
+{
+  "title": "JSON schema for Release Drafter yaml files",
+  "$id": "https://github.com/release-drafter/release-drafter/blob/master/schema.json",
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
TimonVS: I'm by no means an expert in JSON Schema, but I noticed the latest version is draft-07. Do you happen to know why [`joi-to-json-schema`](https://github.com/lightsofapollo/joi-to-json-schema) doesn't support draft-07?
Also if we're to conform to draft-04 I think we have to change `$id` to `id` as that is what changed in draft-06: https://json-schema.org/draft-06/json-schema-release-notes.html.

test/index.test.js
```diff
@@ -1222,4 +1229,46 @@ Previous tag: ''
...
+          msg: 'Invalid config file',
+          err: expect.objectContaining({
+            message: expect.stringContaining(
+              '"search" is required and must be a regexp or a string'
+            )
+          })
+        })
+      )
+    })
+    it('yaml exception', async () => {
```
TimonVS: Needs a blank line for readability :)

```suggestion
    
    it('yaml exception', async () => {
```

test/index.test.js
```diff
@@ -1222,4 +1229,46 @@ Previous tag: ''
       expect.assertions(1)
     })
   })
+  describe('config error handling', () => {
```
TimonVS:
```suggestion
  
  describe('config error handling', () => {
```

test/schema.test.js
```diff
@@ -0,0 +1,68 @@
...
+  describe('invalid', () => {
+    invalidConfigs.forEach(([example, message]) => {
+      it(`${JSON.stringify(example)} is invalid`, () => {
+        const { error } = schema(context).validate(example)
+        expect(error && error.toString()).toMatch(message)
+      })
+    })
+  })
+
+  it('current schema matches json schema', () => {
```
TimonVS: That is a neat little test that ensures that we keep the JSON schema up-to-date :)

TimonVS:
```suggestion
  it('current schema matches the generated JSON Schema, update schema with `yarn generate-schema`', () => {
```

package.json
```diff
@@ -11,6 +11,7 @@
     "now-start": "PRIVATE_KEY=$(echo $PRIVATE_KEY_BASE64 | base64 -d) npm start",
     "test": "jest",
     "test:watch": "jest --watch --notify --notifyMode=change --coverage",
+    "schema": "node ./generate-schema.js print",
```
TimonVS: A bit longer to type, but more clear I think :)

```suggestion
    "generate-schema": "node ./generate-schema.js print",
```

generate-schema.js
```diff
@@ -0,0 +1,22 @@
+// joi to json schema currently does not support v16 of joi
```
TimonVS: Could you move this file to `./bin/generate-schema.js`?

TimonVS:
```suggestion
// joi-to-json-schema currently does not support v16 of Joi (https://github.com/lightsofapollo/joi-to-json-schema/issues/57)
```

generate-schema.js
```diff
@@ -0,0 +1,22 @@
...
+
+const theSchema = {
+  title: 'JSON schema for Release Drafter yaml files',
+  $id:
+    'https://github.com/release-drafter/release-drafter/blob/master/schema.json',
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  ...convert(schema())
+}
+
+// template is only required after deep merged, should not be required in the JSON schema
```
TimonVS: I don't completely understand what this means. Could you elaborate?

index.js
```diff
@@ -8,23 +8,24 @@ const core = require('@actions/core')
...
+    try {
+      config = await getConfig({
+        app,
+        context,
+        getConfig: require('probot-config')
+      })
+    } catch (error) {
+      log({ app, context, error, message: 'Invalid config file' })
+      return
+    }
```
TimonVS: Would it be an idea to put the logging inside `getConfig` so that we can get rid of the `try/catch` here? That would make code a little more readable :)

lib/config.js
```diff
@@ -1,35 +1,21 @@
...
-    branches: context.payload.repository.default_branch
-  }
+  const repoConfig = await fetchConfig(context, CONFIG_NAME, DEFAULT_CONFIG)
 
-  const config = await fetchConfig(context, CONFIG_NAME, defaults)
+  schema(context).validate(repoConfig)
+
+  const { error, value } = schema(context).validate(repoConfig)
+  const config = value
+  if (error) throw error
```
TimonVS: Can we extract this to a function (`validateSchema`) that returns the config or throws?

generate-schema.js
```diff
@@ -0,0 +1,22 @@
+// joi to json schema currently does not support v16 of joi
+const convert = require('joi-to-json-schema')
+const fs = require('fs')
+const { schema } = require('./lib/schema')
+const args = process.argv.slice(2) || []
+
+const theSchema = {
```
TimonVS:
```suggestion
const jsonSchema = {
```

schema.json
casz: `joi-to-json-schema` currently only supports draft-04 as stated in the readme.

We could raise a issue asking for draft-07 support

test/schema.test.js
casz: I thought so myself 🗡

generate-schema.js
casz: If this was required in the JSON schema it would show in your IDE/editor with a warning that `template` is missing.
Which it is not since it could be located in another repo via `_extends`

See: https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/.github/release-drafter.yml

index.js
casz: How would you exit early?

casz:
if we ever have trouble with joi or joi-to-json-schema we can always switch to ajv for schema validation. However for now I think joi provides superior end user validation messages.

casz:
@TimonVS I think I covered it all except for https://github.com/release-drafter/release-drafter/pull/193#discussion_r362118693

> Would it be an idea to put the logging inside `getConfig` so that we can get rid of the `try/catch` here? That would make code a little more readable :)

Exactly how would to exit early without the try and catch?

I guess I could return an null object?
🤔

index.js
TimonVS: You're right, this seems fine actually. Try/catch is just really noisy, not much we can do about that :)

TimonVS: Oops you had already changed it. This is good!

index.js
```diff
@@ -14,17 +14,14 @@ module.exports = app => {
       getConfig: require('probot-config')
     })
 
+    if (config == null) return
```
TimonVS:
```suggestion
    if (config === null) return
```

TimonVS:
The code is all good now (I just had one small code suggestion)! I will take this for a test run tomorrow and if everything goes right it'll be ready to go!

casz:
I'd like to squash all the commits 😓

TimonVS:
Took it for a test run, a couple of things I noted that I think we should fix/improve:

- If the config validation fails, the Action should fail. I'm not sure if we currently do that, if not we can solve it in a separate PR as well. It's possible to implement this with: https://github.com/actions/toolkit/tree/master/packages/core#exit-codes

- Validation errors are not user friendly and actionable

With the following config:

```yml

template: |
  ## What’s Changed
  $CHANGES
yo: test
```

I get this error:

```
15:21:40.609Z  WARN probot: TimonVS/release-drafter-test-schema: Invalid config file
  ValidationError: "yo" is not allowed
      at Object.exports.process (/app/node_modules/@hapi/joi/lib/errors.js:202:19)
      at internals.Object._validateWithOptions (/app/node_modules/@hapi/joi/lib/types/any/index.js:763:31)
      at internals.Object.validate (/app/node_modules/@hapi/joi/lib/types/any/index.js:797:21)
      at validateSchema (/app/lib/schema.js:88:44)
      at getConfig (/app/lib/config.js:15:20)
```

Does Joi have more friendly error messages as well? I read something about [`ValidationError.details`](https://github.com/hapijs/joi/blob/master/API.md#validationerror)? 
We could also print the annotated version (`error.annotate()`), though it might be confusing that it will show the annotated version in JSON even though the config is in YAML (would there be a way to convert the annotated version to YAML?)

Another example of an error message which I think is not very user-friendly:

`ValidationError: child "categories" fails because ["categories" at position 1 fails because [child "title" fails because ["title" is required]]]`

What do you think?

casz:
To customize error messages you simply add an `.error` to them

casz:
Perhaps we could merge this and improve error message in a separate PR? This PR changes a few things that blocks other PRs or requires changes in this PR.

casz:
I think printing error.annotate would be fine.
it would be rather easy to write a replacers to convert json annotated to yaml annotated.

TimonVS:
I just validated that Release Drafter currently doesn't fail when there's no (or an invalid) config either, so this is ready to be merged. We can implement improvements in other PRs :)
I'll be merging this now!

TimonVS:
Thanks again for all the effort you put into this! The JSON schema autocompletion is going to be pretty neat.
We can release this pretty soon.

casz:
https://github.com/SchemaStore/schemastore/pull/895

casz:
Once added to the schema store it should work seamlessly with stuff like
https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml

Hopefully it will even work with the GitHub editor.

casz:
Seems like it is already published: http://schemastore.org/json/

casz:
AND WORKING 👏 

image

TimonVS:
> AND WORKING 👏

Awesome!

This has been released in [v5.6.0](https://github.com/release-drafter/release-drafter/releases/tag/v5.6.0).

PR #194, opened 2019-04-28
Remove ‘regex’ key from the ‘search’ option

casz:
In #185 we added a new `replacers` option, that had two separate sub-keys: `regex` and `search`. This simplifies the config down to a single `search` key that can handle both plain string values, and strings that look like a regular expression.

The `regex` key can no longer be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,11 +148,11 @@ For example, if your project doesn't use patch version numbers, you can set `ver
 
 ## Replacers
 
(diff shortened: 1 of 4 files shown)


toolmantim:
Why don’t we just PR changing the code in here too? I’d prefer PRs that only change a thing at a time, so it’d be nice to do each one on their own, with a well structured PR title and description for future people wanting to understand the changes.

casz:
There @toolmantim fixed the code as well :)

toolmantim:
Thanks @casz! I just updated the PR description too 🙌🏼

PR #219, opened 2019-05-25
Support specifying multiple labels for each category

### LINKED ISSUE #218: Categories: Add support of multiple labels per category
I am working on a global Release Drafter configuration in https://github.com/jenkinsci . The organization contains numerous repositories which use different tags due to historical reasons. I am interested to streamline labeling in our repositories, but it will take a while. 

For example, this is a configuration for bugfix tags:

```yaml
categories:
  - title: 🐛 Bug Fixes
    label: bug
  - title: 🐛 Bug Fixes
    label: fix
  - title: 🐛 Bug Fixes
    label: bugfix
  - title: 🐛 Bug Fixes
    label: regression  
```

From a UX perspective, it would be great to have support of multiple labels definition, e.g.

```yaml
categories:
  - title: 🐛 Bug Fixes
    labels: 
      - bug
      - fix
      - bugfix
      - regression  
```

casz:
fixes #218 

cc @oleg-nenashev

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -3,6 +3,7 @@ const compareVersions = require('compare-versions')
 const { getVersionInfo } = require('./versions')
 const { template } = require('./template')
 const log = require('./log')
+const _ = require('lodash')
 
 const UNCATEGORIZED = 'UNCATEGORIZED'
 
@@ -79,12 +80,20 @@ const contributorsSentence = ({ commits, pullRequests }) => {
 const getCategoriesConfig = ({ config }) => {
   const categoriesConfig = config.categories.reduce((acc, category) => {
     acc[category.label] = category
+    if (category.labels && category.labels.length) {
+      category.labels.map(label => (acc[label] = category))
+    }
     return acc
   }, {})
 
   const orderedCategories = [
     UNCATEGORIZED,
-    ...config.categories.map(category => category.label)
+    ..._.flatMap(config.categories, category => {
+      if (category.label) return category.label
+      else if (category.labels) {
+        return category.labels
+      }
+    })
   ]
 
   return [categoriesConfig, orderedCategories]
@@ -157,6 +166,7 @@ const generateChangeLog = (mergedPullRequests, config) => {
       return acc
     }, [])
     .join('')
+    .trim()
 }
 
 module.exports.generateReleaseInfo = ({
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -86,12 +86,9 @@ const getCategoriesConfig = ({ config }) => {
 
   const orderedCategories = [
     UNCATEGORIZED,
-    ..._.flatMap(config.categories, category => {
-      if (category.label) return category.label
-      else if (category.labels) {
-        return category.labels
-      }
-    })
+    ..._.flatMap(config.categories, category =>
+      category.label ? category.label : category.labels
+    )
   ]
 
   return [categoriesConfig, orderedCategories]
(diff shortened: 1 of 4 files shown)


lib/releases.js
```diff
@@ -157,6 +166,7 @@ const generateChangeLog = (mergedPullRequests, config) => {
       return acc
     }, [])
     .join('')
+    .trim()
```
casz: Removes all the empty newlines at the end, I noticed during testing if I added even more categories, it would result in more empty newlines.

lib/releases.js
```diff
@@ -79,12 +80,20 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+    if (category.labels && category.labels.length) {
+      category.labels.map(label => (acc[label] = category))
+    }
     return acc
   }, {})
 
   const orderedCategories = [
     UNCATEGORIZED,
-    ...config.categories.map(category => category.label)
+    ..._.flatMap(config.categories, category => {
```
casz: we could switch to Node JS 12 and gain [`Array.prototype.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

lib/releases.js
casz: properly better to check
```javascript
if (index + 1 !== orderedCategories.length && categorizedPullRequests[category].length) acc.push('\n\n')
```

casz: Never mind, trim() is the correct thing since it is done as part adding pull requests.

lib/releases.js
```diff
@@ -79,12 +80,20 @@ const contributorsSentence = ({ commits, pullRequests }) => {
 const getCategoriesConfig = ({ config }) => {
   const categoriesConfig = config.categories.reduce((acc, category) => {
     acc[category.label] = category
+    if (category.labels && category.labels.length) {
```
TimonVS: Would it make sense to arrayify `category.label` so that `category.label` and `category.labels` can be treated the same way? 
E.g.: `const labels = !!category.label ? [category.label] : category.labels`

lib/releases.js
```diff
@@ -79,12 +80,20 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+      category.labels.map(label => (acc[label] = category))
+    }
     return acc
   }, {})
 
   const orderedCategories = [
     UNCATEGORIZED,
-    ...config.categories.map(category => category.label)
+    ..._.flatMap(config.categories, category => {
+      if (category.label) return category.label
```
TimonVS: Would it make sense to use a ternary here? I always find them to be more expressive than `if/else` :)

casz: No, cause both can be undefined if you would so like 😢 I prefer the more expressive approach in this case

lib/releases.js
casz: I'd prefer to concatenate the two arrays 😆

lib/releases.js
TimonVS: But it's fine if both are `undefined` right? Then it will just return `undefined`, just like it does now. Or am I misunderstanding?

casz: You are right 🤦‍♂

TimonVS:
@casz I made some changes to your code, hope you don't mind, we can just revert them if you don't agree with the changes. It looks like it's good to go, @toolmantim what do you think?

casz:
@TimonVS Cheers, I don't mind at all! Thank you! LGTM.

casz:
@toolmantim readme is updated, and I fixed the tests :sweat_smile:

PR #226, opened 2019-05-31
Fix branch matching for GitHub Actions

toolmantim:
For some reason unbenkownst to me, Github Actions payload refs are different to the usual webhooks.

For example:

image

> 22:49:29.286Z  INFO probot: toolmantim/release-drafter: Ignoring push. renovate/semver-6.x is not one of: master

This changes it to use the `GITHUB_REF` environment variable.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,7 +35,11 @@ module.exports = app => {
     })
     config['sort-direction'] = validateSortDirection(config['sort-direction'])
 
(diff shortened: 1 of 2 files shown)


toolmantim:
Oh, running the tests themselves in a GitHub Action means that env is populated 🤔 Will need a different approach.

PR #279, opened 2019-08-12
Added a GitHub App Manifest to help those deploying this app

wolfeidau:
This makes it easier to deploy using probot cli as it includes the permissions and events required.


toolmantim:
Thanks @wolfeidau! 💯

Issue #280, opened 2019-08-15
'Rollback' releases


JimNero009:
Hi!
I've been working on adding some functionality to release drafter to 'rollback' a release if a later CI system calls back to Github to mark something as failed. I've got some code on a branch here https://github.com/JimNero009/release-drafter/tree/unrelease-fork.

You can see the basic idea in https://github.com/JimNero009/release-drafter/blob/unrelease-fork/lib/undo.js#L5 -- this function runs on 'status' events, and deletes tags and puts releases back into draft states if there is a status failure.

Would you be interested in adding this kind of thing at all?

Cheers,
Jim

TimonVS:
Hey Jim, sorry for the late reply.
Great idea and thanks for working out a prototype. This is not something we currently want to add to Release Drafter as that would broaden the scope a little too much, we want to focus on the generation of changelogs and not the release process itself per se. I do think this could be an awesome GitHub Action to build :)
I'm closing this issue for now, but do let me know if you have any further thoughts on this!

JimNero009:
Fair enough! Thanks for taking the time to look.

PR #286, opened 2019-08-16
[WIP] Add undo functionality

JimNero009:
As mentioned in #280, this is what I'm basically thinking when it comes to 'rollbacks'. Still a WIP as it needs automated tests, a code update and a local test, but just to give you a flavour.


TimonVS:
As mentioned in https://github.com/release-drafter/release-drafter/issues/280#issuecomment-560573460: awesome idea! But not something we want to focus on for Release Drafter currently.

PR #294, opened 2019-08-23
Add GitHub Actions V2 workflow example

JamesMGreene:
The second version of GitHub Actions, ["now with world-class CI/CD"](https://github.com/features/actions), was [announced on 2019-08-08](https://www.youtube.com/watch?v=E1OunoCyuhY). It is now in public beta and is slated for general availability to all users very soon on 2019-11-13.

As part of this upgrade, the original version of GitHub Actions workflows and its HCL-like syntax are also considered deprecated and support for them will be removed in the near future.

As such, it would be best to feature the new YAML workflow syntax in your Action's README example! 🎉 

---

If you don't like how any part of this reads, feel free to edit it directly. 👌

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,32 @@
 
 ## Usage
 
-You can use Release Drafter as a [GitHub Action](https://github.com/features/actions), by configuring your `.github/main.workflow` file with the following:
+### GitHub Actions V2
+
+You can use the [Release Drafter GitHub Action](https://github.com/marketplace/actions/release-drafter) in a [GitHub Actions V2 Workflow](https://help.github.com/en/articles/about-github-actions) by configuring a YAML-based workflow file, e.g. `.github/workflows/release-management.yml`, with the following:
+
+```yaml
+name: Release Management
+
+on:
+  push:
+    # branches to consider in the event; optional, defaults to all
+    branches:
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks James! Did you get a chance to test that suggested config, to ensure it works? I was just wondering, as I expected a few changes might have been required? I’d also be happy just to drop the Actions v1 support altogether.

JamesMGreene:
@toolmantim Yes, I am actively using that configuration in an Actions V2 Workflow. Although there are new changes available in Actions V2, it is also a guarantee that every V1 Action will remain compatible with V2. The main changes are all to the workflow definitions, not to the Action capabilities. 👌 

Workflow configuration:
https://github.com/github/learning-lab-components/blob/1b0453f9f0a2b2275fd39a8996594e798518d922/.github/workflows/release_management.yml#L13-L16

Example output:
https://github.com/github/learning-lab-components/commit/1b0453f9f0a2b2275fd39a8996594e798518d922/checks

I'll go ahead and remove the Actions V1 section from this PR. :+1:

toolmantim:
Amazing! Thanks so much @JamesMGreene 👍🏼 This is ready to rumble then.

PR #313, opened 2019-10-09
Prevent crash when a pull request author is deleted

### LINKED ISSUE #301: Release Drafter crashes with "Cannot read property 'login' of null in releases.js" when a pull request submitter is deleted
Hi all! We use the GitHub app to run Release Drafter in the Jenkins project. We use a [centralized configuration](https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.yml) which runs across 80 repositories ATM. Currently it works for the most of the repositories really well, except the diagnosability issues (I'd guess migration to GitHub Actions or self-hosting would be a right step for our project scale).

Yesterday we hit an issue with https://github.com/jenkinsci/gradle-plugin . The release drafter was enabled there yesterday, and the App was enabled for the repo. Everything looks fine AFAICT, but the changes are not being generated.

It would be great if somebody could take a look at the application logs so that we kno the root cause of the issue

CC @wolfs

oleg-nenashev:
Fixes: https://github.com/toolmantim/release-drafter/issues/301

To try: `uses: oleg-nenashev/jenkins-release-drafter@v5.2.0-jenkins-4` can be used to try it out for GitHub Actions users

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -62,7 +62,9 @@ const contributorsSentence = ({ commits, pullRequests }) => {
   })
 
   pullRequests.forEach(pullRequest => {
-    contributors.add(`@${pullRequest.author.login}`)
+    if (pullRequest.author) {
+      contributors.add(`@${pullRequest.author.login}`)
+    }
   })
 
   const sortedContributors = Array.from(contributors).sort()
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -150,7 +150,7 @@ const generateChangeLog = (mergedPullRequests, config) => {
             template(config['change-template'], {
               $TITLE: pullRequest.title,
               $NUMBER: pullRequest.number,
-              $AUTHOR: pullRequest.author.login
+              $AUTHOR: pullRequest.author ? pullRequest.author.login : "unknown"
             })
           )
           .join('\n')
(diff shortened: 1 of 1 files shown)


TimonVS:
Hey sorry for the late response and thanks for creating a PR!
There's just one small issue and that's that a missing `author.login` is replaced with `unknown`. This will end up looking like: `* Title of PR (#5) @unknown` because the `@` is defined in the `change-template`. `@unknown` is an actual user. Now I realize that maybe we should have included the `@` for the `$AUTHOR` variable, but that's in the past now and we can't change it without causing a breaking change.
I think we can go with either `\<unknown\>` (@\<unknown\>) or `ghost` (@ghost). What do you think?

lib/releases.js
```diff
@@ -150,7 +152,7 @@ const generateChangeLog = (mergedPullRequests, config) => {
             template(config['change-template'], {
               $TITLE: pullRequest.title,
               $NUMBER: pullRequest.number,
-              $AUTHOR: pullRequest.author.login
+              $AUTHOR: pullRequest.author ? pullRequest.author.login : "unknown"
```
timja:
```suggestion
              $AUTHOR: pullRequest.author ? pullRequest.author.login : "ghost"
```

oleg-nenashev:
fine with me

TimonVS:
After rereading the issue and the suggestion of @timja I think we should go with @ghost. I've already changed it in code so I'll be merging it now :)

PR #325, opened 2019-11-13
More readable release drafter yml example

casz:
no reverse wording of update release draft.

Let's call it release drafter instead of release management


TimonVS:
Thanks! This does indeed look better to me as well.

PR #337, opened 2019-12-20
set upload_url to output

gfreezy:


TimonVS:
Hey @gfreezy, thanks for contributing. I tested it and it works nicely. The only thing that caught my eye is that  [actions/upload-release-asset](https://github.com/actions/upload-release-asset) doesn't update a release asset, so when the workflow runs for a second time, the "Upload Release Asset" step fails. This obviously doesn't have anything to do with your PR, but I just wanted to make you (and possibly others) aware of this fact.

This is the config I used:

```yml
name: Release Drafter

on:
  push:
    # branches to consider in the event; optional, defaults to all
    branches:
      - master

jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@master
      - name: Build project
        run: |
          zip --junk-paths my-artifact README.md
      - name: Release Drafter
        id: release_drafter
        uses: gfreezy/release-drafter@master
        env:
      - name: Upload Release Asset
        id: upload-release-asset 
        uses: actions/upload-release-asset@v1.0.1
        env:
        with:
          upload_url: ${{ steps.release_drafter.outputs.upload_url }}
          asset_path: ./my-artifact.zip
          asset_name: my-artifact.zip
          asset_content_type: application/zip
```

TimonVS:
Fixes: #337

gfreezy:
hi @TimonVS , github asset is not allowed to be overwritten. We need to delete the old asset first before upload the same asset. However actions/upload-release-asset do not handle the deletion.

I have an actions/upload-release-asset [fork](https://github.com/gfreezy/upload-release-asset) which delete the assets that exist before upload the new assets.

PR #338, opened 2019-12-23
Put config in seperate module

TimonVS:
I put all things related to config in a separate module to make the entry file of the app more readable. I've also added tests to make sure it works as intended.


casz:
This will conflict somewhat with #193

Issue #332, opened 2019-11-24
prerelease option


julianladisch:
Currently release-drafter creates all draft releases with the default "prerelease: false".
It would be nice if there were a prerelease configuration option that defaults to false but can be set to true so that release-drafter creates draft releases with "prerelease: true".

TimonVS:
Great suggestion. We'll be implementing this soon.

pdcmoreira:
@TimonVS I was actually trying to implement that myself, together with the ability to specify a custom config name (I can give you a use-case where that can be very useful, if you want):
https://github.com/release-drafter/release-drafter/compare/master...pdcmoreira:support-prerelease

TimonVS:
This has been released in [v5.5.0](https://github.com/release-drafter/release-drafter/releases/tag/v5.5.0).

PR #349, opened 2019-12-24
Add "prerelease" config option

pdcmoreira:
Fixes: https://github.com/release-drafter/release-drafter/issues/332

Allows marking the draft as pre-release when it is created, throught the `prerelease` config option.


casz:
I'd like this to be merged so I can fit it into #193

PR #350, opened 2019-12-24
Allow specifying the config name through an input option

pdcmoreira:
This is useful if you want to create different drafts, depending on the triggered workflow.
You may want to create a pre-release draft in the development branch (targeting this branch) and have a `-alpha` suffix in the tag-template and, separately, create a release draft in the master branch (targeting it), without the suffix in the tag-template and maybe other different options.
So, in the workflow we can pass a different config name as an env variable when using release-drafter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ jobs:
       - uses: release-drafter/release-drafter@v5
         with:
           # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
-          config-name: my-config.yml
+          # config-name: my-config.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ jobs:
     steps:
       # Drafts your next Release notes as Pull Requests are merged into "master"
       - uses: release-drafter/release-drafter@v5
+        with:
+          # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
+          config-name: my-config.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,15 +2,20 @@ const { validateSchema } = require('./schema')
 const { DEFAULT_CONFIG } = require('./default-config')
 const log = require('./log')
 
-const CONFIG_NAME = 'release-drafter.yml'
+const DEFAULT_CONFIG_NAME = 'release-drafter.yml'
 
 module.exports.getConfig = async function getConfig({
   app,
   context,
+  configName,
   getConfig: fetchConfig
 }) {
   try {
-    const repoConfig = await fetchConfig(context, CONFIG_NAME, DEFAULT_CONFIG)
+    const repoConfig = await fetchConfig(
+      context,
+      configName || DEFAULT_CONFIG_NAME,
+      DEFAULT_CONFIG
+    )
 
     const config = validateSchema(app, context, repoConfig)
 
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -1286,39 +1286,10 @@ Previous tag: ''
       process.env['INPUT_CONFIG-NAME'] = 'config-name-input.yml'
 
       // Mock config request for file 'config-name-input.yml'
-      const getConfigScope = nock('https://api.github.com')
-        .get(
-          '/repos/toolmantim/release-drafter-test-project/contents/.github/config-name-input.yml'
-        )
-        .reply(200, {
-          type: 'file',
-          encoding: 'base64',
-          size: 5362,
-          name: 'config-name-input.yml',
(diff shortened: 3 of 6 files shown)


pdcmoreira:
Does it make more sense to load the name from `with` arguments instead of an `env` variable?

TimonVS:
Hey, thanks a lot for contributing! It would indeed make more sense to pass this as a `with` argument, you can use [`core.getInput`](https://github.com/actions/toolkit/tree/master/packages/core#inputsoutputs) for this :)

Could you add a test that demonstrates that it does indeed load from another location if configured? You can simulate Action inputs with environment variables (see https://github.com/actions/toolkit/blob/master/packages/core/src/core.ts#L67). And could you add documentation to the README?

README.md
```diff
@@ -25,6 +25,9 @@ jobs:
     steps:
       # Drafts your next Release notes as Pull Requests are merged into "master"
       - uses: release-drafter/release-drafter@v5
+        with:
+          # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
+          config-name: my-config.yml
```
TimonVS:
```suggestion
          # config-name: my-config.yml
```

Great way of explaining! I do want us to comment it out though as people are going to copy it and wonder why it doesn't work :)

test/index.test.js
```diff
@@ -1222,4 +1224,85 @@ Previous tag: ''
...
+    it('loads from another config path', async () => {
+      /*
+        Mock
+        with:
+          config-name: 'config-name-input.yml'
+      */
+      process.env['INPUT_CONFIG-NAME'] = 'config-name-input.yml'
+
+      // Mock config request for file 'config-name-input.yml'
+      const getConfigScope = nock('https://api.github.com')
```
TimonVS: I am assuming you didn't use `getConfigMock` because you wouldn't be able to do `getConfigScope.isDone()`? I've never used the `isDone` method so I'm not sure :)

test/index.test.js
```diff
@@ -1222,4 +1224,85 @@ Previous tag: ''
...
     })
   })
+
+  describe('with config-name input', () => {
+    it('loads from another config path', async () => {
+      /*
+        Mock
+        with:
+          config-name: 'config-name-input.yml'
+      */
```
TimonVS: Thanks for documenting this so clearly!

README.md
pdcmoreira: You're right! 😆

test/index.test.js
pdcmoreira: Actually, I didn't use `getConfigMock` because it statically requests `/repos/toolmantim/release-drafter-test-project/contents/.github/release-drafter.yml` and I actually want to assert that it requests the correct config file (based on the input).
`getConfigScope.isDone()` is just to check that the request was actually performed.
Basically I'm asserting two things: the config file request and the resulting output body for that configuration.

TimonVS:
Everything looks fine. I'll merge this once I have a release window :)

casz:
I'd prefer to merge #193 before this one 😅

pdcmoreira:
Rebased onto master and solved conflicts.

lib/config.js
```diff
@@ -2,15 +2,20 @@ const { validateSchema } = require('./schema')
 const { DEFAULT_CONFIG } = require('./default-config')
 const log = require('./log')
 
-const CONFIG_NAME = 'release-drafter.yml'
+const DEFAULT_CONFIG_NAME = 'release-drafter.yml'
```
casz: I think this should move to the `default-config.js` 🤔

test/index.test.js
```diff
@@ -1273,4 +1275,85 @@ Previous tag: ''
...
+        .get(
+          '/repos/toolmantim/release-drafter-test-project/contents/.github/config-name-input.yml'
+        )
+        .reply(200, {
+          type: 'file',
+          encoding: 'base64',
+          size: 5362,
+          name: 'config-name-input.yml',
+          path: '.github/config-name-input.yml',
+          content: encodeContent(
```
casz: use `getConfigMock` and rewrite it serve the purpose and avoiding the `release-drafter.yml` ?
Or Write a new function.

lib/config.js
pdcmoreira: The responsibilities seem to be good at the moment:

- `default-config.js`: export a default configuration object
- `config.js`: load and validate a configuration, falling back to a default one
- `index.js`: the controller - it asks `config.js` for the configuration to use, passing the context and the config name if present on the input

It doesn't make much sense to me that `index.js` will import `default-config.js` just to pass something to `config.js` that it already has access to (because it's already importing it).

casz: As I said minor nitpick 😄

pdcmoreira: Alright then! 👍
Let's keep this conversation open, someone else might want to give an opinion.

TimonVS:
This has been released in [v5.6.0](https://github.com/release-drafter/release-drafter/releases/tag/v5.6.0).

PR #355, opened 2019-12-29
Run tests on pull requests

casz:
tests are not running on a PR 😱


TimonVS:
Hah, good catch. We used to run tests with Travis as well but that probably broke when we migrated the repo to an organization. Thanks for fixing!

Issue #203, opened 2019-05-06
Release Target Setting


nchudleigh:
Would be really helpful if the Release Draft Target (optionally) was the latest commitish.

The two issues with using Master as the Target are:
1. PR merged to master in the time that someone is viewing the draft, the merged PR will not be documented on the release. 
2. If the release is modified to link to another tag, the tag will actually mark the current state of master.  So a release candidate, beta, or staging process has the potential to break.

It is a fairly easy thing to automate (https://developer.github.com/v3/repos/releases/#create-a-release)
![image](https://user-images.githubusercontent.com/5580824/57237599-8d4fd200-6ff5-11e9-8964-da7aff51bd1e.png)

If this gets a :+1: I can write and submit this PR myself!

toolmantim:
Thanks, sounds sensible. Any ideas for a configuration setting name?

Also, possibly related to #175?

nchudleigh:
No strong feelings on that, off the top of my head, `target` with configurable types as the value feels pretty clean and expandable.

```yml
target: 'branch' # this would be the default setting as it is now
target: 'commit' # this would target the latest commit to the branch
```
The only downside is this doesnt tie too nicely into the branches array that exists on the config currently, as you cannot specify a different setting per branch.

Without revisiting that though you would be introducing conflicting config.  Let me know what you think :)

nchudleigh:
@toolmantim Any further thoughts on this?

casz:
I would just perhaps just look for the last few tags and find one that matches `tag-template`

nchudleigh:
@casz On the implementation side this is pretty trivial yeah. 

More so asking for input on the settings. Do you have any thoughts on that?

casz:
This becomes rather complex as soon as you run anything besides trunk-based development.

you want the YAML syntax to both support `branches` and `tags` both in wildcard format.
Then to further complicate you want the find a commitish target based on the `tags` pushed and hopefully they match your `tag-template`
To complicate it, even more, you most likely want it to support multiple `tag-templates`
To absolutely further complicate it, you really want to support creating release-draft for multiple branches and multiple tags independent of each other.

jetersen:
Hmm the recent changes to master and the fact that we have `$RESOLVED_VERSION` see use case below or see the readme about [version-resolver](https://github.com/release-drafter/release-drafter#version-resolver):

[...]

Issue #175, opened 2019-04-10
How to configure this with gitflow?


thibmaek:
I tried, really. But I didn't find a way to create the release.

Our repo has base branch set to develop, but I only want to create releases for code ending up on master.

Relevant Release Drafter config: https://github.com/inthepocket/hubble-scripts/blob/develop/.github/release-drafter.yml

Tried first with merging remotes into master and pushing, then tried to merge via PR from develop to master, but the release is not showing up as a Draft when I go to the releases tab.

![image](https://user-images.githubusercontent.com/6213695/55890684-72836c80-5bb3-11e9-9841-fece289c705b.png)

casz:
Change your default branch to master?

Or perhaps we should provide a way to override the default branch.

thibmaek:
That would mean all the PRs point at master by default and checking out the repo across team members starts from master. Kind of defeats the purposes of gitflow I think

sullimander:
We would need the ability to set the release branch separate from the github default branch in order for us to use it.

TimonVS:
Can you walk me through the issue you're having in a little more detail? I can see that you used the `branches` option in your config, why doesn't that give the desired result?

StewartSnow:
Hey we have the same issue here - we follow gitflow and would prefer that all draft releases are created on the master branch, rather than default develop branch (as is common under gitflow). 
An additional config entry to indicate the branch that the release / tag should be drafted on would be most helpful.

thibmaek:
This is indeed the same problem I am encountering

rvanlaak:
In our gitflow-like scenario we have `master` setup as `develop`, and `production` as `master. What we see happening, is that release drafter nicely creates drafts for `master`. 

A problem we have (that sounds similar to the issue description) happens in the following scenario: 

1 - A release branch gets created, testing starts and could take several days
2 - PRs that in the meanwhile get merged in `master` (read: the gitflow develop) and do not make the release - as they are not in the release branch.
3 - The release branch gets merged in `production` (read: the gitflow master) and gets tagged.

The actual problem that happened in the meanwhile: the PRs that get merged in step 2 do not appear in the next release draft.

mkurz:
I think the problems you are facing can be solved with the `filter-by-commitish` config, I came up with this comment on how to change the `target_commitish` of a GitHub release.

mkurz:
You also might be interested in #1061

PR #359, opened 2019-12-30
target-branch config

pdcmoreira:
This config option allows setting the desired target branch for the release.
Also, if this option is set, only the releases targetting the same branch are taking into consideration.

This does not cover all the possible cases as commented on https://github.com/release-drafter/release-drafter/issues/203, but it covers the simplest one - targetting a specific branch - which might also be the most common one.
I think it also helps https://github.com/release-drafter/release-drafter/issues/175.


index.js
```diff
@@ -20,6 +20,8 @@ module.exports = app => {
 
     const branch = ref.replace(/^refs\/heads\//, '')
 
+    const targetBranch = config['target-branch']
```
casz: instead of adding a const I suggest consuming the value from the config object.

pdcmoreira: But then it will need to be consumed on https://github.com/release-drafter/release-drafter/blob/91823a0446bfe714fd7129436d7b09d1af78d056/index.js#L34-L38 and https://github.com/release-drafter/release-drafter/blob/91823a0446bfe714fd7129436d7b09d1af78d056/index.js#L72
Readability will be a little bit worse, I think.

casz: with schema in #193 this empty string would be the default

```suggestion
    const targetBranch = config['target-branch'] || ''
```

index.js
```diff
@@ -29,7 +31,11 @@ module.exports = app => {
       return
     }
 
-    const { draftRelease, lastRelease } = await findReleases({ app, context })
+    const { draftRelease, lastRelease } = await findReleases({
+      app,
+      context,
+      targetBranch
```
casz:
```suggestion
      config['target-branch']
```

index.js
```diff
@@ -62,7 +68,8 @@ module.exports = app => {
           tag_name: releaseInfo.tag,
           body: releaseInfo.body,
           draft: true,
-          prerelease: config.prerelease
+          prerelease: config.prerelease,
+          ...(targetBranch ? { target_commitish: targetBranch } : {})
```
casz: To improve readability, depends whether the rest api accepts empty string. Though that is easily testable.
```suggestion
          target_commitish: config['target-branch']
```

index.js
casz:
```suggestion
```

index.js
pdcmoreira: Since the behaviour of passing `null` or `''` to the REST service is undocumented in https://developer.github.com/v3/repos/releases/#create-a-release (or even https://octokit.github.io/rest.js/#octokit-routes-repos-create-release), I don't think it is wise to expect it to always fallback to the default value when such values are passed.

casz: If `config['target-branch']` instead would default to `context.payload.repository.default_branch` then it would follow the rest api standard.

TimonVS:
Thanks for the PR! I'll review this soon.

casz:
I'd prefer to merge #193 before this one 😅

pdcmoreira:
Rebased onto master, solved conflicts and added the config option to schema validation.

index.js
```diff
@@ -59,7 +63,9 @@ module.exports = app => {
           tag_name: releaseInfo.tag,
           body: releaseInfo.body,
           draft: true,
-          prerelease: config.prerelease
+          prerelease: config.prerelease,
+          target_commitish:
+            config['target-branch'] || context.payload.repository.default_branch
```
casz: the default should be handled by the schema.
```suggestion
            config['target-branch']
```

lib/schema.js
```diff
@@ -78,6 +78,10 @@ const schema = context => {
 
     template: Joi.string().required(),
 
+    'target-branch': Joi.string()
+      .allow(null)
+      .default(DEFAULT_CONFIG['target-branch']),
```
casz: See line 8-12
https://github.com/release-drafter/release-drafter/blob/5e307e1c8f08c923e0ae855e21a6faf64bce12ea/lib/schema.js#L8-L12

```suggestion
      .default(defaultBranch),
```

lib/default-config.js
```diff
@@ -11,7 +11,8 @@ const DEFAULT_CONFIG = Object.freeze({
   replacers: [],
   'sort-by': SORT_BY.mergedAt,
   'sort-direction': SORT_DIRECTIONS.descending,
-  prerelease: false
+  prerelease: false,
+  'target-branch': null
```
casz:
```suggestion
```

lib/schema.js
```diff
@@ -78,6 +78,10 @@ const schema = context => {
 
     template: Joi.string().required(),
 
+    'target-branch': Joi.string()
+      .allow(null)
```
casz:
```suggestion
```

casz:
Hmm the schema test should have failed (Mind looking into why?)

To regenerate the schema, run `yarn generate-schema`

pdcmoreira:
There is a `YAMLException` and a probot `WARN` when running tests, while loading the configuration from `config-with-yaml-exception.yml`, which seems to be intended, in order to test the schema validator. But all the tests are passing anyways.
The result appears to be exactly the same before and after running `yarn generate-schema`.

lib/schema.js
pdcmoreira: This would change the logic and introduce a breaking change.

When we're creating the draft, it's ok to fall back to the default branch, because the draft's target branch should be specified anyway:
https://github.com/release-drafter/release-drafter/blob/c642710689e15269b262c16e6ffd9eaa5740ba39/index.js#L67-L68

But when we're searching for releases, we should either:

- Filter by branch, if there is a `target-branch` config
- Not filter by branch, if there is no `target-branch` config (or it's null), taking any release from any branch into consideration to calculate the version (previous behaviour)

The branch filter is passed in to `findReleases`:
https://github.com/release-drafter/release-drafter/blob/c642710689e15269b262c16e6ffd9eaa5740ba39/index.js#L32

casz: Than add to properties, one for `filtered-branch` and one for `target_commitish`

To me this seems like going beyond the use case for release-drafter but I could be wrong.

TimonVS:
I haven't had the time to review this functionality just yet. This PR requires a little more consideration and testing then most other PRs as I want to be sure it benefits use cases explained in #203 and #175 without driving us in a corner where we have to make a breaking change in the future :)

pdcmoreira:
@TimonVS I have an use case that uses a branch for staging deployment (pre-releases) and a branch for production deployment (releases) and this feature allows for that kind of separation.

pdcmoreira:
Rebased onto master.
Following @casz suggestions, I gave it some thought and it wouldn't be worse if we have separate config options instead:
- target-commitish: The target_commitish to set in the draft;
- target-commitish-filter: Filter the releases considered for the next version calculation by target_commitish.

pdcmoreira:
@casz For some reason, the re-request review button does nothing. I've also requested a review from @TimonVS, since it's been a while.

say8425:
Is it something going on? I need this feature.

pdcmoreira:
Updated the whole thing.

jetersen:
Any chance you would like to solve the merge conflicts :)

pdcmoreira:
> Any chance you would like to solve the merge conflicts :)

@jetersen sure, but @TimonVS said this needs a lot more thought.

I don't see the point in keeping this PR updated until we specify the final solution.

rogerluan:
I'd love to see this released to use in my project 🚀  Is there anything I can help with here? Which specific comments above are pending discussion @pdcmoreira @TimonVS ?

jeffrey-kinesso:
I'm also looking for this feature! 👍

mshick:
Could `target-commitish` be added to `action.yml` as an input? I'd very much like this feature to incorporate via Actions.

jetersen:
feel free to resubmit PR with conflicts resolved.

PR #363, opened 2020-01-03
Override calculated version, tag and name with input

pdcmoreira:
Allow overriding the resulting version, tag and name throught input options.
This is useful for workflows scoped by branches (https://github.com/release-drafter/release-drafter/pull/359) that need to transition a version from one branch to another - GitFlow, for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,17 @@ jobs:
         with:
           # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
           # config-name: my-config.yml
+          # (Optional) override the calculated version
+          # version: '2.1.1'
+          # (Optional) override the resulting tag
+          # version: 'v2.1.1-alpha'
+          # (Optional) override the resulting name
+          # version: 'v2.1.1-alpha (Code Name: Foxtrot Unicorn)'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
+
+**Note:** If provided, the `version`, `tag` and `name` input options will be used (in this order) to extract a version number instead of calculating one.
 
 If you're unable to use GitHub Actions, you can use the Release Drafter GitHub App. Please refer to the [Release Drafter GitHub App documentation](docs/github-app.md) for more information.
 
--- a/README.md
+++ b/README.md
@@ -32,8 +32,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
 
-**Note:** If provided, the `version`, `tag` and `name` input options will be used (in this order) to extract a version number instead of calculating one.
-
 If you're unable to use GitHub Actions, you can use the Release Drafter GitHub App. Please refer to the [Release Drafter GitHub App documentation](docs/github-app.md) for more information.
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ jobs:
           # (Optional) override the resulting tag
           # tag: 'v2.1.1-alpha'
           # (Optional) override the resulting name
-          # version: 'v2.1.1-alpha (Code Name: Foxtrot Unicorn)'
+          # name: 'v2.1.1-alpha (Code Name: Foxtrot Unicorn)'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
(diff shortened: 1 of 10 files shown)


TimonVS:
Thanks for the PR @pdcmoreira. This is an awesome addition. I'm going to give it a test run!

README.md
```diff
@@ -25,10 +25,19 @@ jobs:
     steps:
       # Drafts your next Release notes as Pull Requests are merged into "master"
       - uses: release-drafter/release-drafter@v5
+        with:
+          # (Optional) override the calculated version
+          # version: '2.1.1'
+          # (Optional) override the resulting tag
+          # version: 'v2.1.1-alpha'
```
TimonVS:
```suggestion
          # tag: 'v2.1.1-alpha'
```

README.md
```diff
@@ -25,10 +25,19 @@ jobs:
...
+          # (Optional) override the resulting name
+          # version: 'v2.1.1-alpha (Code Name: Foxtrot Unicorn)'
```
TimonVS:
```suggestion
          # name: 'v2.1.1-alpha (Code Name: Foxtrot Unicorn)'
```

lib/versions.js
```diff
@@ -1,7 +1,10 @@
 const semver = require('semver')
 
 const splitSemVer = input => {
-  const version = semver.inc(input.lastVersion, input.inc, true)
+  const version =
+    !input.increment && input.inc === 'patch'
```
TimonVS: I think it's pretty clever to not increment the patch version when providing a version or tag with an input. But I feel like it may not be as intuitive for our users. What about adding a new template variable: `$INPUT_VERSION` and leaving the behavior of the patch version like it was before (e.g. if input version is `v2.2.1`, `$NEXT_PATCH_VERSION` will become `v2.2.2` and `$INPUT_VERSION` will be `v2.2.1`).

pdcmoreira: Yeah, good idea. I'll change that.

pdcmoreira:
Rebased onto master and separated the manual input version from the calculated ones.
@TimonVS Please check if that's what you had in mind.

pdcmoreira:
Rebased onto master and fixed a bug with the mocked env that was leaking into other tests.

pdcmoreira:
Rebased onto master.
By having the input version and the calculated version separated, I noticed that there was no way to create a generic template that would print the input version if available or the calculated one otherwise, so I've added a new `$RESOLVED_VERSION` variable that does just that.

TimonVS:
`$RESOLVED_VERSION` is a great addition @pdcmoreira! I'm testing it one last time and then I'll merge and release it.

TimonVS:
After testing this I noticed that this only works for newly created drafts. I think we should also propagate this through draft updates. I think this is technically a breaking change, but I would consider it a bug if we don't update the draft, so I think a minor version in this case would still be fine.

TimonVS:
@pdcmoreira @casz what do you think about propagating the release's name and tag changes in draft updates too? As mentioned in my previous comment, it's technically a breaking change, but I think it can be considered a bug fix because currently the release name and tag for draft releases won't update even though the input version might have changed. But, people might rely on this functionality to update the title or tag of a release manually without having Release Drafter override it with a draft update. What are your thoughts on this?

I have a reproduction for this issue here: https://github.com/TimonVS/rd-override-version. You can test this by having Release Drafter generate a draft first and then change the input version here: https://github.com/TimonVS/rd-override-version/blob/master/.github/workflows/release-drafter.yml#L16. You'll see that the version in the body is updated but that the version in the title and tag stay the same.

@pdcmoreira I rebased your PR btw so that I could test with the latest changes.

casz:
I honestly thought it was a bug but not one I cared to report 🐛😅 Yes I think it is a bug.

pdcmoreira:
True, I think it should be considered a bug, because if the calculated version changes, it stops being semver compliant.
But it's also true that it might stop fitting some use cases.
My suggestion would be to "fix" the "bug" and add an option to ignore version re-calculation on draft updates.

TimonVS:
Thanks for your input! I'll create a fix for it, once that's done we'll release this. I've temporarily removed the documentation in this PR so that I can merge it now without having incorrect documentation on the master branch.

jetersen:
@pdcmoreira I had issues with failing tests due to `mockedEnv` on my Windows box.
Seems to be related only to the `INPUT_CONFIG-NAME` `mockedEnv` 😓

jetersen:
The mockedEnv issue disappeared.

PR #364, opened 2020-01-03
Add name and version number to outputs

zackify:
I want to take the version outputted from the release, and use it for an iOS and Android app's version number

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,9 +81,12 @@ module.exports = app => {
 
 function setActionOutput(releaseResponse) {
   const {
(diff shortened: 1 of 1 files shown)


TimonVS:
Hey @zackify! That's a fantastic idea. I'd think that you're looking for `tag_name` instead of the name of a release though?
If you choose to expose the `tag_name` (which I think would be a very useful output), we should not remove the `v` as you can do that in your own workflows using https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#format.

jetersen:
superseded by #538

PR #368, opened 2020-01-05
Config schema validation improvements

TimonVS:
Implemented some improvements to the config schema validation from #193:

- Set Action status to failed if an invalid config is supplied
- Make schema validation more forgiving by allowing unknown properties
- Log JSON representation of config with error annotations

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -21,10 +21,13 @@ module.exports.getConfig = async function getConfig({
     log({ app, context, error, message: 'Invalid config file' })
 
     if (error.isJoi && runnerIsActions()) {
-      console.log(
-        'Config with error annotations (JSON representation with defaults applied)'
-      )
-      console.log(error.annotate())
+      log({
+        app,
+        context,
+        message:
+          'Config with error annotations (JSON representation with defaults applied)\n' +
+          error.annotate()
+      })
     }
 
     if (runnerIsActions()) {
--- a/lib/config.js
+++ b/lib/config.js
@@ -3,6 +3,7 @@ const { validateSchema } = require('./schema')
 const { DEFAULT_CONFIG } = require('./default-config')
 const log = require('./log')
 const { runnerIsActions } = require('./utils')
+const Table = require('cli-table')
 
 const CONFIG_NAME = 'release-drafter.yml'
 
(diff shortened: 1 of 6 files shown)


TimonVS:
@casz I'd very much appreciate your review :)

lib/config.js
```diff
@@ -17,6 +19,20 @@ module.exports.getConfig = async function getConfig({
     return config
   } catch (error) {
     log({ app, context, error, message: 'Invalid config file' })
+
+    if (error.isJoi && runnerIsActions()) {
```
TimonVS: This is what it looks like now:

image

Not perfect, because the config is in YAML and the config with error annotations is in JSON.

Alternatively I could log a table like so:

```
| Property   | Error message                               |
| ---------- | ------------------------------------------- |
| prerelease | "prerelease" must be a boolean              |
| sort-by    | "sort-by" must be one of [merged_at, title] |
```

TimonVS: I'm convinced the JSON representation is confusing and noisy, I'm going to change it.

package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@actions/core": "^1.2.0",
     "@hapi/joi": "15.1.1",
+    "cli-table": "^0.3.1",
```
casz: Perhaps better to use a more managed package: https://www.npmjs.com/package/cli-table3

lib/config.js
TimonVS: As a table:

image

Much better if I say so myself :)

lib/schema.js
```diff
@@ -85,8 +85,11 @@ const schema = context => {
 module.exports.schema = schema
 
 const validateSchema = (app, context, repoConfig) => {
-  const { error, value } = schema(context).validate(repoConfig)
-  const config = value
+  const { error, value: config } = schema(context).validate(repoConfig, {
+    abortEarly: false,
```
casz: Nice, I was thinking of adding this but the PR was already pretty big.

package.json
TimonVS: Good one! I'll update that tomorrow.

casz: If you give me commit access to release-drafter/release-drafter I can contribute it as a co-maintainer 🚀

casz:
I would have loved to see an test for printing the table, as a reviewer it is hard to understand the final outcome after the code cleanup 🤔

TimonVS:
> I would have loved to see a test for printing the table, as a reviewer it is hard to understand the final outcome after the code cleanup 🤔

You're right. It'll be a bit weird to test this though. I will see if I can add a test in the coming days.

PR #376, opened 2020-01-08
use probot adapter - @zeit/ncc

jetersen:


casz:
Unable to get node_modules vendoring working

Failing on finding the debug module for some odd reason even after adding the module.
https://github.com/casz/test-release-drafter-action-adapter/runs/380338648

casz:
zeit@ncc not working either
https://github.com/casz/test-release-drafter-action-adapter/runs/380346142

casz:
Might be able to rework it, https://github.com/npm/node-semver/issues/306

casz:
we have green build 🙌 
https://github.com/casz/test-release-drafter-action-adapter/commit/33d89556eb67c2b03bfd9c9cb1dbbcc85748a90a/checks?check_suite_id=403506647

casz:
You can follow a model, much like how actions/cache does it, where they only add dist folder on release branches: https://github.com/actions/cache/tree/releases/v1/dist

casz:
Instead of a minute to release drafter it is mere seconds now:

https://github.com/release-drafter/release-drafter/commit/742103fb215cffdf7e75a201365b1d0e3e16a959/checks?check_suite_id=398471655
vs
https://github.com/casz/test-release-drafter-action-adapter/runs/392579834

ajcann:
I've been using this 3486aa7f1059bfbfa572c1092b74fb1bf32ee1c8 for actions across a dozen or so repos for the last day with great results - average run time has been seconds rather than typically over a full minute. This is particularly helpful with some of my teams tendency to get a bit quick-clicky and publish a release before notes have updated after merging a PR. Hoping this makes it into mainline!

casz:
Conflicts resolved.

TimonVS:
Awesome work @casz!

lib/commits.js
```diff
@@ -1,6 +1,8 @@
 const log = require('./log')
 const paginate = require('./pagination')
-const _ = require('lodash')
+const get = require('lodash/get')
+const uniqBy = require('lodash/uniqBy')
+const flatten = require('lodash/flatten')
```
TimonVS: Are these changes required to get it to work or to reduce the compiled output?

lib/versions.js
```diff
@@ -1,20 +1,25 @@
-const semver = require('semver')
+const inc = require('semver/functions/inc')
+const major = require('semver/functions/major')
+const minor = require('semver/functions/minor')
+const patch = require('semver/functions/patch')
+const parse = require('semver/functions/parse')
+const coerce = require('semver/functions/coerce')
```
TimonVS: Same question here, is this required to make it work? If so, do you know why that is?
I'd prefer to keep just one import because `semver.coerce` is a lot clearer in its intent than just `coerce` and if it just takes a `require('semver')` to break the build then I fear that will happen sooner rather than later. Please correct me if I'm wrong though :)

lib/commits.js
casz: I am trying to reduce the compiled output here yes.

lib/versions.js
casz: `require('semver')` does not load functions at the same time it adds a lot of code to ncc output.
you can use `require('semver/preload')` but again it adds even more to the ncc output.

casz: Feel free to test it out locally :)

TimonVS:
This is really great. The one thing we need to figure out is how to run PR branches. I test PR branches quite frequently, for this to continue working we need to be able to compile on the fly. Maybe we can add a configuration option that will take care of that?

casz:
if you use husky (not everybody does) we could setup a precommit or prepush that look for changes in src and if there is changes on dist?

TimonVS:
True, but that would cause every PR to have merge conflicts on the dist output. I'm seeing to solutions:

1. A config option `build: true` that runs a script before running Release Drafter. The script runs `yarn install` and `yarn build`.
2. Add a GitHub Action workflow that builds every PR and pushes the build to a branch or tag and comments on the PR with a branch or tag that can be used to test with. We can push these tags or branches to a separate repo: `release-drafter/test-builds`.

casz:
second option could be possible 🤔
There are enough github actions to get it done 😅

casz:
@TimonVS what about keeping the docker based action in the master branch?
And than on release generate the dist folder and have that as the action.yml?

casz:
See 
https://github.com/GitTools/actions/blob/master/gitversion/setup/action.yml
https://github.com/GitTools/actions/blob/master/gitversion/execute/action.yml
https://github.com/GitTools/actions/blob/master/action.yml

TimonVS:
> See
> https://github.com/GitTools/actions/blob/master/gitversion/setup/action.yml
> https://github.com/GitTools/actions/blob/master/gitversion/execute/action.yml
> https://github.com/GitTools/actions/blob/master/action.yml

This makes me think we can put the old `action.yml` and Dockerfile in a separate directory and use it as `release-drafter/release-drafter/docker@v5`?

casz:
@TimonVS I think a separate folder is the easiest.

ajcann:
Any chance this will make it into the mainline? We've been using 3486aa7f1059bfbfa572c1092b74fb1bf32ee1c8 for the last few months across ~50 repos with great performance.

ianwalter:
We did have a couple hiccups while using this branch where the release notes didn't get updated even though they should have. I think maybe there was a failed request to the GitHub API for whatever reason but no non-zero exit code and no errors were logged, only:
```
Run toolmantim/release-drafter@3486aa7f1059bfbfa572c1092b74fb1bf32ee1c8
  env:
    GITHUB_TOKEN: ***
22:44:57.855Z  WARN probot:
  Probot is unable to retrieve app information from GitHub for event subscription verification.
  
  If this error persists, feel free to raise an issue at:
    - https://github.com/probot/probot/issues
22:44:58.249Z  INFO probot: org/app: Found 14 releases
22:44:58.250Z  INFO probot: org/app: No draft release found
22:44:58.250Z  INFO probot: org/app: Last release: v0.4.2
22:44:58.251Z  INFO probot: org/app: Fetching all commits for branch master since 2020-03-05T14:58:45Z
22:44:58.524Z  INFO probot: org/app: Creating new draft release
```
After re-running it with a blank commit some hours later, the release notes were updated.

Jetersen:
I'll see if I can clear the conflicts and clear up another action setup for using the docker approach.
Dependency updates should resolve the non 0 exit code.

index.js
```diff
@@ -88,7 +88,7 @@ function setActionOutput(releaseResponse) {
   const {
     data: { id: releaseId, html_url: htmlUrl, upload_url: uploadUrl }
   } = releaseResponse
-  core.setOutput('id', releaseId)
+  core.setOutput('id', releaseId.toString())
```
Jetersen: 😱 Cannot believe the tests did not catch this!
This was not caught because we did not return a reply to the release drafting post 😭

Caught it on the action lucky and is now caught by tests.

```
    TypeError: (s || "").replace is not a function

      89 |     data: { id: releaseId, html_url: htmlUrl, upload_url: uploadUrl }
      90 |   } = releaseResponse
    > 91 |   core.setOutput('id', releaseId)
         |        ^
      92 |   core.setOutput('html_url', htmlUrl)
      93 |   core.setOutput('upload_url', uploadUrl)
      94 | }

      at escapeData (node_modules/@actions/core/src/command.ts:78:6)
      at Command.toString (node_modules/@actions/core/src/command.ts:71:31)
      at Object.issueCommand (node_modules/@actions/core/src/command.ts:25:28)
      at Object.setOutput (node_modules/@actions/core/src/core.ts:84:3)
      at setOutput (index.js:91:8)
      at setActionOutput (index.js:83:5)
```

of course once could argue that escapeData should handle escaping numbers...

docker/action.yml
```diff
@@ -0,0 +1,8 @@
+name: 'Release Drafter'
+description: 'Drafts your next release notes as pull requests are merged into master.'
+runs:
+  using: 'docker'
+  image: '../Dockerfile'
```
Jetersen: 😱 this just works out of the box, I was surprised that the workdir defaults to repo!

Awesome job @github 😍
🎉 @TimonVS

Jetersen: https://github.com/Jetersen/test-release-drafter-action-adapter/commit/a3a3381132d1c87d380dfd1ae2d46eb6fb107314

jetersen:
I honestly rather spend time code splitting the probot and github action bits.
Would enable for a future with more release drafter related actions and possibility of cli tool and typescript!

PR #384, opened 2020-01-13
Fix draft update from failing

TimonVS:
Updates to an existing draft were failing if the existing draft didn't have a tag name set. I added a check for `tag_name` on the existing draft, if it doesn't exist we don't set it instead of setting it to `null` (which is the value of `tag_name` in the previous draft if no tag name was set).


toolmantim:
Woohoo! Thanks for the fix. Was getting hit by this one the other day.

PR #397, opened 2020-01-26
Deploy to Heroku when release is published

TimonVS:


jetersen:
Added a more complete release flow. Felt so cumbersome to release: https://github.com/release-drafter/release-drafter/blob/master/.github/workflows/release.yml

Locally you release via:
```bash
yarn version --patch
```

Potentially could add a workflow to bump version 😅

PR #409, opened 2020-01-31
Use release tag commit date instead of released date

### LINKED ISSUE #222: Incorrect timestamp is used when drafting
@TimonVS seems that the GraphQL is using the date from the tag/release when it was created instead of using the date from when the tagged commit was created. That way, not all PRs will show :cry: 

https://github.com/jenkinsci/accurev-plugin

image
image

mangolas:
This change is about using commit date rather than Github releases published date for the basis of searching which commits should be used to fill next draft release changelog.

Github release published date is unchangeable and it is set to the creation time of the release. Even though tag is most often created with the release and their dates are same, this is not always the case, sometimes release tags have to be manipulated outside of release.

I couldn't figure a way to do this change in one GraphQL query, so I'm introducing a separate query here to fetch the commit date of the tag.

Fixes #222

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -54,6 +54,35 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
   }
 `
 
+const findTagCommitTimestampQuery = /* GraphQL */ `
+  query findTagCommitTimestamp($name: String!, $owner: String!, $tag: String!) {
+    repository(name: $name, owner: $owner) {
+      object(expression: $tag) {
+        ... on Commit {
+          committedDate
+        }
+      }
+    }
+  }
+`
+
+const findTagCommitTimestamp = async ({ context, tag }) => {
+  const { owner, repo } = context.repo()
+  const variables = { name: repo, owner, tag }
+  const dataPath = ['repository', 'object', 'committedDate']
+
+  let data = await context.github.graphql(
+    findTagCommitTimestampQuery,
+    variables
+  )
+  if (!_.has(data, dataPath)) {
+    throw new Error(
+      `Data doesn't contain 'committedDate' field when querying tag '${tag}'`
+    )
+  }
+  return _.get(data, dataPath)
+}
+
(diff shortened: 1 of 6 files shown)


casz:
Perhaps we could raise a request for release to also have tag creation date?

mangolas:
Sorry, could you elaborate a bit, I don't quite follow your question.

casz:
We could ask GitHub to update the GraphQL api to include the tag with commit creation date on the release object.

We asked for feature like this earlier https://github.community/t5/GitHub-API-Development-and/Get-pull-request-associated-with-a-commit/m-p/20305/highlight/true#M1100

mangolas:
Aa, I see, yeah, why not, it would be handy to have it directly on the release.

casz:
Hey @mangolas 
I think you can actually manage to do it in one go after reading: https://github.community/t5/GitHub-API-Development-and/GraphQL-How-to-get-a-list-of-commits-between-tags-or-releases/m-p/24982#M1895

```
{
  repository(owner: "release-drafter", name: "release-drafter") {
    releases(last: 2) {
      nodes {
        tagName
        createdAt
      }
    }
  }
}
```

casz:
That actually requires a release exist with tags 🤔

casz:
Hmm should this not be updated to simplify what you want?

https://github.com/release-drafter/release-drafter/blob/36d7384e147b7cc02b55be4cc799f30e96c5e1fd/lib/releases.js#L23-L29

mangolas:
Yeah, unfortunately the repository search is not very helpful for the draft release case.

Could you elaborate on your second question, what simplification you have in mind?

casz:
The `lastRelease` comes from the lines I linked above and since you already are using the `lastRelease.tag_name` to find the creation date of the `tag`

I believe you could use the above GraphQL to see if you can find the releases and their tag creation date with just the GraphQL

mangolas:
Ok, now I get what you mean. Mm, maybe it would indeed be possible to replace `listReleases` call with GraphQL. But I'm not sure doing it and refactoring the tests would be worth the effort.

PR #412, opened 2020-02-02
Fix overlapping label categorization

### LINKED ISSUE #293: Category with multiple labels produces a header per label
Using the latest (`v5.2.0`) as a GitHub Action in my repo's workflow, I am seeing adverse behavior from #219 

My `release-drafter.yml` configuration file:

```yaml
name-template: '$NEXT_PATCH_VERSION'
tag-template: '$NEXT_PATCH_VERSION'
categories:
  - title: '🚀 Features'
    labels:
      - 'enhancement'
  - title: '🐛 Bug Fixes'
    labels:
      - 'bug'
  - title: '🛠️ Maintenance'
    labels:
      - 'maintenance'
      - 'dependencies'
template: |
  ## What's Changed

  $CHANGES
``` 

Notice that the category with the title "Maintenance" has 2 labels under it: "maintenance" and "dependencies".

When I have a Draft Release open that features merged PRs from _each_ of those labels, the Draft Release contents end up with a duplicate header:

image

### LINKED ISSUE #403: Multiple labels not grouping under one heading
Expected behavior: If there are multiple labels under one grouping, they are expected to be grouped under that heading.
Actual behavior: The heading is being repeated for each label under that title/heading.

Screenshots:
Config file:
Screen Shot 2020-01-29 at 1 40 53 PM

Result:
Screen Shot 2020-01-29 at 1 31 40 PM

The above have labels bug, bugfix, and sentry respectively.

casz:
resolves #293
resolves #403

This should fix categorization overlapping. I'd like to have someone review the code around the looping. Wonder if we could optimize the looping.

I'd like to avoid using `reduce` as I feel it makes this area of the code particular hard to read 😅

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -86,25 +86,32 @@ const categorizePullRequests = (pullRequests, config) => {
     return { ...category, pullRequests: [] }
   })
 
-  // exclude pull requests with exclude labels first than push any uncategorized pull requests to UNCATEGORIZED category
-  const filterPullRequests = pullRequest => {
+  const filterExcludedPullRequests = pullRequest => {
     const labels = pullRequest.labels.nodes
 
     if (labels.some(label => excludeLabels.includes(label.name))) {
       return false
-    } else if (
+    }
+    return true
+  }
+
+  const filterUncategorizedPullRequests = pullRequest => {
+    const labels = pullRequest.labels.nodes
+
+    if (
       labels.length === 0 ||
       !labels.some(label => allCategoryLabels.includes(label.name))
     ) {
       uncategorizedPullRequests.push(pullRequest)
       return false
     }
-
     return true
   }
 
   // we only want pull requests that have yet to be categorized
-  const filteredPullRequests = pullRequests.filter(filterPullRequests)
+  const filteredPullRequests = pullRequests
+    .filter(filterExcludedPullRequests)
+    .filter(filterUncategorizedPullRequests)
 
   categorizedPullRequests.map(category => {
     filteredPullRequests.map(pullRequest => {
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -79,25 +79,27 @@ const contributorsSentence = ({ commits, pullRequests }) => {
 }
 
 const getCategoriesConfig = ({ config }) => {
-  const categoriesConfig = [{ title: UNCATEGORIZED }, ...config.categories]
-
-  return categoriesConfig.map(category => {
+  const categories = []
+  const orderedCategories = [
+    { title: UNCATEGORIZED },
+    ...config.categories
+  ].map(category => {
     if (!category.labels) category.labels = []
     if (category.label) {
       // treat label as labels
       category.labels.push(category.label)
       delete category.label
     }
+    categories.push(...category.labels)
     return { ...category, pullRequests: [] }
   })
+
+  return [categories, orderedCategories]
 }
 
-const categorizePullRequests = (
-  pullRequests,
-  orderedCategories,
-  excludeLabels
-) => {
-  const categories = orderedCategories.flatMap(category => category.labels)
+const categorizePullRequests = (pullRequests, config) => {
+  const excludeLabels = config['exclude-labels']
+  const [categories, orderedCategories] = getCategoriesConfig({ config })
 
   // exclude pull requests with exclude labels first than push any uncategorized pull requests to UNCATEGORIZED category
   const filterPullRequests = pullRequest => {
@@ -112,6 +114,7 @@ const categorizePullRequests = (
       orderedCategories[0].pullRequests.push(pullRequest)
       return false
     }
+
     return true
   }
 
@@ -140,11 +143,9 @@ const generateChangeLog = (mergedPullRequests, config) => {
     return config['no-changes-template']
   }
 
-  const categoriesConfig = getCategoriesConfig({ config })
   const categorizedPullRequests = categorizePullRequests(
     mergedPullRequests,
-    categoriesConfig,
-    config['exclude-labels']
+    config
   )
   return categorizedPullRequests
     .reduce((acc, category, index) => {
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -104,23 +104,21 @@ const categorizePullRequests = (
     const labels = pullRequest.labels.nodes
 
     if (labels.some(label => excludeLabels.includes(label.name))) {
-      return null
+      return false
     } else if (
       labels.length === 0 ||
       !labels.some(label => categories.includes(label.name))
     ) {
       orderedCategories[0].pullRequests.push(pullRequest)
-      return null
+      return false
     }
-    return pullRequest
+    return true
   }
 
   const filterUncategorized = category => category.title !== UNCATEGORIZED
 
   // we only want pull requests that have yet to be categorized
-  const uncategorizedPullRequests = pullRequests
-    .map(filterPullRequests)
-    .filter(Boolean)
+  const uncategorizedPullRequests = pullRequests.filter(filterPullRequests)
 
   orderedCategories.filter(filterUncategorized).map(category => {
     uncategorizedPullRequests.map(pullRequest => {
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -78,28 +78,13 @@ const contributorsSentence = ({ commits, pullRequests }) => {
   }
 }
 
-const getCategoriesConfig = ({ config }) => {
-  const categories = []
-  const orderedCategories = [
-    { title: UNCATEGORIZED },
-    ...config.categories
-  ].map(category => {
-    if (!category.labels) category.labels = []
-    if (category.label) {
-      // treat label as labels
-      category.labels.push(category.label)
-      delete category.label
-    }
-    categories.push(...category.labels)
+const categorizePullRequests = (pullRequests, config) => {
+  const { 'exclude-labels': excludeLabels, categories } = config
+  const allCategoryLabels = categories.flatMap(category => category.labels)
+  const uncategorizedPullRequests = []
+  const categorizedPullRequests = [...categories].map(category => {
     return { ...category, pullRequests: [] }
   })
-
-  return [categories, orderedCategories]
-}
-
-const categorizePullRequests = (pullRequests, config) => {
-  const excludeLabels = config['exclude-labels']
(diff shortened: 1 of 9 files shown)


lib/releases.js
```diff
@@ -80,74 +80,86 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+
+  const filterUncategorized = category => category.title !== UNCATEGORIZED
+
+  // we only want pull requests that have yet to be categorized
+  const uncategorizedPullRequests = pullRequests
+    .map(filterPullRequests)
+    .filter(Boolean)
+
+  orderedCategories.filter(filterUncategorized).map(category => {
+    uncategorizedPullRequests.map(pullRequest => {
```
casz: Perhaps there is a better way to achieve this without looping over each pull request for each category.

lib/releases.js
```diff
@@ -80,74 +80,86 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
 
-const categorizePullRequests = (pullRequests, categories, excludeLabels) => {
-  return pullRequests.reduce(
-    (acc, pullRequest) => {
+const categorizePullRequests = (
+  pullRequests,
+  orderedCategories,
+  excludeLabels
+) => {
+  const categories = [...orderedCategories.flatMap(category => category.labels)]
```
casz:
```suggestion
  const categoryList = orderedCategories.flatMap(category => category.labels)
```

lib/releases.js
```diff
@@ -80,74 +80,86 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+      return null
+    }
+    return pullRequest
+  }
+
+  const filterUncategorized = category => category.title !== UNCATEGORIZED
+
+  // we only want pull requests that have yet to be categorized
+  const uncategorizedPullRequests = pullRequests
+    .map(filterPullRequests)
```
casz: for some reason I cannot pass this to filter and inside the `filterPullRequests` pass true or false back.

test/index.test.js
```diff
@@ -669,6 +669,122 @@ Previous tag: ''
...
+          .reply(
+            200,
+            require('./fixtures/__generated__/graphql-commits-overlapping-label.json')
+          )
+
+        nock('https://api.github.com')
+          .post(
+            '/repos/toolmantim/release-drafter-test-project/releases',
+            body => {
+              expect(body).toMatchInlineSnapshot(`
```
casz: We should use `toMatchInlineSnapshot` for all our other inline matching 😅

casz:
Would appreciate a reviews from @JamesMGreene @TimonVS 🙇

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+
+  return categoriesConfig.map(category => {
+    if (!category.labels) category.labels = []
+    if (category.label) {
+      // treat label as labels
+      category.labels.push(category.label)
+      delete category.label
+    }
+    return { ...category, pullRequests: [] }
+  })
```
TimonVS: This should probably be part of the config validation. Can you move this to `config.js` or `schema.js` as `validateCategoriesConfig` or `normalizeCategoriesConfig`?

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
 }
 
-const categorizePullRequests = (pullRequests, categories, excludeLabels) => {
-  return pullRequests.reduce(
-    (acc, pullRequest) => {
+const categorizePullRequests = (
+  pullRequests,
+  orderedCategories,
```
TimonVS:
```suggestion
  categoriesConfig,
```

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+  excludeLabels
+) => {
+  const categories = orderedCategories.flatMap(category => category.labels)
```
TimonVS: Categories and labels are not the same thing. I'm not sure if my suggestion is the best name, but something along these lines would be an improvement.

```suggestion
  const allCategoryLabels = orderedCategories.flatMap(category => category.labels)
```

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+  const filterPullRequests = pullRequest => {
+    const labels = pullRequest.labels.nodes
+
+    if (labels.some(label => excludeLabels.includes(label.name))) {
+      return false
+    } else if (
+      labels.length === 0 ||
+      !labels.some(label => categories.includes(label.name))
+    ) {
+      orderedCategories[0].pullRequests.push(pullRequest)
```
TimonVS: 1. We shouldn't mutate inputs as that can cause subtle bugs.
2. This can cause bugs when we decide to change the order of uncategorized PRs. Can you push these to a separate array?

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
-    (acc, pullRequest) => {
+const categorizePullRequests = (
+  pullRequests,
+  orderedCategories,
+  excludeLabels
+) => {
+  const categories = orderedCategories.flatMap(category => category.labels)
+
+  // exclude pull requests with exclude labels first than push any uncategorized pull requests to UNCATEGORIZED category
+  const filterPullRequests = pullRequest => {
```
TimonVS: The comment describes what the function is doing. I prefer to only have comments to describe the "why". If we encode this in the name of the function, the comment becomes unnecessary. In order to do this we must first make this function responsible for one thing: `filterExcludedPullRequests`. We can create a second function which is responsible for putting uncategorized PRs in a separate array.

lib/releases.js
```diff
@@ -80,74 +79,84 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
-      ) {
-        acc[UNCATEGORIZED].push(pullRequest)
-      } else {
-        labels.forEach(label => {
-          if (!acc[label.name]) acc[label.name] = []
-
-          acc[label.name].push(pullRequest)
-        })
+      if (labels.some(label => category.labels.includes(label.name))) {
+        category.pullRequests.push(pullRequest)
```
TimonVS: Also here we shouldn't mutate the input :)
We can create a new array which contains the results.

lib/releases.js
casz: Yes schema can totally take care of most of it.

Now the function only handles unwrapping categories into allCategoryLabels, orderedCategories

casz: The label to labels and making sure it is an array has moved to schema.json

lib/releases.js
casz: I'd like to avoid looping over the same thing too many times 😓 But that's just me.

lib/releases.js
casz: Created a new array.

lib/releases.js
casz: created a separate array

casz:
Any chance we can get this merged? We have plenty of open pull requests. 😅

test/index.test.js
TimonVS: That's a great idea!

lib/releases.js
```diff
@@ -79,91 +78,95 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+
+  // we only want pull requests that have yet to be categorized
+  const filteredPullRequests = pullRequests
+    .filter(filterExcludedPullRequests)
+    .filter(filterUncategorizedPullRequests)
+
+  categorizedPullRequests.map(category => {
+    filteredPullRequests.map(pullRequest => {
+      // lets categorize some pull request based on labels
+      // due note that having the same label in multiple categories
```
TimonVS:
```suggestion
      // note that having the same label in multiple categories
```

casz:
👏 Thanks @TimonVS

oleg-nenashev:
Uh, I was actually using the label duplication in order to simplify splitting multi-entry changelogs. E.g. see https://github.com/jenkinsci/plugin-pom/pull/271 where I inject two entries into PR title and put two labels categorizing each of them. Then I just copy-edit the release draft 

Mandatory XKCD comics for such comment:

image

https://xkcd.com/1172/

casz:
Should be fairly simple to add a condition enables it again.

Jetersen:
@oleg-nenashev btw that case is still supported. We loop over each category so if a PR has multiple labels that land into multiple categories you still get the PR duplicated.

Also clearly noted in my code comment:
https://github.com/release-drafter/release-drafter/blob/ec0f2c5e261a34c78e21e978fedac2810daf2165/lib/releases.js#L116-L126

Jetersen:
@oleg-nenashev btw that case is still supported. We loop over each category so if a PR has multiple labels that land into multiple categories you still get the PR duplicated.

Also clearly noted in my code comment:
https://github.com/release-drafter/release-drafter/blob/ec0f2c5e261a34c78e21e978fedac2810daf2165/lib/releases.js#L116-L126

The duplication we removed was when multiple labels was added to multiple categories.

So for instance with the following config:
```yaml
categories:
  - title: 🐛 Bug Fixes
    labels:
      - bug
      - fix
      - bugfix
      - regression
```
a PR with the labels `bug` and `fix` would previously \
be duplicated twice inside the same `🐛 Bug Fixes` category

Now with this PR it will only be included once

Jetersen:
perhaps the comment should be updated with or duplicated if a pr has multiple labels in multiple categories.

jetersen:
@timja this PR should fix the bug you experienced in slack plugin:
image
https://github.com/jenkinsci/slack-plugin/pull/675
https://github.com/jenkinsci/slack-plugin/pull/682


as `bug` and `fix` this made it "duplicated" in the same category which this PR fixes.

Hopefully that makes the case clearer @oleg-nenashev for when the bug will occur.

if a PR has multiple labels that fit into multiple categories, that is not considered as duplication.

@TimonVS @toolmantim would appreciate a release 😅

ssbarnea:
As this was finally fixed, how long should we wait till this will reach the app at https://github.com/apps/release-drafter ? That bug is making the result really ugly https://github.com/ansible-community/molecule/releases

jetersen:
@ssbarnea finally got access to heroku and npm :)
So I managed to deploy a new version of the probot app.

PR #304, opened 2019-09-11
Autorelease functionality

JimNero009:
Allow the release drafter to auto-publish the releases it drafts, if a user so wants.
Works by inspecting labels on the PRs that have gone into making the draft, then publishing the draft with the version that best fits.

PR #424, opened 2020-02-13
Allow creating published releases

dfreeman:
This PR introduces an additional (optional) action input: `publish`. When set to `true`, it will cause the generated release notes not to be created in a draft state.

I'd been trying to work out a way of automating the process of publishing the draft produced by release-drafter when a new tag was created, and it finally occurred to me that the simplest approach would be to have release-drafter accept a flag for doing that itself.

I realize `draft` is in the name of this project, though, so I'll of course understand if the response to this is "nope, this is out of scope" 😄

I also couldn't find any documentation of the existing action inputs (other than `config-name` being mentioned in the sample config), so I wasn't sure where to document this if it _is_ something you'd be interested in accepting.

Regardless, thanks for maintaining this great project!


dfreeman:
I should also mention that I spent a bit of time looking at #304, but found that I didn't need release-drafter to do any intelligent decision making on its own. 

Rather, based on the context of the action (e.g. running in response to a tag or using outputs from previous workflow steps), I could just set the inputs to the release-drafter step accordingly.

index.js
```diff
@@ -53,6 +53,8 @@ module.exports = app => {
       name: core.getInput('name') || undefined
     })
 
+    const shouldPublish = core.getInput('publish') === 'true'
```
casz: what if I spelled it `True` or `TRUE`

index.js
```diff
@@ -61,7 +63,7 @@ module.exports = app => {
           name: releaseInfo.name,
           tag_name: releaseInfo.tag,
           body: releaseInfo.body,
-          draft: true,
+          draft: !shouldPublish,
```
casz: Not a understandable name in context of draft especially because you negate the condition.
`shouldDraft` might make more sense and then you would not have to negate here?

index.js
dfreeman: Good call. It looks like going case-insensitive is [consistent with Github’s own actions](https://github.com/actions/checkout/blob/f858c22e963bc60bc9d01c3d105c52a45a7deb6e/src/input-helper.ts#L85)

index.js
dfreeman: Sure, ok. I’m happy to make these changes, but I’d also love to get a sense of whether the overall approach here is something likely to be accepted or if that would be wasted effort 🙂

index.js
```diff
@@ -53,6 +53,8 @@ module.exports = app => {
       name: core.getInput('name') || undefined
     })
 
+    const shouldDraft = core.getInput('publish').toLowerCase() !== 'true'
+
     let createOrUpdateReleaseResponse
     if (!draftRelease) {
       log({ app, context, message: 'Creating new draft release' })
```
bartaz: Just noticed this when testing this branch in the workflow.
Even when release is published (`shouldDraft` is false), the logs say `Creating new draft release` or `Updating existing draft release` which not true when release is published, not drafted.

dfreeman: Good catch. @toolmantim/@TimonVS if you're interested in eventually merging this PR, do you have a preference for how to update the language here?

casz:
Thanks for solving the conflict :)

index.js
dfreeman: (When resolving the conflict with latest `master`, I just removed the world "draft" from the message, but I'm happy to make other updates if there's a preference to get more detail in there)

TimonVS:
Thanks for the PR! Would you mind detailing how you'll use the `publish` input? What gets to decide to put `publish` to `true` in your case?

dfreeman:
In the couple of projects at work where we've been trialing release-drafter, it's been paired with a few other workflow steps including [`detect-and-tag-new-version`](https://github.com/marketplace/actions/detect-and-tag-new-version). We use the output from that step (whether it created a new tag based on a change in e.g. `package.json`) to determine whether to set `publish: true`.

I could also imagine a flow where the maintainer manually creates git tags when the time comes, and has a workflow with release-drafter set up specifically to trigger on new tags, with `publish: true` simply hard-coded in its config.

TimonVS:
Cool! Would you mind sharing the workflow for that?
In any case, I'll be merging this because it sounds like it could be useful for some advanced use cases.

dfreeman:
Sure! I'm hoping to write up a quick blog post or something sometime soon if I get the chance, but here's one I set up at work this morning:

```yml
name: Release

on:
  push:
    branches:
      - master

jobs:
  release:
    name: Check and Release New Version
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Code
        uses: actions/checkout@v2
        with:
          fetch-depth: 2

      - name: Check for Version Change
        id: check-version
        uses: salsify/action-detect-and-tag-new-version@v1

      - name: Publish New Package Version
        if: steps.check-version.outputs.tag
        run: |
          # Publishes the package to our internal registry

      - name: Publish Release Notes
        if: steps.check-version.outputs.tag
        # We can switch to a tagged version once 5.8 is published
        uses: release-drafter/release-drafter@f677696
        with:
          publish: true
          tag: ${{ steps.check-version.outputs.tag }}
        env:
```

With this setup, running `release-drafter` at all is conditional on whether a tag was created, so `publish` is hard set to `true`. You could also write something like this, though, to keep draft notes around beforehand and then just publish them when the time comes:

```yml
      - name: Publish Release Notes
        uses: release-drafter/release-drafter@f677696
        with:
          publish: ${{ steps.check-version.outputs.tag != '' }}
          tag: ${{ steps.check-version.outputs.tag }}
        env:
```

Either way, this makes releasing a new package version just a matter of merging a PR that bumps the version specifier in source.

pshelby:
@dfreeman Thanks for this feature, I'm really enjoying the simplicity and functionality of release-drafter!  I've noticed, however, that the `publish` functionality was checked into master but isn't yet available in a `v*` release.  Is it by design to not have this logic in `v5` yet?

[Commits on master but not yet in v5.7.0](https://github.com/release-drafter/release-drafter/compare/v5.7.0...master)

Jetersen:
at the moment there is actually a bug in v5.7.0 see #445 
and see https://github.com/release-drafter/release-drafter/pull/376#discussion_r392698467

pshelby:
> at the moment there is actually a bug in v5.7.0 see #445
> and see [#376 (comment)](https://github.com/release-drafter/release-drafter/pull/376#discussion_r392698467)

Cool, thanks for the quick info!

jetersen:
Release inc :)

jetersen:
This is actually missing docs 😓 @dfreeman would you be the man for adding some?

dfreeman:
@jetersen like I mentioned in the original description, none of the action inputs seem to be documented anywhere, so I'm not sure where these docs should even go—it seems we're missing a full section somewhere 🙂

jetersen:
There is a [action outputs](https://github.com/release-drafter/release-drafter#action-outputs) section, you could start an action input section? :)

dfreeman:
I likely won't have time to get to that in the next few days, but if no one beats me to it before next week I'm happy to take a stab

jetersen:
FYI

This can be even prettier with #546 
See this example
```yaml
name: Release Drafter

on:
  push:
    branches:
      - master
    tags:
      - v*
    release:
      types: [deleted]

jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
      - name: Draft Release
        with:
          publish: ${{ contains(github.ref, "ref/tags") }}
        env:
```

timja:
Your example didn't work for me @jetersen,
I ended up with:

```
publish: startsWith(github.ref, "refs/tags")
```

https://github.com/jenkinsci/dark-theme-plugin/blob/master/.github/workflows/release-drafter.yml#L16

jetersen:
Neat, this was just a brain dump from me 🙈

jetersen:
You properly need to update the on conditions:

https://github.com/jenkinsci/dark-theme-plugin/blob/fb51bcd2ccb721e32e0abe835bfe9f0cc1376382/.github/workflows/release-drafter.yml#L3-L7

muuki88:
Thanks for the helpful comments.

Would you merge a PR that adds an example to the `README.md` ?

I'm currently using this

```yml
name: Release

on:
  push:
    tags:
      - 'v*'

jobs:
  release:
    name: Check and Release New Version
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Code
        uses: actions/checkout@v2
        with:
          fetch-depth: 2

      # Parse the release version from git tag
      - name: Get release version from tag
        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV

      # Validate and publish
      - name: Publish / Deploy
        run: echo 'publish your lib or deploy your app here'

      - name: Publish Release Notes
        with:
          publish: true
          tag: ${{ env.RELEASE_VERSION }}
        env:
``

PR #445, opened 2020-03-15
fix setOutput

Jetersen:
This adds body reply so when `setOutput` is hit they actually have something to output


Jetersen:
As the master branch is currently broken for those using master branch, I am going to merge this.

PR #500, opened 2020-04-22
Add include-labels option

### LINKED ISSUE #499: Add config option for `include-labels`
There is currently a configuration option to specify labels to exclude.  However for certain repositories (mono-repos come to mind) it would be helpful to have the opposite - to have a configuration option so that the release ONLY includes PRs with the specified labels.

This option would be added to the `release-drafter.yml` and would look something like the following:

```
include-labels:
  - 'application_foo'
```

In this particular example only PR's that have the `application_foo` label would be included in the drafted release.

One point of conversation would be to think about what would happen if the user had both `include-labels` and `exclude-labels` in the configuration.  I believe that `exclude-labels` should take precedence.  That is - if a PR has a label that is in the `include-labels` list as well as a label that is in the `exclude-labels` list the PR should be excluded.  If the user has the same exact label in both lists the configuration should be invalidated.

kreyusb:
Adding feature discussed in: https://github.com/release-drafter/release-drafter/issues/499

Note, feature has not been agreed upon yet. Consider this WIP until feature has consensus.

fixes #499

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -79,7 +79,7 @@ const contributorsSentence = ({ commits, pullRequests }) => {
 }
 
 const categorizePullRequests = (pullRequests, config) => {
-  const { 'exclude-labels': excludeLabels, categories } = config
+  const { 'exclude-labels': excludeLabels, 'include-labels': includeLabels, categories } = config
   const allCategoryLabels = categories.flatMap(category => category.labels)
   const uncategorizedPullRequests = []
   const categorizedPullRequests = [...categories].map(category => {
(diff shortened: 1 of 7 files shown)


lib/releases.js
```diff
@@ -80,6 +80,7 @@ const contributorsSentence = ({ commits, pullRequests }) => {
 
 const categorizePullRequests = (pullRequests, config) => {
   const { 'exclude-labels': excludeLabels, categories } = config
+  const { 'include-labels': includeLabels } = config
```
jetersen, line 83:
```suggestion
  const { 
    'exclude-labels': excludeLabels,
    'include-labels': includeLabels,
    categories 
  } = config
```

jetersen:
Is this still to be considered WIP @kreyusb ?

kreyusb:
@jetersen We have been using it like this on our Company configuration for the last couple of weeks with no changes.  I would say it is good to go.

darkdreamingdan:
Can we go ahead and merge this?  Seems like everyone is happy.   I see nothing blocking and concerned good code is just getting stuck and we're getting merge conflicts.

jetersen:
There is currently a conflict, do you mind solving in kreyusb?
Than we are ready to merge.

jetersen:
Many thanks!

PR #513, opened 2020-05-06
Add support for automatic version resolution

darkdreamingdan:
# Motivation
In our team, the person initiating the release often doesn't have context on whether the release should be patch, minor or major. People working on invididual PRs have better judgement on whether their change is substantial enough to be major/minor/patch.  Therefore, they should apply a github label to decide the nature of the change.  Currently, release-drafter provides no way to automatically determine the version number.

This change adds the ability to determine what the next version number should be based upon currentVersion + maxIncrement(major, minor, patch).

Please see updates to the `README.md` for details.

# Feedback required
- I'm not experienced in JS, so would appreciate advice and tips where necessary about code style and design.  I've tried to do this in a way that makes sense from my understanding of the code.
- Please give an early indication of the direction of travel of this PR.  I'm not sure if i've gone about it according to the best way possible.
- ~Please give advice on testing.  Again, because i am a JS dufus I didn't quite understand the testing structure, and the `resolveVersionKeyInc` method remains untested.  This would certainly need to be addressed before merging.~

Thanks, and look forward to contributing!!


jetersen:
a similar PR tried it but got stuck #304

darkdreamingdan:
> a similar PR tried it but got stuck #304

Hmm i see.  It looks like a much larger scoped PR.  I'm not a fan of his implementation of `getWinningLabelFrom` but it looks like it works nevertheless.  How do you suggest I proceed? I am happy to combine efforts to get the other PR merged, or to keep this a tightly scoped PR over here.

I think my implementation is a bit more elegant, but would look to your direction.  Keen to get something in for my team :) .

jetersen:
I do agree your changes seems more elegant.

Could you run `yarn run generate-schema` and commit the `./schema.json` changes?

I will look into generating test fixtures for your changes and add the tests to your PR, does that sound good?

jetersen:
Remember with GitHub actions you can already tests your changes by modifying the `uses:` in the GitHub action workflow

darkdreamingdan:
I've regenerated the `schema.json` now.  

I was also able to understand the testing structure.  I haven't made new generated fixtures because of the requirement to have access to the test repo, and haven't made new custom fixtures because of how bloated they are.  I instead decided to monkeypatch the existing generated fixture and i think this way the test is more readable.  I don't think further work on testing is required from yourself - thankyou for offering to help!

This PR is ready for final review/merge in my view.  Thanks! 😄

pdcmoreira:
@darkdreamingdan I've just created a PR very similar to yours (long story short, I had this in mind for a while but should've checked again if this was already PR'd).

The major difference is that, in my case, you just configure the labels for major and minor increments, all else will simply fallback to 'PATCH'.
Implementation-wise, it's all very similar, exceptions are:
- You allow to configure the default incrementation
- You have a priority map
- I take into consideration the excluded PRs/labels (re-using the existing logic)
- I made a small performance-improvement by not calculating the version if there is an `inputVersion`

A couple comments:
I don't know if allowing to configure the default incrementation is useful or not, maybe in some very specific use-cases.
Also, I don't think it's necessary to have a priority map, it is always major -> minor -> patch.

Anyway, if this gets merged, I'll just probably close my PR.

darkdreamingdan:
> @darkdreamingdan I've just created a PR very similar to yours (long story short, I had this in mind for a while but should've checked again if this was already PR'd).
> 
> The major difference is that, in my case, you just configure the labels for major and minor increments, all else will simply fallback to 'PATCH'.
> Implementation-wise, it's all very similar, exceptions are:
> 
>     * You allow to configure the default incrementation
> 
>     * You have a priority map
> 
>     * I take into consideration the excluded PRs/labels (re-using the existing logic)
> 
>     * I made a small performance-improvement by not calculating the version if there is an `inputVersion`
> 
> 
> A couple comments:
> I don't know if allowing to configure the default incrementation is useful or not, maybe in some very specific use-cases.
> Also, I don't think it's necessary to have a priority map, it is always major -> minor -> patch.
> 
> Anyway, if this gets merged, I'll just probably close my PR.

Hey! Ah snap, it happens :) .

Some thoughts:
- Regarding default incremementation, actually in our project we want to use 'minor' as the default because patch is only for bug fixes to an existing release (This is according to the SEMVER spec).  For us, generally bugfixes happen more rarely than general additions.  So i think there's a good case for keeping it
- The priority map isn't really significant per se.  It was just a functional implementation to declare the priority of major -> minor -> patch because i can leverage `Math.max` to do heavy lifting instead of `if` statements.
- I agree with the feature of exclude-labels.  If it's okay with you, i will cherry pick this code into this PR and we can get it merged?

Cheers

darkdreamingdan:
I've added support for exclude-labels now.  I've simplified your implementation a bit by using a closure.

@jetersen , are we good to approve this? :)

pdcmoreira:
@darkdreamingdan All right, all good! At least my PR was useful for something... 😅 
The only thing I don't like on your implementation is the abbreviation in the name of the method `resolveVersionKeyInc`, could it not be something like `resolveIncrementation`? Or even `resolveVersionIncrementation`?
Sorry for the nitpick!

darkdreamingdan:
> @darkdreamingdan All right, all good! At least my PR was useful for something... 😅
> The only thing I don't like on your implementation is the abbreviation in the name of the method `resolveVersionKeyInc`, could it not be something like `resolveIncrementation`? Or even `resolveVersionIncrementation`?
> Sorry for the nitpick!

I guess i had picked `resolveVersionKeyInc` because this harmonizes with the nomenclature of `versionKey` and `input.inc()` in versions.js.

```
const splitSemVer = (input, versionKey = 'version') => {
  if (!input[versionKey]) {
    return null
  }

  const version = input.inc
    ? semver.inc(input[versionKey], input.inc, true)
...
```

I don't feel that strongly about it, and agree that `Inc` is ambigious.  I've changed this to versionKeyIncrement now, but think we should have `Key` inside because that's how we refer to it in other places.

See what you think.

pdcmoreira:
@darkdreamingdan Seems good! Make a review request to @jetersen and @toolmantim

darkdreamingdan:
I don't seem to have access to add reviewers.  @jetersen / @toolmantim - got a spare brain handy for review? :)

README.md
```diff
@@ -96,6 +107,7 @@ You can configure Release Drafter using the following key in your `.github/relea
 | `sort-by`             | Optional | Sort changelog by merged_at or title. Can be one of: `merged_at`, `title`. Default: `merged_at`.                                                                                                                  |
 | `sort-direction`      | Optional | Sort changelog in ascending or descending order. Can be one of: `ascending`, `descending`. Default: `descending`.                                                                                                 |
 | `prerelease`          | Optional | Mark the draft release as pre-release. Default `false`.                                                                                                                                                           |
+| `version-resolver`    | Optional | Adjust the \$RESOLVED_VERSION variable using labels. Refer to [Version Resolver](#version-resolver) to learn more about this                                                                                      |
```
jetersen, line 110:
```suggestion
| `version-resolver`    | Optional | Adjust the `$RESOLVED_VERSION` variable using labels. Refer to [Version Resolver](#version-resolver) to learn more about this                                                                                      |
```

README.md
```diff
@@ -118,6 +130,7 @@ You can use any of the following variables in your `template`, `name-template` a
 | `$NEXT_PATCH_VERSION` | The next patch version number. For example, if the last tag or release was `v1.2.3`, the value would be `v1.2.4`. This is the most commonly used value. |
 | `$NEXT_MINOR_VERSION` | The next minor version number. For example, if the last tag or release was `v1.2.3`, the value would be `v1.3.0`.                                       |
 | `$NEXT_MAJOR_VERSION` | The next major version number. For example, if the last tag or release was `v1.2.3`, the value would be `v2.0.0`.                                       |
+| `$RESOLVED_VERSION`   | The next resolved version number, based on GitHub labels. Refer to [Version Resolution](#version-resolution) to learn more about this.                  |
```
jetersen, line 133: Have you tested how this works with `$NEXT_MAJOR_VERSION` and `$NEXT_MINOR_VERSION`?
I guess it has no effect.

See usage here: https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/.github/release-drafter.yml

lib/releases.js
```diff
@@ -169,6 +174,31 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
+  }
+  const labelToKeyMap = Object.fromEntries(
+    Object.keys(priorityMap)
+      .flatMap(key => [
+        config['version-resolver'][key].labels.map(label => [label, key])
+      ])
+      .flat()
+  )
+  const keys = mergedPullRequests
+    .filter(getFilterExcludedPullRequests(config['exclude-labels']))
```
jetersen, line 191: I am more inclined to merge #500 first, so you may need to rewrite this

test/index.test.js
```diff
@@ -1752,5 +1752,231 @@ Previous tag: ''
...
+        })
+
+        expect.assertions(1)
+      })
+      it('minor beats patch', async () => {
+        getConfigMock('config-with-custom-version-resolver.yml')
+
+        const response = require('./fixtures/__generated__/graphql-commits-merge-commit.json')
+        response.data.repository.ref.target.history.nodes[0].associatedPullRequests.nodes[0].labels.nodes = [
+          { name: 'PAT' }
```
jetersen, line 1843: I would prefer actual generated test fixtures rather than modifying a fixture.
I'll look into generating the needed test fixtures.

lib/releases.js
darkdreamingdan: Flicking through that PR, think it'd be very easy to merge so happy for that approach (hopefully we don't get stuck though)

README.md
darkdreamingdan: I don't modify those variables so there won't be any effect. The existing unit tests already have testing for those variables, so don't see why it would be a problem.

jetersen:
😓 conflicts after style and #500 Hopefully your up for resolving them?

darkdreamingdan:
I've merged with master now :) .

bin/generate-fixtures.js
```diff
@@ -37,13 +37,13 @@ branches.forEach((branch) => {
         name: REPO_NAME,
         branch,
         withPullRequestBody: true
-      },
-    }),
+      }
```
jetersen, line 40: Seems your fighting prettier new config 😓

jetersen:
Tomorrow I will add some labels to those PRs in our test repo.
Been busy with other open source work 😓

bin/generate-fixtures.js
darkdreamingdan: Sorry i didn't notice all these changes! i'll try and fix this diff tomorrow

jetersen: most likely due to a old yarn install of prettier 🗡️

darkdreamingdan:
I've remerged now.  Hopefully the diff is readable this time!

jetersen:
@darkdreamingdan I have added some more labels to the PR in our test repo. Could you run the `yarn generate-fixtures` or `npm run generate-fixtures` and commit those.

Perhaps there should be an option to disable the automatic version resolution (I think my changes might break a few tests) 😓

jetersen:
or perhaps I should change the labels slightly https://github.com/TimonVS/release-drafter-test-repo/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed

darkdreamingdan:
> @darkdreamingdan I have added some more labels to the PR in our test repo. Could you run the `yarn generate-fixtures` or `npm run generate-fixtures` and commit those.
> 
> Perhaps there should be an option to disable the automatic version resolution (I think my changes might break a few tests) 😓

It is already optional - the version will only increment if you use `$RESOLVED_VERSION` (most of the tests use `$NEXT_PATCH_VERSION`). 

Thanks for updating the testing repo.  I have now updated the tests to use the generated fixtures. Give it a look :) .

lib/default-config.js
```diff
@@ -7,9 +7,9 @@ const DEFAULT_CONFIG = Object.freeze({
   'no-changes-template': `* No changes`,
   'version-template': `$MAJOR.$MINOR.$PATCH`,
   'version-resolver': {
-    major: { labels: ['major'] },
-    minor: { labels: ['minor'] },
-    patch: { labels: ['patch'] },
+    major: { labels: [] },
+    minor: { labels: [] },
+    patch: { labels: [] },
```
jetersen, line 12: Hmm already optional? Seems the latest commit has now made it optional, unless I am mistaken @darkdreamingdan

lib/releases.js
```diff
@@ -76,38 +76,50 @@ const contributorsSentence = ({ commits, pullRequests }) => {
...
+    return { ...category, pullRequests: [] }
+  })
+
+  const filterExcludedPullRequests = getFilterExcludedPullRequests(
+    excludeLabels
+  )
+
+  const filterIncludedPullRequests = getFilterIncludedPullRequests(
+    includeLabels
+  )
```
jetersen, line 120: This reads funny...

Could we not call getFilterIncludedPullRequests() directly in the filtering below on line [137-138](https://github.com/darkdreamingdan/release-drafter/blob/781e89dd260b0fdd18bfddb540005c710ccab74f/lib/releases.js#L137-L138)?

jetersen: Wait why was getFilterExcludedPullRequests and getFilterIncludedPullRequests added when they are not used else.

I had them enclosed because nobody else used them...

darkdreamingdan: > This reads funny...
> 
> Could we not call getFilterIncludedPullRequests() directly in the filtering below on line [137-138](https://github.com/darkdreamingdan/release-drafter/blob/781e89dd260b0fdd18bfddb540005c710ccab74f/lib/releases.js#L137-L138)?

Yes, this is possible.  I just wasn't making significant changes to this function for this PR, and this is superficial.

> Wait why was `getFilterExcludedPullRequests` and `getFilterIncludedPullRequests` added when they are not used else.
> 
> I had them enclosed because nobody else used them...

They are used.  They are re-used in the version-resolver filtering, which is why i've pulled them out.

lib/default-config.js
darkdreamingdan: The RESOLVED_VERSION variable means it was always optional to use the version-resolver, because unless you use the RESOLVED_VERSION template variable in your config.yml then the version won't increment.  

The above change is not related to making version resolver optional.  Because the default config does a deep merge, i had to remove the default labels of 'major', 'minor', 'patch', because it is impossible to remove them if they are in the default config.  I noticed this within unit tests after i updated them.  So by default, $RESOLVED_VERSION will just increment the patch version unless you provide label config.  This is now fixed.

lib/releases.js
jetersen: Right than proceed with removing the const for `filterExcludedPullRequests` and `filterIncludedPullRequests`.

darkdreamingdan: Done

lib/default-config.js
jetersen: Good thing we have unit test 🗡️

darkdreamingdan:
Awesomebeans!  Thanks @jetersen.   All systems go i think 🚀 .  Happy for you to merge.

jetersen:
Marvelous! Dan, thanks for your contribution!

PR #516, opened 2020-05-12
Calculate resolved version based on increment config

pdcmoreira:
This makes the `$RESOLVED_VERSION` template variable to use the next patch/minor/major calculated version, depending on the PR types since the last release.

It is possible to configure the increment in the config file:
```yml
increment:
  major-labels:
    - 'breaking'
    - 'breaking-changes'
  minor-labels: 'feature'
```

So, if there is any PR with the configured major-labels, the `$RESOLVED_VERSION` will be the next major. If not, it checks the same for minor-labels, resolving to the next minor.
In any other case (didn't match those labels or not configured at all), it falls back to the next patch version (the current behaviour in master).

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -78,22 +78,22 @@ const contributorsSentence = ({ commits, pullRequests }) => {
   }
 }
 
-const categorizePullRequests = (pullRequests, config) => {
-  const { 'exclude-labels': excludeLabels, categories } = config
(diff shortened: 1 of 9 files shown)


jetersen:
somewhat overlapping #513

pdcmoreira:
> somewhat overlapping #513

Damn! I've checked for PRs with this a while ago, should've checked again.

jetersen:
superseded by #513

PR #517, opened 2020-05-13
New desc

shrutiverma29:
description added

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -33,6 +33,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    body
                     author {
                       login
                     }
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -3,7 +3,7 @@ const { SORT_BY, SORT_DIRECTIONS } = require('./sort-pull-requests')
 const DEFAULT_CONFIG = Object.freeze({
   'name-template': '',
   'tag-template': '',
-  'change-template': `* $TITLE (#$NUMBER) @$AUTHOR`,
+  'change-template': `* $TITLE (#$NUMBER) @$AUTHOR $DESC`,
   'no-changes-template': `* No changes`,
   'version-template': `$MAJOR.$MINOR.$PATCH`,
   categories: [],
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -144,7 +144,8 @@ const generateChangeLog = (mergedPullRequests, config) => {
         template(config['change-template'], {
           $TITLE: pullRequest.title,
           $NUMBER: pullRequest.number,
-          $AUTHOR: pullRequest.author ? pullRequest.author.login : 'ghost'
+          $AUTHOR: pullRequest.author ? pullRequest.author.login : 'ghost',
+          $DESC: pullRequest.body
         })
       )
       .join('\n')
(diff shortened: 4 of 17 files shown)


.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
+/workspace.xml
```
jetersen, line 2: You need to also remove `.idea` from the PR changes.

lib/default-config.js
```diff
@@ -3,7 +3,7 @@ const { SORT_BY, SORT_DIRECTIONS } = require('./sort-pull-requests')
 const DEFAULT_CONFIG = Object.freeze({
   'name-template': '',
   'tag-template': '',
-  'change-template': `* $TITLE (#$NUMBER) @$AUTHOR`,
+  'change-template': `* $TITLE (#$NUMBER) @$AUTHOR $DESC`,
```
jetersen, line 6: Should not be added to the default, as I think this is optional.

test/fixtures/config/config-with-yaml-exception.yml
```diff
@@ -1,4 +1,4 @@
-change-template: - #$NUMBER '$TITLE' @$AUTHOR
+change-template: - #$NUMBER '$TITLE' @$AUTHOR '$DESC'
```
jetersen, line 1: No need to add it here...

test/schema.test.js
```diff
@@ -26,7 +26,10 @@ const invalidConfigs = [
   [{ template: { '👶': 'a' } }, 'must be a string'],
   [{ template: null }, 'must be a string'],
   [{ template: '' }, 'is not allowed to be empty'],
-  [{ 'change-template': ['* $TITLE (#$NUMBER) @$AUTHOR'] }, 'must be a string'],
+  [
+    { 'change-template': ['* $TITLE (#$NUMBER) @$AUTHOR $DESC'] },
```
jetersen, line 30: add a new test

test/fixtures/config/config-with-resolved-version-template.yml
```diff
@@ -11,7 +11,7 @@ categories:
     label: 'documentation'
   - title: '🧰 Maintenance'
     label: 'maintenance'
-change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
+change-template: '- $TITLE @$AUTHOR (#$NUMBER) $DESC'
```
jetersen, line 14: Should add a new test.

lib/default-config.js
shrutiverma29: Can you suggest an alternate way to add optional description in this template?

jetersen: Simply remove it, and you inside your `release-drafter.yml` file you can override the `change-template` to include description.
Changing the default is a breaking change.

shrutiverma29: But I want to add 'PR Description' which comes from api in body tag.

shrutiverma29: @jetersen Oh yes, it's working. Thanks!

lib/commits.js
```diff
@@ -33,6 +33,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    body
```
jetersen, line 36: this

lib/releases.js
```diff
@@ -144,7 +144,8 @@ const generateChangeLog = (mergedPullRequests, config) => {
         template(config['change-template'], {
           $TITLE: pullRequest.title,
           $NUMBER: pullRequest.number,
-          $AUTHOR: pullRequest.author ? pullRequest.author.login : 'ghost'
+          $AUTHOR: pullRequest.author ? pullRequest.author.login : 'ghost',
+          $DESC: pullRequest.body
```
jetersen, line 148: and this, is the only required change for the `$DESC` being available

shrutiverma29: Thanks @jetersen 
what's the process of making a github action so that I can use the updated code in my private repo?

PR #531, opened 2020-05-23
Allow adding the body of the PR to the release

gavinlove:
We make extensive use of the pull request body and automatically send out the release body to a group of people when the release is deployed.

As such I would like to stop having to copy the PR body manually into the release notes. 

I have added a test however it is not very good because I think I need to be a PR in this [repo]( https://github.com/TimonVS/release-drafter-test-repo) with a body to do a proper test. If there is a way to get that or another way to add a test body please do let me know.

fixes #517

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,7 @@ module.exports = (app) => {
       context,
       branch,
       lastRelease,
+      config,
     })
 
     const sortedMergedPullRequests = sortPullRequests(
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -33,6 +33,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    body
                     author {
                       login
                     }
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -7,6 +7,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
     $name: String!
     $owner: String!
     $branch: String!
+    $withPullRequestBody: Boolean!
     $since: GitTimestamp
     $after: String
   ) {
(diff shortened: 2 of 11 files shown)


jetersen:
Can you please revert all formatting changes?

gavinlove:
Sure, although it was the pre-commit hook that made the changes. I will disable revert.

jetersen:
I would prefer if those style changes are done in a separate pr

gavinlove:
Makes sense!

lib/commits.js
```diff
@@ -33,6 +33,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    body
```
jetersen, line 36: let me regenerate after adding some body to those PRs :)

jetersen:
of course the style PR introduced breaking changes.

gavinlove:
Always the way :(

lib/commits.js
jetersen: Perhaps the GraphQL can optionally only query for this if `change-template` includes `$BODY` thus reducing the amount of data required and reducing the time to response?

index.js
```diff
@@ -40,6 +40,7 @@ module.exports = (app) => {
       context,
       branch,
       lastRelease,
+      config
```
jetersen, line 43: According to style it should be 😅 
```suggestion
      config,
```

jetersen:
Hmm I want to add lint CI to save time with styling 😅

index.js
gavinlove: Doh! yeah, I disabled the hook!

lib/commits.js
gavinlove: Good call on this when I tested it I noticed the difference in the time to respond.

jetersen:
@gavinlove I have added some bodies to the fixtures, can you try and regenerate?

gavinlove:
Will do

PR #538, opened 2020-05-28
Set the release name and tag_name as outputs of the action

alonsohki:
This is necessary if you are planning to execute additional steps after creating the release draft. As the release name and tag are autogenerated, it is essential that the action provides these.


jetersen:
Marvelous! Thanks for your contribution ❤️

PR #541, opened 2020-05-29
Use created_at tag date instead of published_at

### LINKED ISSUE #222: Incorrect timestamp is used when drafting

darkdreamingdan:
Hey again!

This PR is intended to superseed https://github.com/release-drafter/release-drafter/pull/409 .
fixes #409 
fixes #222 

In order to unblock #409, i thought I would try and reduce the number of API calls as you suggested.  In the process of doing this, i realised an extra call isn't even required, and GraphQL is not required to get the tag's commit date.  This property is already available with the existing calls, and minimal code modifications are required.  This is actually stated in the GitHub documentation:

image

This means there is no need to use a `findTagCommitDate` function, because `created_at` is exactly this.

This PR achieves the following:
- Removes the use of `published_at` and instead uses `created_at`.  
- Sorting of releases is also done with `created_at` rather than `published_at`.  I think this was the intention when i look at the code comments just above this section.
- I have also addressed the bug that @JHFarrant described relating to squash merging.  The `since` GraphQL is an inclusive call, so we want to remove any commits from the previous tag.  A test case for this has been added.  This has required a small addition of `committedDate` to the existing GraphQL query for commit information.

Hopefully you'll agree it's a leaner implementation to fix this bug! :) .  Let me know your thoughts :) .


lib/commits.js
```diff
@@ -23,6 +23,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
               }
               nodes {
                 id
+                committedDate
```
jetersen, line 26: Indent looks off

lib/commits.js
```diff
@@ -76,20 +77,25 @@ module.exports.findCommitsWithAssociatedPullRequests = async ({
...
       context.github.graphql,
       module.exports.findCommitsWithAssociatedPullRequestsQuery,
-      { ...variables, since: lastRelease.published_at },
+      { ...variables, since: lastRelease.created_at },
       dataPath
     )
+    // GraphQL call is inclusive of commits from the specified dates.  This means the final
+    // commit from the last tag is included, so we remove this here.
+    commits = _.get(data, [...dataPath, 'nodes']).filter(
+      (commit) => commit.committedDate != lastRelease.created_at
```
jetersen, line 97: Should be greater than?

jetersen: I get it now. The tagged commit is the thing you want to filter out 😅

darkdreamingdan:
Just live tested this on a repo, all good from me. Merge as you like :) .

lib/commits.js
JHFarrant: My only hesitation here is that comparing dates could cause issues in the future. We could filter out more than just the first PR. 

For example, two merges made in the same second unlikely to happen with the GUI but could definitely happen if branches are merged locally using a script and pushed. 

Does this DateTime include milliseconds?

The alternate way is more complicated and would require a second rest API call in `releases.js` to retrieve the reference of the latest release tag. Then one could compare the commit hashes. Something like:
```node
commits = _.get(data, [...dataPath, 'nodes']).filter(
      (commit) => commit.oid != lastRelease.tag_ref
```

jetersen: Honestly I like the tag ref filter more

darkdreamingdan: It is indeed milliseconds so the chance of collision is rare. Reference is possible, but as you mention would require another API call (commit hash is not available in listReleases either by REST or GraphQL) . It is not possible to do this with 2 api calls as we have currently using commit hash. We would end up with a similar solution to #409. My view is that hash is more elegant at the cost of an additional API call. This current solution is more pragmatic and still a clear improvement over what exists on master.

Open to suggestions around way forward.

jetersen: Hmm would prefer to avoid the extra API call over very rare case of exact milliseconds 😅

darkdreamingdan: Just realised I made a mistake, format is timestamp is:
`2019-04-27T13:38:46Z`
So I suppose it is seconds not milliseconds.

jetersen: Even with seconds I would still argue we would need to save the API call.

jetersen:
Marvelous! Thanks for your continued contribution @darkdreamingdan so dreamy! 😍

jetersen:
Hmm kind of related.
I think we would benefit from using the GraphQL for releases as well, since we can ask GitHub to order them: https://developer.github.com/v4/object/repository/#releases

darkdreamingdan:
> Hmm kind of related.
> I think we would benefit from using the GraphQL for releases as well, since we can ask GitHub to order them: https://developer.github.com/v4/object/repository/#releases

I thought a little bit about this.  It only saves a tiny amount of computation in JavaScript.  I originally thought about only grabbing the most recent 2 releases (as you had suggested in #409 ), and then checking if it's draft or non-draft to determine `lastRelease` and `draftRelease`.  

However, the `sortReleases` function has a feature which grabs the latest semantic version.  Grabbing only the last 2 versions would not support this - it checks through all release and finds the latest one by semantic version (and falls back to release date).  GitHub GraphQL only supports ordering by creation date and alphabetical name, not by the tag version semantically (as far as i can tell).  This means `sortReleases` feature to allow you to tag older releases if you follow semantic versioning would stop working.

Hence, unless we want to change behaviour, we have to grab all releases and do ordering within release-drafter rather than let GitHub do it.  So TLDR, the benefit of using GitHub ordering is not really a lot, because release-drafter reorders them with semantic versions anyway for most cases.

jetersen:
I had forgotten about our `sortReleases` method 😓

PR #546, opened 2020-06-01
Support triggable references (branches, tags). Ignore triggable references on GitHub action

### LINKED ISSUE #141: Auto-publish release feature
Since #91 we've been able to get many of our repos using release drafter, with a nice Releases page and automatic patch version bumps. Thanks again for this tool!

One additional feature that we'd be really interested in contributing is automatic _publication_ of releases. For many of our repos we typically want to release a new version after each PR is merged.

I thought of creating a separate probot to do this in the spirit of separation of concerns. However there is no hook for 'draft release created'; Release Drafter, as the creator of the draft release, is really in a unique position of being able to do this.

So I'd like to propose adding a feature that would:

- be _opt-in_ via a new config flag
- require a template for release tag and name
- automatically publish a release after it has been created
- ... if/unless some configurable label is present on the just-merged PR

I'd be keen to get your thoughts on this!

### LINKED ISSUE #203: Release Target Setting

### LINKED ISSUE #404: Multiple release branches and multiple release drafts
First of all, congratulations for this automation. I think it's pure gold.

Our use case in the [Micronaut team](https://github.com/micronaut-projects) is that we create releases not only from `master`, but also from branches such as `1.0.x`, `1.1.x`, etc. In this scenario, release-drafter overwrites whatever draft version it finds, and this isn't desirable in all cases.

Would it be possible to setup different release-drafter configurations so that we could have multiple release drafts at a given moment, and then release-drafter would filter the draft to update based on some regexp or similar?

Maybe this is related to #203 and #175

### LINKED ISSUE #515: Draft based on branch release/v*
I was trying to configure the release-drafter to work with branches. But it doesn't seem to work or I did something wrong. Does it work only with master?

I would like the release-drafter to be executed only e.g.
```
  push:
    branches:
      - 'release/v*' 
```
This happens because the release is done via TeamCity. From there the build happens and etc.

Is it possible?

Cheers.

### LINKED ISSUE #520: Question - Hotfix vs actual develop branch
Been using your action for some times now and I really like it. There's one use case though that I cannot get to work, and I was wondering if I am missing a parameter or something.

Right now I'm working on, lets say, version 1.1.0, every now and then I push a PR to develop and me release is updated with every PR since 1.0.0. Perfect.

Now I have to push a hotfix to production, so I create a branch with my production version, fix my bug and push it to prod. Once I am done I create a release v1.0.2, and I publish it. 

If I do that, the next time I push a PR to develop, Release Drafter will update my draft release with every PR since the last release, so I now only have the PR since I created v1.0.2 and lost all the ones before that. 

Is there a way to tell Release Drafter to ignore a certain tag, or to get everything from develop since the last release in develop. Or will I have to do manual manipulation every time I'll have to do a hotfix in prod? I'm guessing you guys already had to do that kind of thing, so I came to ask for your input. 

If this is not the place to ask question, sorry about that. I didn't find any other place. 

Thanks!

### LINKED ISSUE #527: Question - draft after release deleted is not working
Consider this simple flow:
1. A PR is merged to the default branch
1. Release-drafter builds a new draft release with said PR
1. The Draft Release is published, with a tag is set for it (let's call it v1.1.1)
1. Someone deletes the v1.1.1 release

Now what should happen, ideally, is that a new flow is triggered by using the following statement on the release-drafter.yml file:

```yaml
on:
  push:
    branches:
      - master
  release:
    types: [deleted]
```

However, after doing that, the draft is not being built, without printing any messages whatsoever. By reading the code, I expected at least to see a message similar to `Ignoring push. :branch_name is not one of: [...]`. I've done some tests to check the context on which the action runs after the `release deleted` event, and it seems that it points to the tag (on this case, v1.1.1, same name as the release), but I'm not entirely sure how to deal with that.

Is there any good way to deal with this?

### LINKED ISSUE #542: Published releases doesn't seem to work?
I am testing the https://github.com/release-drafter/release-drafter/pull/424#issuecomment-598830640 because I want to achieve the following: https://github.com/goreleaser/goreleaser-action/issues/207

When using the 5.9.0 release I get the following "##[warning]Unexpected input 'publish', valid inputs are ['entryPoint', 'args']"  when using the example in the PR. I also tried it with ENV but it doesn't work either.

Goal: Use release drafter to draft a release with all the PR's then when the goreleaser releases the code it uses the notes from the draft or converts the draft to original release. But what happens is that the draft stays and goreleaser creates a release next to it, while they both have the same version name. I might be missing something here why it doesn't do what I expected :)

### LINKED ISSUE #545: Ignore branches config when running on GitHub Actions
When release-drafter is run on GitHub Actions they have better way of determining what branch is trigger-able.
Since GitHub action is free real estate, we do not care much about restriction users design on the workflows.

[See the workflow docs about `on`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)

https://github.com/release-drafter/release-drafter/blob/6b5b16358bd49becf320405ee92143d11dcb6869/lib/triggerable-branch.js#L7-L20

relates to 
#515
#542 
#527
#520
#404
#203

### LINKED ISSUE #550: Action on push tag does not work after the release is published.
Action on push tag does not work after the release is published.
I have an action to release a build on push tag action:
image
And I also add the default release-drafter.yml file to prepare release draft. 
Everything works perfectly, except that when I publish this draft my build-installer action does not start. At the same time, I see that the tag has been added. Also, if I create a draft manually before combining the PR, everything works, and the build-installer action starts.
The only difference that I see is that when a draft was created with an action (release-drafter.yml), the user who releases the build is githubactions:
image
Can this be fixed?

jetersen:
While exploring this issue I found many issues that this PR will greatly improve the experience for.

- Allow GitHub action to always trigger, let the GitHub workflow decide the trigger ;)
- GraphQL supports receiving the fully qualified reference, so let us just pass that.
- some dead flatten code, actually handled by Joi.
- Support `ref/tags` when running on GitHub app

cc @toolmantim @TimonVS 

There should be minimal impact as seen in the test changes we support both `branches` and `references` definition.

@arminc this should fully unlock any GitHub workflow you desire.

Note to self: Release notes should make clear `branches` is renamed to `references`

fixes #550 
fixes #545
fixes #515
fixes #542
fixes #527
fixes #520
fixes #404
fixes #203
fixes #141


README.md
```diff
@@ -92,26 +92,41 @@ template: |
...
+| `sort-direction`      | Optional | Sort changelog in ascending or descending order. Can be one of: `ascending`, `descending`. Default: `descending`.                                                          |
+| `prerelease`          | Optional | Mark the draft release as pre-release. Default `false`.                                                                                                                    |
+| `version-resolver`    | Optional | Adjust the `$RESOLVED_VERSION` variable using labels. Refer to [Version Resolver](#version-resolver) to learn more about this                                              |
 
 Release Drafter also supports [Probot Config](https://github.com/probot/probot-config), if you want to store your configuration files in a central repository. This allows you to share configurations between projects, and create a organization-wide configuration file by creating a repository named `.github` with the file `.github/release-drafter.yml`.
 
+## References
```
jetersen, line 115: Should properly move down 😓

jetersen:
@darkdreamingdan would you mind reviewing this PR 🙏

SebastianThiebaud:
This is great. These changes would allow this action to work in a git flow context. Thanks for submitting this PR. I will patiently wait for a merge :)

PR #571, opened 2020-06-20
Update example in README to use version resolution

bpg:
The configuration example in `README.md` has a `version-resolver` section with labels that supposed to control the version name / tag strings. However, this example dos not work out of the box, which could be frustrating, especially for new users.

This tiny PR is to update name/tag templates with `$RESOLVED_VERSION`, so the version labels, if present, will actually be used.


jetersen:
Thanks for catching that 😅

PR #582, opened 2020-07-01
Allow adding the URL of the PR to the release notes

b00men:
We are actively uses a repo's release notes in scope of other repos.  
This makes a github PR's short link (`#NUMBER`) lead to a wrong place.

So I would like to have an option to use an issue's complete URL instead of hardcoded replacements.

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -35,6 +35,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    url
                     body @include(if: $withPullRequestBody)
                     author {
                       login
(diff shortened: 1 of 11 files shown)


lib/commits.js
```diff
@@ -35,6 +35,7 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   nodes {
                     title
                     number
+                    url
```
jetersen, line 38: could we do the same include trick as we did for body? Again to avoid pulling extra data that is not needed.

b00men: It makes sense

b00men: Done

jetersen:
@b00men marvelous, thanks for your contribution!

PR #600, opened 2020-07-22
Make the generated schema compliant with draft04

PierreBtz:
The schema was containing an empty array for the required field, which is incompatible with the draft 04 specification.

I hesitated to simply remove the field as technically we are currently filtering a list containing a single element, however I fear it's error prone for the future.

Tell me if the js code is wrong, js is not really my strong suit, happy :)

### DIFF
--- a/bin/generate-schema.js
+++ b/bin/generate-schema.js
@@ -13,7 +13,13 @@ const jsonSchema = {
 }
 
 // template is only required after deep merged, should not be required in the JSON schema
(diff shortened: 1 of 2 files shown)


PierreBtz:
@jetersen any way to get this PRs moving? I would like to get it to finish up https://github.com/jenkinsci/.github/issues/21.
Thanks!

timja (approved):

PR #632, opened 2020-08-27
Allow more than just a push event to run the action

MindaugasLaganeckas:
We would like to propose a change, where it is the github action determines, when a release-drafter should be run.
We have a usecase, where it is a workflow dispatch event, that triggers the action and this line: ```app.on('push', async (context) => {``` prevents the action to run. 

Maybe there is a better way to solve this problem rather than replacing ```push``` with ```*```. Your input is welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ const log = require('./lib/log')
 const core = require('@actions/core')
 
 module.exports = (app) => {
-  app.on('push', async (context) => {
+  app.on('*', async (context) => {
     const { shouldDraft, configName, version, tag, name } = getInput()
 
     const config = await getConfig({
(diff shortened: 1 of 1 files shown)


jetersen:
Thanks for your contribution. I know this is an simply change. But this would affect the github app application as well.

It would be much preferred to wait for #549

MindaugasLaganeckas:
No problem! Take your time! :smile:

jetersen:
I have an extended weekend coming up, so hopefully soon(tm)

jetersen:
don't mind me, just testing something.

/rebase

jetersen:
👋 @MindaugasLaganeckas I chose to make this conditional to support both use cases.

jetersen (approved):

Issue #656, opened 2020-09-30
Branch specific configuration?


joakime:
It would be nice if release-drafter has the ability to specify multiple configurations with different `target-branch` to manage multiple branches in parallel.

This is a concept seen on other actions, like dependabot.

An example of this for dependabot is can be seen at https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/.github/dependabot.yml

Looks like ...

``` yml
version: 2
updates:
  - package-ecosystem: "maven"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-9.4.x"

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-9.4.x"

  - package-ecosystem: "maven"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-10.0.x"

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-10.0.x"

  - package-ecosystem: "maven"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-11.0.x"

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "jetty-11.0.x"
```

If release-drafter supported this `target-branch` concept, then it could be configured to manage multiple active releases in larger projects that have multiple active branches all with interleaved releases.

mkurz:
This is possible thanks to the `filter-by-commitish` config introduced in #657 / #683.

It took me a while to figure out how things work, that's why I am going to share my learnings. Actually, I set up an example repo which I used as a playground to get things working: https://github.com/mkurz/release-drafter-test/ You can fork that repo and just try yourself.

So let's assume you currently have two branches: a `master` branch (which is your GitHub default branch) and a `2.8.x` branch.
The `master` branch is your dirty development branch in which you work towards a 2.9.0 release - so later you will create a `2.9.x` branch off it, like you did with the `2.8.x` branch long time ago. From the `2.8.x` branch you released versions `2.8.1`, `2.8.2`,... Actually `2.8.0` was tagged on the master branch back then, and only afterwards the 2.8.x branch was created, like you would tag the 2.9.0 release from the master branch and afterwards create the 2.9.x branch from it.

So now let's configure the release drafter:
[...]

PR #657, opened 2020-10-02
filter releases to consider only the target branch 

fujifish:
Consider only releases targeting the same branch as the branch being used for the drafting the new release.
Releases that are for other branches will not be considered for the calculation for the changes.

This also possibly addresses #656

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,7 @@ module.exports = (app) => {
       return
     }
 
-    const { draftRelease, lastRelease } = await findReleases({ app, context })
+    const { draftRelease, lastRelease } = await findReleases({ ref, app, context })
     const {
       commits,
       pullRequests: mergedPullRequests,
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -17,7 +17,7 @@ const sortReleases = (releases) => {
   }
 }
 
-module.exports.findReleases = async ({ app, context }) => {
+module.exports.findReleases = async ({ ref, app, context }) => {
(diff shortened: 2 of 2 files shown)


jeacott1:
@fujifish , maybe this can help? - 
 I'm doing something similar, generating and updating a draft release everytime a PR receives a push event - using a different github action though.
 the naming scheme I'm using is v${maven-semver}-${pullrequest-id}, and I limit activity on that PR to the same branch name (because otherwise other branches that might happen to have the same git sha pushed to them, but with no associated PR will also trigger :(  ) - _does anyone else think github actions are a bit of a mess?_

in my case adding something like this mostly works and I think should work for drafter too (on merge the PR is already closed so you'll need to account for that by allowing finding of closed PRs):

from a push event
using 8BitJonny/gh-get-current-pr in a previous step to find the related PR from the current sha
```
-name: drafter
 if: ${{ steps.PR.outputs.number && (steps.PR.outputs.pr != '') && fromJson(steps.PR.outputs.pr).head.ref == env.GITHUB_REF_NAME }}
 ...
```

jetersen (approved):
This seems to solve an issue with supporting multiple releases.

mkurz:
> This seems to solve an issue with supporting multiple releases.

Yes it does. I wrote a howto with my experience setting up release drafter for multiple branches: https://github.com/release-drafter/release-drafter/issues/656#issuecomment-806184887

PR #667, opened 2020-10-10
switch project to lerna to prepare for splitting code base

jetersen:


jetersen:
😓 I was afraid of this merge conflicts 😆 Hopefully a rebase should do it correctly.

jetersen:
MEH not sure it makes sense to split it.

PR #676, opened 2020-10-18
Annotated tags support when publishing

### LINKED ISSUE #578: ERROR event: Data doesn't contain `nodes` field.
Work well before, but now error occur

Action url: https://github.com/ibuler/koko/runs/820944535
Error:  
```
Make sure the `paginatePath` is set to the field you wish to paginate and that the query includes the `nodes` field. 
(id=67d387bf-cb1c-4407-8d16-b062ecfd4dd9)   Error: Data doesn't contain `nodes` field. Make sure the `paginatePath` is 
set to the field you wish to paginate and that the query includes the `nodes` field.       at paginate (/app/lib/pagination.js:21:11)       at async module.exports.findCommitsWithAssociatedPullRequests (/app/lib/commits.js:88:12)  
     at async /app/index.js:40:9   --   event: {     "event": "push",     "id": "67d387bf-cb1c-4407-8d16-b062ecfd4dd9",     "repository": "ibuler/koko"   }
```

### LINKED ISSUE #615: Error when using annotated tags
The current graphql query [1]  for fetching the commit history for a tag seems to break when the tag is an annotated tag. The following error will occur, as seen in action run in [2]:

```text
07:36:40.424Z ERROR event: Data doesn't contain `nodes` field. Make sure the `paginatePath` is set to the field you wish to paginate and that the query includes the `nodes` field. (id=69466cf8-b0e1-429d-8965-2f68d713fc4c)
  Error: Data doesn't contain `nodes` field. Make sure the `paginatePath` is set to the field you wish to paginate and that the query includes the `nodes` field.
      at paginate (/app/lib/pagination.js:21:11)
      at async module.exports.findCommitsWithAssociatedPullRequests (/app/lib/commits.js:102:12)
      at async /app/index.js:40:9
```

If this query was uplifted to the below it seems to work:

```graphql
    repository(name: $name, owner: $owner) {
      object(expression: $ref) {
          ... on Commit {
            history(first: 100, since: $since, after: $after) {
              totalCount
              pageInfo {
                hasNextPage
                endCursor
              }
              nodes {
                id
                committedDate
                message
                author {
                  name
                  user {
                    login
                  }
                }
                associatedPullRequests(first: 5) {
                  nodes {
                    title
                    number
                    url @include(if: $withPullRequestURL)
                    body @include(if: $withPullRequestBody)
                    author {
                      login
                    }
                    baseRepository {
                      nameWithOwner
                    }
                    mergedAt
                    isCrossRepository
                    labels(first: 10) {
                      nodes {
                        name
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
```

where using the object query directly (instead of ref), seems to produce the almost the same output. would need to refactor this function to:
* object query doesn't use `target` so would need to update accordingly
## References

[1] https://github.com/release-drafter/release-drafter/blob/master/lib/commits.js#L5
[2] https://github.com/castlemilk/graphs/runs/948957120?check_suite_focus=true

timja:
fixes #578
fixes #615

Tested in https://github.com/timja/dummy-repo-drafter/releases

cc @jetersen

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -13,43 +13,41 @@ module.exports.findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
     $after: String
   ) {
     repository(name: $name, owner: $owner) {
-      ref(qualifiedName: $ref) {
-        target {
-          ... on Commit {
-            history(first: 100, since: $since, after: $after) {
-              totalCount
-              pageInfo {
-                hasNextPage
-                endCursor
+      object(expression: $ref) {
+        ... on Commit {
+          history(first: 100, since: $since, after: $after) {
+            totalCount
+            pageInfo {
+              hasNextPage
+              endCursor
+            }
+            nodes {
+              id
+              committedDate
+              message
+              author {
+                name
+                user {
+                  login
+                }
               }
-              nodes {
-                id
-                committedDate
-                message
-                author {
-                  name
-                  user {
+              associatedPullRequests(first: 5) {
+                nodes {
+                  title
+                  number
+                  url @include(if: $withPullRequestURL)
+                  body @include(if: $withPullRequestBody)
+                  author {
                     login
                   }
-                }
-                associatedPullRequests(first: 5) {
-                  nodes {
-                    title
-                    number
-                    url @include(if: $withPullRequestURL)
-                    body @include(if: $withPullRequestBody)
-                    author {
-                      login
-                    }
-                    baseRepository {
-                      nameWithOwner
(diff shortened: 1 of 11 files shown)


timja:
One thing I did notice, not really directly related to this PR is that release drafter when publishing will publish a release based on the version resolvers next expected one, rather than publishing the release against the tag the GitHub action is running against.

Future rfe I think unless there’s some configuration that can make that work now?

jetersen (approved):

PR #683, opened 2020-10-21
Filter releases to consider only the target branch

fujifish:
Filter releases to consider only the target branch, this time by enabling a config option named `filter-by-commitish`.
The default is `false` so previous behaviour is unchanged.

Should fix the breaking changes of #657


fujifish:
Any chance this will get reviewed?

fujifish:
@jetersen, would be great to see this merged sometime soon if possible.
Thanks!

jetersen (approved):
LGTM

PR #711, opened 2020-11-17
Improve example to use default branch

ssbarnea:
As all new projects are using main instead of default we make use of github macro in order to have an example that works regardless which is current default branch.

See https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/ for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ on:
   push:
     # branches to consider in the event; optional, defaults to all
     branches:
-      - master
+      - $default-branch
 
 jobs:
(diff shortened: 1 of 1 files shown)


jetersen (approved):

ssbarnea:
While this was merged and conforming with github documentation I soon realised that apparently it does not work as advertised.

Did anyone observe the same issue?

jetersen:
it only works for workflow templates once you clone them.

PR #758, opened 2021-01-16
Fix incorrect configuration path

milewski:
By default, the path is not within the workflows folder


milewski:
Nevermind I just realized that is referring to the github action, not configuration file

PR #770, opened 2021-01-28
Container compatibility

rofafor:
Github Enterprise doesn't have official Action support yet and the release-drafter needs still to be run in GITHUB_APP mode:

- Fix the Docker entrypoint to actually run without errors
- Avoid creating any Actions output, if we aren't in Actions mode
- Add a healthcheck endpoint for Kubernetes deployments in Github App mode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,4 +11,5 @@ LABEL "com.github.actions.color"="orange"
 WORKDIR /app
 COPY . .
 RUN yarn --frozen-lockfile
-ENTRYPOINT [ "/app/node_modules/.bin/probot", "receive", "/app/index.js" ]
+ENTRYPOINT [ "/app/node_modules/.bin/probot" ]
+CMD [ "receive", "/app/index.js" ]
--- a/index.js
+++ b/index.js
@@ -86,7 +86,9 @@ module.exports = (app) => {
       })
     }
 
-    setActionOutput(createOrUpdateReleaseResponse, releaseInfo)
+    if (runnerIsActions()) {
+      setActionOutput(createOrUpdateReleaseResponse, releaseInfo)
+    }
   })
 }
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -11,4 +11,4 @@ LABEL "com.github.actions.color"="orange"
 WORKDIR /app
 COPY . .
 RUN yarn --frozen-lockfile
-ENTRYPOINT [ "/app/node_modules/.bin/probot", "receive", "/app/index.js" ]
\ No newline at end of file
+ENTRYPOINT [ "/app/node_modules/.bin/probot", "run", "/app/index.js" ]
```
jetersen, line 14: hmm, can you create a new Dockerfile for the `Dockerfile` action to use?
https://github.com/release-drafter/release-drafter/blob/8ef32da3eaa2ce71a765835e12aae337792e4492/docker/action.yml#L5

Perhaps call the file `action.Dockerfile`

Dockerfile
```diff
@@ -11,4 +11,5 @@ LABEL "com.github.actions.color"="orange"
 WORKDIR /app
 COPY . .
 RUN yarn --frozen-lockfile
-ENTRYPOINT [ "/app/node_modules/.bin/probot", "receive", "/app/index.js" ]
\ No newline at end of file
+ENTRYPOINT [ "/app/node_modules/.bin/probot" ]
+CMD [ "receive", "/app/index.js" ]
```
jetersen, line 15: I see, you simply override the command 👏

Dockerfile
rofafor: I splitted the current ENTRYPOINT into ENTRYPOINT + CMD, so the default behaviour is the same. One can overwrite it with a own command like  `run /app/index.js`. Better now?

jetersen: Yup

jetersen (approved):

PR #771, opened 2021-01-29
Add commitish to config

### LINKED ISSUE #734: Consider having commitish as input
Hey there,

I have setup my workflow to prepare and release from a release branch. It would be greate if I could create the release with the commitish set to my release branch. Currently if have to change this manually

subsetpark:
This adds a config entry that allows developers to specify a branch in their config, which will then be passed as the `target_commitish` param. Fixes #734.


action.yml
```diff
@@ -38,6 +38,11 @@ inputs:
       A boolean indicating whether the relase being created or updated is a prerelease.
     required: false
     default: ''
+  commitish:
```
jetersen, line 41: Missing under actions input in the readme: https://github.com/release-drafter/release-drafter#action-inputs

jetersen (approved):
LGTM, one doc issue and than we are good!

action.yml
subsetpark: Thanks for the review, @jetersen ! Resolved in 0836e2a

danielcweber:
Is there a known issue/cause why the target-commitish is being picked up correctly (rel-branch), yet the tag always ends up on the main-branch?

image

PR #774, opened 2021-02-01
Add autolabeler

rofafor:
As release-drafter is catogorizing releases notes based on labels, it's nice to integrate optional labeling support as well. Check the README for configuration.
image
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const { sortPullRequests } = require('./lib/sort-pull-requests')
 const log = require('./lib/log')
 const core = require('@actions/core')
 const { runnerIsActions } = require('./lib/utils')
+const ignore = require('ignore')
 
 module.exports = (app, { getRouter }) => {
   const event = runnerIsActions() ? '*' : 'push'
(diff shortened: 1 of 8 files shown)


jetersen:
Hmm an action already exists for this: https://github.com/actions/labeler
I wonder if you could use the trick I did for https://github.com/release-drafter/release-drafter/blob/master/action.js to get that as a probot?

jetersen:
Hmm, I like where this is heading though and it does seem different enough from actions/labeler.

rofafor:
The actions/labeler supports only file paths, but this one tracks branch naming (for those using gitflow naming scheme), PR titles and descriptions (fix/bug/etc. keywords, ticketing system ids).

index.js
```diff
@@ -21,6 +22,105 @@ module.exports = (app, { getRouter }) => {
     })
   }
 
+  app.on(
+    [
+      'pull_request.opened',
+      'pull_request.reopened',
+      'pull_request.synchronize',
+    ],
```
jetersen, line 30: Have you tested that this also works for GitHub actions?

rofafor: Not yet

jetersen: Would you mind testing it? 😅

lib/template.js
```diff
@@ -52,5 +52,33 @@ function validateReplacers({ context, replacers }) {
...
+          }),
+          title: autolabel.title.map((reg) => {
+            return toRegex(reg)
+          }),
+          body: autolabel.body.map((reg) => {
+            return toRegex(reg)
+          }),
+        }
+      } catch (e) {
+        console.log({
```
RomanovX, line 72:
```suggestion
        log({
```

I think your habit here turned to console.log rather than the built in log

rofafor: Yes. I'm forward-porting these from an old fork, so mistakes happen :)

index.js
rofafor: Will do.

rofafor: Seems to working with my initial tests. I updated the README howto enable.

package.json
```diff
@@ -25,6 +25,7 @@
     "cli-table3": "0.6.0",
     "compare-versions": "3.6.0",
     "escape-string-regexp": "4.0.0",
+    "ignore": "^5.1.8",
```
jetersen, line 28: Please set it fixed and remember to update the `yarn.lock`

```suggestion
    "ignore": "5.1.8",
```

jetersen (approved):
LGTM, one minor nitpick than we are ready to merge.

package.json
rofafor: done

PR #789, opened 2021-02-19
Configurable functionality

rofafor:
The main motivation here is to have some control which events are dispatched to the actual "releaser" code as it gets processed with every event now: configuring the release-drafter for `pull_request` only, makes it generate draft releases as well.

This can now be controlled via action inputs for both "releaser" and "autolabeler" independently and both functionalities are enabled by default.


jetersen (approved):
LGTM, makes total sense.

rofafor:
Is there something preventing the merge?

jetersen:
Nope, just me being forgetful 😅

Issue #315, opened 2019-10-18
Date based releases?


whyman:
We use a scheme that is pretty common of using the day of a release to tag it.

What work would need to be done to have this happen?

casz:
You would need to implement some more logic here:

https://github.com/toolmantim/release-drafter/blob/864af063817df9c43c6f538364eec24d9fc3b1bc/lib/versions.js#L21-L34

shadoath:
@jetersen Any example of how to make this happen?

jetersen:
@shadoath currently not supported.
You would need to add the logic here:
https://github.com/release-drafter/release-drafter/blob/864af063817df9c43c6f538364eec24d9fc3b1bc/lib/versions.js#L21-L35

I imagine to would check if the template contains something like `$DATE` and than use the an new function that knows how to convert `$DATE` into a date version.
It should properly check the last release to support `20210210.1`, `20210210.2`, `20210210.3` so that you can release multiple times on the same date.

jetersen:
Someone sad to see all these 👍 and no one implementing as it should be fairly simple to add.

0rax:
We implement a quick hack in a fork of ours here https://github.com/moneway/release-drafter/commit/82edd3e3a06863eab9797d1818a31690cbc907f1 to allow some version of calendar versioning in `release-drafter` though this is based on quite some old code.

I will take a look at updating this implementation and creating a PR here for that as I am currently migrating away from our fork and webhook based system and was going to fix this by implementing a custom version resolver as part of a Github Actions pipeline. If this can help others, it's better to have this as a feature instead of a hack.

As the change was pretty trivial, it might be quick to re-implement there, though, as far as I remember, I wasn't able to update the version number when new PRs came in.

0rax:
This issue is not as trivial as it seems in the end. Handling everything present on https://calver.org/ is pretty hard. Mainly handling formatting for incremental version specifier (i.e: MICRO version after a date such as `2021.01_1` for a `YYYY.0M_MICRO` template).

Handling a simple date-based system seems pretty easy, the addition of other identifier after that makes it way harder (especially with optional identifier that are only present when needed).

jetersen:
Starting with something simple? Than if someone needs it they can contribute :)

0rax:
I might just commit what I have right now, which works pretty well and just does not handle optional identifier. I just have one small bug to squash out !

0rax:
[...]

PR #807, opened 2021-03-05
[WIP] Add support for calendar versioning

0rax:
This was created following the request and discussion in #315.

I am currently doing some "real world" tests to validate that this works.

I think this PR also lacks some configuration example and some documentation in the README that I will add later.


shadoath:
What is the best way to test this in the 'real world'?

jetersen:
@shadoath point your action at their branch `moneway:feature/calver-tmpl` so
```yaml
jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
      - uses: moneway/release-drafter@feature/calver-tmpl
```
The benefit of GitHub actions 👏

jetersen:
@0rax any plans to finish it? :)

0rax:
@jetersen, sadly I didn't have time to work on this lately and the code here is pretty much incompatible now with the current codebase.

Someone on my team also found out that most of the work made here was pretty much useless has we can just generate the version using `date` in a Github actions such as:

```
name: Release Drafter

on:
  push:
    branches:
      - main

jobs:
  draft-release:
    runs-on: ubuntu-20.04
    steps:
      - name: Generate CalVer version
        id: calver
        run: |
          export VERSION=$(date "+%Y.%m.%d")
          echo ::set-output name=version::${VERSION}
          echo "Version set to ${VERSION}"
        with:
          tag: ${{ steps.calver.outputs.version }}
          name: ${{ steps.calver.outputs.version }}
          version: ${{ steps.calver.outputs.version }}
        env:
```

I would definitely suggest anyone to use something like this for the simple reason that it can support any arbitrary CalVer format without having to implement them manually in release-drafter (CalVer being a pretty permissive in its format).

The only issue here is that it doesn't handle "MINOR" version (i.e: `2022.02.12_1`) which might make auto publishing of version impossible in some cases. Though in my case, as I do push releases manually from drafts, I can detect the potential conflict easily.

Here is how this would look like with a Draft and a Release on the same "version number" when using Release Draft:
image

And here is how Github will warn you that a tag already exist, prompting you to add the "MINOR" suffix manually.
image

To be completely honest, there might event be a world where we could use the https://github.com/actions/github-script to detect the existing tag or create an github-action addon that would just handle CalVer tag generation which might be useful to more people.

jetersen:
Yup combining actions are great 😊

jetersen:
@0rax I think you could easily fix `2022.02.12_1` via `gh release view --json tagName` :)

jetersen:
Closing this as @0rax mentioned this can be solved by generating output in a step before release-drafter.

Issue #835, opened 2021-04-07
"ValidationError: child "template" fails because ["template" is required]\n'


JustinGrote:
https://github.com/JustinGrote/Press/actions/runs/727435035

I've tried every combination of just the workflow with no config, specifying the config with only template, etc. but I always get this message.

jetersen:
The config has to be on the default branch.

JustinGrote:
@jetersen thank you! makes testing a little tricky but it's not a big deal. I submitted a PR to clarify that in the readme. Thanks.

SebSept:
Damned ! I wish I had this information 2 hours ago :/

kreempuff:
@jetersen is there a reason it **must** be on the default branch? This has bitten me quite a few times, it'd be nice if it was configurable. Happy to contribute a change if it's not a GitHub limitation

jetersen:
> @jetersen is there a reason it **must** be on the default branch? This has bitten me quite a few times, it'd be nice if it was configurable. Happy to contribute a change if it's not a GitHub limitation

Not a github limit. Could be patched.

JustinGrote:
Ran into this again, drives me crazy I can't test my release drafters and have to iterate on main to get it working

kreempuff:
@JustinGrote I'll tackle this when I get some time in the near future 😬

jetersen:
> > @jetersen is there a reason it **must** be on the default branch? This has bitten me quite a few times, it'd be nice if it was configurable. Happy to contribute a change if it's not a GitHub limitation
> 
> Not a github limit. Could be patched.

Patching it would result in security issues.

JustinGrote:
@jetersen why would it be a security issue? PRs won't have a sufficient GITHUB_TOKEN to draft a release, and I can control who creates branches in my repo and what branch those actions enact. Only people who could "abuse" this would be people who can draft their own releases anyways, maybe you can clarify?

jetersen:
Person with repo write permission can modify the action to run draft and create a release and most likely release a binary with nasty bits in it.

GITHUB token has permission to both draft and publish releases as the action even offers to publish a release.

I prefer that the action requires to modified on the default branch as that usually provides extra safe guards with branch protection.

PR #836, opened 2021-04-07
Docs: Clarify config must be on default branch

JustinGrote:
Resolves https://github.com/release-drafter/release-drafter/issues/835

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ If you're unable to use GitHub Actions, you can use the Release Drafter GitHub A
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


bob-bins:
Can we get this merged in? Its omission in the current docs is very painful for first time users

martinda:
Hello @JustinGrote . Could you also please change the first line in the README file where it says the `yml` file is in `.github/workflows/release-drafter.yml` because that does not work. The file must be in `.github/release-drafter.yml`.  Thanks.

jetersen (approved):

PR #865, opened 2021-05-14
Use commitish from action inputs. Fixes #824.

### LINKED ISSUE #824: Tag always ends up on the main-branch
Hi,

first of all, thanks for this great workflow.

I develop on the "main" branch. On a release, I create a branch "rel" and bump the version (in a version.json), thereby creating a new commit). "rel" is now one commit ahead of "main".

I override the "tag" and the "commitish" parameter of the workflow. Commitish seems to successfully pickup "rel" as the branch to work on. Yet, the tag always ends up on "main".

My workflow is [here](https://github.com/ExRam/ExRam.Gremlinq/blob/main/.github/workflows/build.yml#L191). The latest run is [this](https://github.com/ExRam/ExRam.Gremlinq/runs/2204557867?check_suite_focus=true). Note that I have forcefully pushed the 8.4.0-tag to be on the right commit.

Thanks a lot.

danielcweber:
The commitish parameter was previously ignored from action inputs and only read from the static config.yml, when the Readme and action.yml advertised the parameter as action inputs.

Fix by @MariuszKogut

Fixes #824

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,6 +131,7 @@ module.exports = (app, { getRouter }) => {
       tag,
       name,
       disableReleaser,
+      commitish,
     } = getInput()
 
     const config = await getConfig({
@@ -181,6 +182,7 @@ module.exports = (app, { getRouter }) => {
       name,
       isPreRelease,
       shouldDraft,
+      commitish,
     })
 
     let createOrUpdateReleaseResponse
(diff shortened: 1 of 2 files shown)


jamesnswithers:
Works well for me on creating a new draft release but it doesn't seem to update the target commitish of existing draft releases.

jetersen (approved):

PR #892, opened 2021-06-10
add `env-var-template-variables` option

tjenkinson:
wip

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,27 +103,28 @@ template: |
 
 You can configure Release Drafter using the following key in your `.github/release-drafter.yml` file:
 
-| Key                    | Required | Description                                                                                                                                                                |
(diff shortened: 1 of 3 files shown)


jetersen:
I don't think this makes sense as you could potentially extract secrets.

jetersen:
I did but the potential is still there and not everybody has configured their branch with branch protection and etc.
Even the potential for a bad actor that gets access to a GitHub account and can suddenly extract all org secrets that are available for a repo or repo secrets.
Basically this is just a big NO NO.

jetersen:
There are already protection for logging secrets so that is less likely.
Of course they could setup a server and send the secrets to it.
I just don't think ENV is the right way.

If you want further customization you could potentially use the action output body and then use something as simple as `envsubst` inside your action.

I don't want to support this use case. In your workflow you can add any steps to achieve this.

Issue #569, opened 2020-06-18
$CONTRIBUTORS variable takes contributors from a skipped PR


curquiza:
Hello everyone, thanks for your useful tool!!! 😄 
In the [meilisearch organization](https://github.com/meilisearch/), we use release-drafter with GHA on most of our repositories for a few weeks and we are really satisfied!

I used to skip the PR of Dependabot thanks to the `exclude-labels:` field. It works perfectly: the PR is indeed not added to the release draft.

But the `dependatbot` contributor is still there.

<img width="1385" alt="Capture d’écran 2020-06-18 à 10 09 52" src="https://user-images.githubusercontent.com/20380692/84995087-defce280-b14b-11ea-93ad-5a1b03ec4a81.png">

Is it possible to include only contributors from non-skipped PR in the `$CONTRIBUTOR` variable?

FYI, here is my template:

```yaml
name-template: 'v$RESOLVED_VERSION 🐍'
tag-template: 'v$RESOLVED_VERSION'
exclude-labels:
  - 'skip-changelog'
version-resolver:
  minor:
    labels:
      - 'breaking-change'
  default: patch
categories:
  - title: 'Breaking changes'
    label: 'breaking-change'
template: |
  ## Changes

  $CHANGES

  Thanks again to $CONTRIBUTORS! 🎉
no-changes-template: 'Changes are coming soon 😎'
sort-direction: 'ascending'
```

Sorry if it's a duplicated issue but I found nothing.

Have a nice day!

jetersen:
for a quick workaround you could use the replacers to get read of him 😅

jetersen:
This regex would mostly do the job: `(?:and )?@dependabot-preview(?:\[bot\])?,?`

curquiza:
Hello! Thanks for your quick answer.
I tried this, but it does not work. I probably did something wrong.
My template:

```yaml
name-template: 'v$RESOLVED_VERSION 🐍'
tag-template: 'v$RESOLVED_VERSION'
exclude-labels:
  - 'skip-changelog'
version-resolver:
  minor:
    labels:
      - 'breaking-change'
  default: patch
categories:
  - title: 'Breaking changes'
    label: 'breaking-change'
template: |
  ## Changes

  $CHANGES

  Thanks again to $CONTRIBUTORS! 🎉
no-changes-template: 'Changes are coming soon 😎'
sort-direction: 'ascending'
replacers:
[...]

PR #898, opened 2021-06-19
Added exclude-contributors and no-contributors-template configuration options

ThomasKasene:
As I mentioned in #569, I think it would be nice if we, as repository owners, would be able to remove ourselves from the list of contributors, so that we can more easily highlight 3rd party contributions.

I'm no JS developer myself and know precious little about Node, but I've made a best effort at adding a configuration option named `exclude-contributors` which removes specified usernames from the `$CONTRIBUTORS` variable.

I decided to not exclude the names that come from `commit.author.name` because if I understand the API correctly, they are not necessarily unique.

PS: I'm not sure if I committed all the necessary files, so that's something that the reviewer should double-check. I'm also not sure if I set up my development environment properly; there was one test suite that failed, even before I started making my changes:
```
 test/config.test.js
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```


ThomasKasene:
Actually, I just realized that since I'm filtering values away, there's a chance the `else`-block here will end in an `ArrayIndexOutOfBoundsException` or something similar:
```javascript
  const sortedContributors = Array.from(contributors).sort()
  if (sortedContributors.length > 1) {
    return (
      sortedContributors.slice(0, sortedContributors.length - 1).join(', ') +
      ' and ' +
      sortedContributors.slice(-1)
    )
  } else {
    return sortedContributors[0]
  }
```

I suppose it ought to be re-written into something like this:
```javascript
  } else if (sortedContributors.length == 1) {
    return sortedContributors[0]
  } else {
    // What here?
  }
```
but then, what of the remaining branch? Should I also add a new configuration option called something like `no-contributors-template` that can be used here?

ThomasKasene:
I wrote the code for a `no-contributors-template` option too, but I'm waiting for feedback from @jetersen or somebody else before I commit it to this branch, since they might want it as a separate pull request.

jetersen:
could simply change it to the following

```javascript
  } else if (sortedContributors.length == 1) {
    return sortedContributors[0]
  }
```

The method does not require to return any value.

ThomasKasene:
Interesting. I suppose that's another piece of weirdness of JavaScript. In that case, it is certainly an option, but what will the result be if the function doesn't return anything? If it's an empty String, then I think I would still like to be able to customize the message. Consider the following `template`:

```
template: |
  ## Contributors
  $CONTRIBUTORS
```

If all the contributors are filtered away, the release notes would contain the "Contributors" header, but nothing underneath it, which would look a little awkward.

But if you wish, I can add your suggestion to this branch, and submit a separate pull request for a `no-contributors-template` option?

jetersen:
Perhaps move all the header, filtering and sorting together and simply check if length is zero after contributors are filtered. Feel free to push to this PR otherwise you are making it hard for me to review ☺️

ThomasKasene:
Maybe we're misunderstanding one another, but I was talking about the `template` option, not a `contributors-template` (which currently doesn't exist). Maybe I could've been more clear and written a complete example:
```
exclude-contributors:
  - dependabot

template: |
  $CHANGES

  ## Contributors
  $CONTRIBUTORS
```

If a release is being made that only contains changes by `dependabot`, then the last part of the release notes would contain the "Contributors" header, but nothing else below it.

Given the same changes by `dependabot`, but with the following configuration, then at least there would be a little message under the header so it wouldn't look so empty:
```
exclude-contributors:
  - 'dependabot'

no-contributors-template: 'No third-party contributions this release.'

template: |
  $CHANGES

  ## Contributors
  $CONTRIBUTORS
```

I think these two options (`exclude-contributors` and `no-contributors-template`) are two that can be reviewed separately because they don't really affect one another (apart from the fact that they both add `config` to the signature of `contributorsSentence`), but it's up to you! 😃

jetersen:
@ThomasKasene could you try and build with yarn to reduce the changes in dist folder?

`yarn prettier && yarn lint --fix && yarn build`

ThomasKasene:
@jetersen Absolutely! I'll take a look at it soon.

Do you want me to undo the spacing on the "Configuration Options" table in README.md as well? My lines ended up being a bit long and so I decided to expand the "width" of the other rows as well, but I realize now that it might cause unnecessary merge conflicts.

jetersen:
@ThomasKasene the linter will take care of it.

ThomasKasene:
Okay, nothing happened when I ran your suggested commands. I guess that means I either did it correctly by hand, or that the linter wasn't applied. Either way, dist/index.js got a little better now.

jetersen:
`git merge master` & `yarn install` is important :)

Anyhow ended up doing it for ya

jetersen (approved):

PR #918, opened 2021-08-10
Fix prerelease string getting removed and add $COMPLETE version variable

tjenkinson:
Currently if the version is a prerelease the prerelease string is discarded because `semver.coerce` discards it.

With this PR we now use `semver.parse` instead and fall back to coerce.

It also updates `splitSemVer` to always return a string instead of sometimes a `SemVer` instance (which currently becomes a string in `template` due to it having a `toString`) and sometimes string.

Finally it adds a `$COMPELTE` variable that can be used to access the full version string, otherwise it's still not possible to get the prerelease info in the template.

### DIFF
--- a/lib/versions.js
+++ b/lib/versions.js
@@ -18,6 +18,7 @@ const splitSemVer = (input, versionKey = 'version') => {
     $MAJOR: semver.major(version),
     $MINOR: semver.minor(version),
     $PATCH: semver.patch(version),
+    $COMPLETE: version,
   }
 }
 
--- a/lib/versions.js
+++ b/lib/versions.js
@@ -5,12 +5,9 @@ const splitSemVer = (input, versionKey = 'version') => {
     return null
   }
 
-  let version
-  if (input.inc) {
-    version = semver.inc(input[versionKey], input.inc, true)
-  } else {
-    version = input[versionKey].version
-  }
+  const version = input.inc
+    ? semver.inc(input[versionKey], input.inc, true)
+    : input[versionKey].version
 
   return {
     ...input,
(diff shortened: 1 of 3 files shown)


test/versions.test.js
```diff
@@ -36,14 +36,20 @@ describe('versions', () => {
...
+        name: 'Some release',
+      },
+      null,
+      'v10.0.3-alpha'
+    )
 
     expect(versionInfo.$NEXT_MAJOR_VERSION.version).toEqual('11.0.0')
     expect(versionInfo.$NEXT_MINOR_VERSION.version).toEqual('10.1.0')
-    expect(versionInfo.$NEXT_PATCH_VERSION.version).toEqual('10.0.4')
+    expect(versionInfo.$NEXT_PATCH_VERSION.version).toEqual('10.0.3')
```
tjenkinson, line 50: This was previously wrong. The next patch version after `v10.0.3-alpha` is actually `10.0.3` not `10.0.4`, but it was becoming `10.0.4` before because `10.0.3-alpha` was incorrectly becoming `10.0.3` internally

jetersen:
@tjenkinson A PR merged resulted in some conflict 😅 Could you please take a look?

lib/versions.js
```diff
@@ -5,16 +5,20 @@ const splitSemVer = (input, versionKey = 'version') => {
...
 
-  const version = input.inc
-    ? semver.inc(input[versionKey], input.inc, true)
-    : semver.parse(input[versionKey])
+  let version
+  if (input.inc) {
+    version = semver.inc(input[versionKey], input.inc, true)
+  } else {
+    version = input[versionKey].version
+  }
```
jetersen, line 13: Could this be made an const again?

jetersen (approved):

PR #940, opened 2021-09-08
Added a gray stroke around the letters of the logo

### LINKED ISSUE #937: Github dark mode and image in readme
In github dark mode the image in the `README.md` looks like this:

image

In dark high contrast it's worse:

image

ThomasKasene:
I agree with the points made in #937 and made a version of the logo that has a gray stroke around the letters. I think I was able to find a color and a stroke width that is relatively unoffensive in all the available GitHub themes, but I'm open to other suggestions too 😃 

I'm attaching some screenshots of what it looks like for convenience, but if you wish to test yourself, it'll be viewable in the fork:
https://github.com/ThomasKasene/release-drafter/tree/readme-logo-border

Default dark:
default_dark

Dark dimmed:
dark_dimmed

Dark high contrast:
dark_high_contrast

Default light:
default_light

This closes #937


ThomasKasene:
I should also mention, I have not used the logo.sketch to make this change, nor have I generated an updated logo.png. I'm uncertain whether either of these files are even desirable to keep?

jetersen:
@ThomasKasene perhaps we can do better?

https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981 ?

ThomasKasene:
I'm not 100% sure I understand what you mean. Are you saying that the gray color I picked is good, but that we can have two different SVG images as the logo (ie. using the `#gh-dark-mode-only` or `#gh-light-mode-only selectors` as described [here](https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/77))?

jetersen:
I think the grey color is a good solution. We can go with that.

I was asking you if you thought we could do better with the new selectors. Perhaps instead of grey border you wanted to use white color or something else.

ThomasKasene:
Honestly, based on what I read in that issue thread, using Github's selectors poses an "issue" for those who's using any markdown renderer other than Github's renderer (they will see both logos instead of only one).

I'm not the owner of this project, but if it were up to me, I would not have wanted to use that solution, personally.

PR #973, opened 2021-11-06
Mod error message when invalid config

nowsprinting:
Added the following messages to the error when an invalid configuration file:

1. This message will also be output if the configuration file is not found
2. configuration file must reside in your default branch

~~I wanted to add a message for when a config file was not found, but it was difficult due to its structure, so I decided not to.~~ fixed

~~Two failure tests in  config.test.js, but that is unrelated to this fix.~~ fixed

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -24,8 +24,12 @@ module.exports.getConfig = async function getConfig({ context, configName }) {
       log({
         context,
         message:
-          'Config validation errors, please fix the following issues in release-drafter.yml:\n' +
-          joiValidationErrorsAsTable(error),
+          'Config validation errors, please fix the following issues in ' +
+          (configName || DEFAULT_CONFIG_NAME) +
+          ':\n' +
+          joiValidationErrorsAsTable(error) +
+          '\nNote: This message will also be output if the configuration file is not found. ' +
+          'And the configuration file must reside in your default branch.',
       })
     }
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,6 +1,5 @@
 const core = require('@actions/core')
 const { validateSchema } = require('./schema')
-const { DEFAULT_CONFIG } = require('./default-config')
 const log = require('./log')
 const { runnerIsActions } = require('./utils')
 const Table = require('cli-table3')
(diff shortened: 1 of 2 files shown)


jetersen:
@nowsprinting can you use remove node_modules and rebuild using yarn?
`yarn prettier && yarn lint --fix && yarn build`

lib/config.js
```diff
@@ -24,8 +24,12 @@ module.exports.getConfig = async function getConfig({ context, configName }) {
       log({
         context,
         message:
-          'Config validation errors, please fix the following issues in release-drafter.yml:\n' +
-          joiValidationErrorsAsTable(error),
+          'Config validation errors, please fix the following issues in ' +
+          (configName || DEFAULT_CONFIG_NAME) +
+          ':\n' +
+          joiValidationErrorsAsTable(error) +
+          '\nNote: This message will also be output if the configuration file is not found. ' +
```
jetersen, line 31: Should be possible to say `if (repoConfig == null)` ?

nowsprinting:
Sorry, I mistake. plz wait.

lib/config.js
nowsprinting: I tried it, and it seems that `repoConfig` does not become null even if the config file does not exist.

jetersen:
https://github.com/probot/probot/blob/718025d161b30875db26894af8cc24cf73b1f148/src/context.ts#L210-L212

We could avoid passing in the default config and merge the default config in our self.

nowsprinting:
Thanks, @jetersen 
Fixed it, and fixed existing tests using spy.

jetersen (approved):

lib/config.js
```diff
@@ -24,7 +31,9 @@ module.exports.getConfig = async function getConfig({ context, configName }) {
       log({
         context,
         message:
-          'Config validation errors, please fix the following issues in release-drafter.yml:\n' +
+          'Config validation errors, please fix the following issues in ' +
```
jsoref, line 34: If the error relates to the file on the default branch, then this error message should mention that.

jetersen: That is done exactly here:
https://github.com/release-drafter/release-drafter/blob/0bc3bc59906b1dac1b608cc7ff5441fcb86ad6c9/lib/config.js#L15-L21

jsoref: That's if it's missing.

If it's found and invalid, which branch did it read?

jetersen: Then joiValidationErrorsAsTable(error) will print out the errors?

jsoref: But will it mention the branch?

jetersen: We can't really know which "branch" as the config gets merged. So it could be in default branch or on the `.github` repo you need to fix the config.

jsoref: Ah.

Could I convince you to reparse the configurations separately when you get an error? This wouldn't slow down the default path (beyond a slightly larger file), but would result in a much better user experience.

jetersen: I don't want to create a custom version of https://github.com/probot/octokit-plugin-config to satisfy slightly better user experience. User should be able to comprehend release-drafter config setup. They are the once setting it up. You can keep it as simple as you want or as complex as you want.

Again potentially you could have endless `_extends` like 10 levels deep if you wanted.

Issue #971, opened 2021-11-03
commitish configuration option not present in schema.json file


menno-ll:
The [Configuration options in the readme](https://github.com/release-drafter/release-drafter#configuration-options) state that we can use the `commitish` property to define which branch a release is created for.
However, looking at the [`schema.json`](https://github.com/release-drafter/release-drafter/blob/master/schema.json) file, it's not present there.

jetersen:
Feel free to open a PR.

menno-ll:
Thereyougo! :)
https://github.com/release-drafter/release-drafter/pull/989

PR #989, opened 2021-12-14
Add commitish configuration option to schema.json

menno-ll:
fixes https://github.com/release-drafter/release-drafter/issues/971

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -102,6 +102,10 @@
       "default": false,
       "type": "boolean"
     },
+    "commitish": {
+      "default": "",
+      "type": "string"
+    },
     "replacers": {
       "default": [],
       "type": "array",
--- a/schema.json
+++ b/schema.json
@@ -103,8 +103,16 @@
       "type": "boolean"
     },
     "commitish": {
-      "default": "",
-      "type": "string"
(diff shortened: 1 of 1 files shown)


schema.json
```diff
@@ -102,6 +102,10 @@
       "default": false,
       "type": "boolean"
     },
+    "commitish": {
+      "default": "",
+      "type": "string"
+    },
```
jetersen, line 108: I forgot we actually are able to generate this using `yarn generate-schema` and this is what it generates based on the `/lib/schema.js`
```suggestion
    "commitish": {
      "anyOf": [
        {
          "type": "string",
          "enum": [""]
        },
        {
          "default": "",
          "type": "string"
        }
      ]
    },
```

jetersen, line 108: I forgot we actually are able to generate this using `yarn generate-schema` and this is what it generates based on the `/lib/schema.js`
```suggestion
    "commitish": {
      "anyOf": [
        {
          "type": "string",
          "enum": [""]
        },
        {
          "default": "",
          "type": "string"
        }
      ]
    },
```

menno-ll:
🙌

PR #997, opened 2021-12-28
Add tag-prefix config option

mxj4:
This config option makes it possible to filter previous releases by git tag name prefix, thus makes it possible to identify tags for a single component in a monorepo.

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -3,6 +3,7 @@ const { SORT_BY, SORT_DIRECTIONS } = require('./sort-pull-requests')
 const DEFAULT_CONFIG = Object.freeze({
   'name-template': '',
   'tag-template': '',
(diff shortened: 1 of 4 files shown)


masashi-sutou:
@mxj4 

- This PR sounds good for support monorepo.
- Why is the PR status still draft? Do you still need to add any features?

mxj4:
Sorry I moved to something else since release drafter do not work with our release workflow (need to release from any commit in repo, not just master HEAD). I see you already started with another similar PR, if you still need this I can turn it into a published PR.

Kledal:
Looks like the original question was implemented in https://github.com/release-drafter/release-drafter/pull/1092

jetersen:
fixed in #1092

PR #1013, opened 2022-01-16
add support for moving uncategorized to a category with no labels

### LINKED ISSUE #139: Add $CATEGORIES as a template variable
Currently the categories will just be appended to the end of the template. I would like to control where in the template they go. So if we had a `$CATEGORIES` variable that could go in the `template` string that would solve the problem.

An example
```yml
name-template: v$NEXT_PATCH_VERSION
tag-template: v$NEXT_PATCH_VERSION
categories:
  - title: Bug Fixes
    label: fix
change-template: '- $TITLE (#$NUMBER) @$AUTHOR'
template: |
  $CATEGORIES

  ## Other Changes

  $CHANGES
```

would result in this:

```md

## Bug Fixes

- fix: example (#2) @smyrick

## Other Changes

- docs: other change (#3) @smyrick
```

### LINKED ISSUE #551: Allow to define category order/priority
This is a feature request to allow to set the category order/priority.

Very important to be able to move more important categories like "Breaking Changes" to the top of the changelog.

jetersen:
fixes #139
fixes #551

👏 Finally adding support for moving the categorized group

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -129,6 +129,10 @@ const categorizePullRequests = (pullRequests, config) => {
     return { ...category, pullRequests: [] }
   })
 
+  const uncategorizedCategoryIndex = categories.findIndex(
+    (category) => category.labels.length === 0
+  )
+
   const filterUncategorizedPullRequests = (pullRequest) => {
     const labels = pullRequest.labels.nodes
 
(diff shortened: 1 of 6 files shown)


mehdihadeli:
@jetersen Hi,
Is there any example or guideline for this? I can't find anything about its setup.

jetersen:
@mehdihadeli The test fixture shows the use case.

You can also see it here: https://github.com/jenkinsci/.github/blob/47b68b30f98b9608f8dea2683ac54757002ff47b/.github/release-drafter.yml#L56

PR #1039, opened 2022-01-21
Add support to ignore pre-releases

### LINKED ISSUE #579: What's changed since latest **public**
Now the list of changes is formed from all PRs merged since the last release, no matter whether it's a pre-release or public release.
It may be useful to compile a list of changes since latest **public** release. To keep track of all items that are going to be publicly released.

```yaml
...

use-latest-public: true # defaults to false

categories:
  - title: '🚀 Features'
...
```

chrgernoe:
Fixes #579.

I need this feature for another project and therefore implemented the solution as proposed in the issue. Because this is my first time I contribute to the project and I'm new to JavaScript, I need a bit guidance. I also would like to implement a test but I'm a bit lost what the best strategy is and how this framework works.


jetersen:
@chrgernoe could you try and use `yarn install` to install `node_modules` and `yarn prettier && yarn lint --fix && yarn build` to build the dist/index.js ?

lib/releases.js
```diff
@@ -45,6 +48,12 @@ const findReleases = async ({ ref, context, config }) => {
     log({ context, message: `No draft release found` })
   }
 
+  if (preRelease) {
+    log({ context, message: `Pre-release: ${preRelease.tag_name}` })
+  } else {
+    log({ context, message: `No pre-release found` })
+  }
```
jetersen, line 55: This should not log if `use-latest-public` is true?

jetersen:
I think you would need to add an test inside `index.test.js` where payload fixture only has prereleases and one with a mix of prerelease and releases. Than test the combination of `use-latest-public`

chrgernoe:
> @chrgernoe could you try and use `yarn install` to install `node_modules` and `yarn prettier && yarn lint --fix && yarn build` to build the dist/index.js ?

Yes, I already did this but didn't checked it in since it was just a draft. I will do this as soon the PR is ready.

chrgernoe:
> I think you would need to add an test inside `index.test.js` where payload fixture only has prereleases and one with a mix of prerelease and releases. Than test the combination of `use-latest-public`

OK. Thanks for the hint. I will try to implement this and also extend the doc.

chrgernoe:
What is the best way to generate a new GraphQL file for the following sequence, which I need for testing the new feature:

* Release 1
* PR1
* PR2
* Pre-Release 2
* PR3
* PR4
* Release 2

Currently there is only a GraphQL file for `PR1 ... PR4` available.

ssbarnea:
I someone will have time to revive this because that is currently an annoying problem with release drafter as every time we decide to create a pre-release we endup with incomplete release notes for the follow-up release, requiring us to do a lot of copy/paste to restore the full changelog.

As stated in the title, any pre-release should not be counted when building the changelog.

jetersen:
I know @blast-hardcheese and I had a discussion for rendering everything for a prerelease since the last actual release.

blast-hardcheese (approved):
This looks like it would work, the logic looks correct and what I would have expected in an implementation. Thanks for fixing this!

chrgernoe:
@jetersen: The test is still open because of the question I asked above but unfortunately received no answer.

jetersen:
Should be able to look at existing tests and copy one of the fixtures and add a prerelease.
Another option is to prepare one of the test repos with a case and add it to the fixture generator.

One of the reasons to no answer is driving pull request over the finish line requires a village 😅
Sometimes my focus is elsewhere. Right now it is on V6 of release drafter

PR #1049, opened 2022-02-01
Expand vars in tag and name inputs

### LINKED ISSUE #1048: $RESOLVED_VERSION not expanded when set via action input
Using the `name` input:
```yml
...
jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
        with:
          name: "$RESOLVED_VERSION"
...
```
the release draft will be named `$RESOLVED_VERSION`, which obviously is not what I expected.

However inside the config file
```
name-template: '$RESOLVED_VERSION'
```
the release draft will correctly be named with the version, e.g. `1.0.0`.

Now, the `name` overrides the `name-template`:
<kbd></kbd>

I expect both configs behave identically, I would have guessed variable expansion takes places after the override happened, so it shouldn't matter where the config string came from.

mkurz:
Fixes #1048

### DIFF
--- a/test/fixtures/config/config-with-name-and-tag-template.yml
+++ b/test/fixtures/config/config-with-name-and-tag-template.yml
@@ -1,3 +1,3 @@
 template: Placeholder with example. Automatically calculated values based on previous releases are next major=$NEXT_MAJOR_VERSION, minor=$NEXT_MINOR_VERSION, patch=$NEXT_PATCH_VERSION.
-name-template: 'v1.2.3 (Code name: Hello World)'
-tag-template: v3.4.5
+name-template: 'v$RESOLVED_VERSION-beta (Code name: Hello World)'
+tag-template: v$RESOLVED_VERSION-beta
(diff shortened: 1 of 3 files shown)


jetersen (approved):
Looks good to me.
Could you add a test.

You should be able to find a mention of `GITHUB_ACTION` inside `index.test.js` via search.

mkurz:
@jetersen Done. Tests added (after some problems do get things running :wink:). Can you merge?

test/fixtures/config/config-with-name-and-tag-template.yml
```diff
@@ -0,0 +1,3 @@
+template: Placeholder with example. Automatically calculated values based on previous releases are next major=$NEXT_MAJOR_VERSION, minor=$NEXT_MINOR_VERSION, patch=$NEXT_PATCH_VERSION.
+name-template: 'v1.2.3 (Code name: Hello World)'
+tag-template: v3.4.5
```
jetersen, line 3: ? Could we use some $RESOLVED_VERSION so at least the test looks sane 😓 I cannot handle that your getting one version in name and a different in tag.

jetersen (changes requested):
😓 Test output looks odd, would prefer the config was aligned and the test overrides the name or tag properly.

mkurz:
Better? What I want to show is that once the tag gets overriden, once the name.

jetersen (approved):

jetersen:
> Better? What I want to show is that once the tag gets overriden, once the name.

Much 😅 The tests clearly shows what your looking for 🙂

jetersen:
Thank you for your contribution 🙌

mkurz:
Thank you!

PR #1050, opened 2022-02-01
Add header and footer

### LINKED ISSUE #1047: Add header and footer template configs
When having multiple release drafts a nice feature would be `header` and `footer` templates. These should work exactly like the `template` config (you can also think of it like splitting the original `template` config into three parts).
The use case we have is that we want the same footer on each release note, no matter which repository in our organization gets released. In the footer we could thank our premium sponsors, add additional generic information (like support contact addresses), etc. Because different repos have different release notes bodies, we could put a `base.yml` release notes in our `.github` repo that contains the `footer` template and then we could just `_extend` from that base tempate, still giving each release draft the possibility to have it's own custom `template` config.
Right now we would have to add this information to each release draft, which obviously isn't really nice (specially if you have to add premium sponsors, you have to updated each draft instead of just one base draft).

mkurz:
Fixes #1047

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ You can configure Release Drafter using the following key in your `.github/relea
 | Key                        | Required | Description                                                                                                                                                                |
 | -------------------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `template`                 | Required | The template for the body of the draft release. Use [template variables](#template-variables) to insert values.                                                            |
+| `header`                   | Optional | Will be prepended to `template`. Use [template variables](#template-variables) to insert values.                                                                           |
+| `footer`                   | Optional | Will be appended to `template`. Use [template variables](#template-variables) to insert values.                                                                            |
 | `category-template`        | Optional | The template to use for each category. Use [category template variables](#category-template-variables) to insert values. Default: `"## $TITLE"`.                           |
 | `name-template`            | Optional | The template for the name of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                     |
 | `tag-template`             | Optional | The template for the tag of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                      |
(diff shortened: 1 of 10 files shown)


mkurz:
@jetersen After adding tests this is ready to review and hopefully merge. Thanks!

jetersen (approved):

jetersen:
Again thank you for your contribution 🙌

mkurz:
Again thank you!

jetersen:
I'll put a tag on first thing tomorrow 😉 but bedtime is calling 😊

mkurz:
:+1: Actually there is a bug regarding `_extends` in https://github.com/probot/octokit-plugin-config, I am working on a fix right now, so maybe we can wait with the release. Let's talk tomorrow.

mkurz:
@jetersen That is the bug I was talking about which also affects release-drafter: https://github.com/probot/octokit-plugin-config/issues/166
Fix is here: https://github.com/probot/octokit-plugin-config/pull/167

jetersen:
whether dependency supports .github repo _extends or not have nothing to do with header and footer? 🤔

mkurz:
Actually we want to put footer in a main config and a couple of configs will extend from it, all placed within the `.github` repo. So for us we need both to work correctly, so that's why I thought lets wait for the `octokit-plugin-config` fix and put everything in one release.
However of course technically they are not related.
AFAIK releasing is just pushing tag, feel free to do so. However as soon as the octokit-plugin-config fix lands it would be nice to cut another release. Thanks!

jetersen:
a release is only a tag away so meh 😅

mkurz:
@jetersen so we have a new release of octokit-plugin-config: https://github.com/probot/octokit-plugin-config/releases/tag/v1.1.5
Can you release also release a new relase drafter version with that dependency? And also move the v5 tag? Thanks!

jetersen:
@mkurz https://github.com/release-drafter/release-drafter/releases/tag/v5.18.1

mkurz:
@jetersen Thank you very much! Have a nice weekend!

PR #1058, opened 2022-02-07
Only consider commits before `commitish`, if provided

### LINKED ISSUE #1053: Only consider commits that come before `commitish`?
As I understand right now, only branch names are properly supported as values for `commitish`, having something like a commit sha there does not yield expected results. This is basically a suggestion to support at least commit hashes there (and by extension perhaps everything else?).

Imagine a `master` branch with commits `A -> B -> C (HEAD)` (all of them PR merge commits to keep it simple), and a workflow for that branch which is roughly "test -> build -> run release-drafter". Each of the commits triggers this workflow, and runs for successive commits are not concurrent.
So if the workflow for `B` is running, and `C` is merged meanwhile, I would still like release-drafter to:
- create/update the release targeting commit `B`, since that's what's tested and built so far. Using `commitish: ${{ github.sha }}` works great for creating and https://github.com/release-drafter/release-drafter/pull/1052 enables the same for updating.
- include only `A` and `B` in the release notes. Instead, `C` is also included even though it's not yet "ready", and as I understand there's currently no way to tell it otherwise.

Does this make any sense? :)

Having poked around the code a little I am tempted to try something like `after = commitish` in this query:
https://github.com/release-drafter/release-drafter/blob/master/lib/commits.js#L20
But I wonder if I'm thinking in the right direction.

mikeroll:
Resolves #1053 

When searching for commits to be included in the release notes, if `commitish` is supplied, it is first resolved to a commit sha. That commit sha is then used a starting point for the graphql query. This allows us to generalize the process and support both "static" (commits, tags) and "dynamic" (branches) `commitish` values.
This does not affect `target_commitish` of the release, which will always be the value provided, with no modifications.

Demo repository is here:
https://github.com/mikeroll/release-drafter-1053/releases
https://github.com/mikeroll/release-drafter-1053/actions
Each of the 4 actions correspond to release-drafter configured differently. Added @jetersen to contributors if you want to play around, any of those actions may be restarted to see the different behaviours.
`no-commitish` - uses the repository's default branch, as before. The action is for an old commit, but even if re-run it will include all the later commits in the release, as before.
`commitish-as-branch` - same as above, with the branch specified explicitly.
`commitish-as-sha{,-2}` - the sha of the commit being built (`${{ github.sha }}`) is set as the release target (already worked before) and only the commits up to and including this one are included in the release notes, i.e. the build for commit 3 does not add commit 4.

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,6 +167,7 @@ module.exports = (app, { getRouter }) => {
       await findCommitsWithAssociatedPullRequests({
         context,
         ref,
+        up_to_commitish: commitish,
         lastRelease,
         config,
       })
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -68,6 +68,7 @@ const findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
 const findCommitsWithAssociatedPullRequests = async ({
   context,
   ref,
+  up_to_commitish,
   lastRelease,
   config,
 }) => {
@@ -84,17 +85,35 @@ const findCommitsWithAssociatedPullRequests = async ({
   const dataPath = ['repository', 'object', 'history']
   const repoNameWithOwner = `${owner}/${repo}`
 
+  log({ context, message: `Fetching commits for reference ${ref}` })
+
+  // Resolving commitish to a commit sha so that we can use it as a GraphQL cursor
+  let up_to_commit
+  if (up_to_commitish) {
+    let resolved_commitish = await context.octokit.repos.getCommit({
+      owner,
+      repo,
+      ref: up_to_commitish,
+    })
+    up_to_commit = resolved_commitish.data.sha
+    log({ context, message: `Including commits up to ${up_to_commit}` })
+  }
+
   let data, commits
   if (lastRelease) {
     log({
       context,
-      message: `Fetching all commits for reference ${ref} since ${lastRelease.created_at}`,
+      message: `Including commits since ${lastRelease.created_at}`,
     })
 
     data = await paginate(
       context.octokit.graphql,
       findCommitsWithAssociatedPullRequestsQuery,
-      { ...variables, since: lastRelease.created_at },
+      {
+        ...variables,
+        since: lastRelease.created_at,
+        after: up_to_commit && `${up_to_commit} 0`,
+      },
       dataPath
     )
     // GraphQL call is inclusive of commits from the specified dates.  This means the final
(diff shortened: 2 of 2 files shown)


mkurz:
I think you completely misused the `commitish` config. The `commitish` config solely exists to set the `target_commitish` property of a github release. So basically that property means "this release was done from this ref" (ref usually being a branch).
Now what you do is you take this config, use it for something completely different. If you create a release with this pull request applied and use the `commitish` config for your "up-to-commit" concept, this will end up creating a release whose meta data ( = target_commitish property which got set via release drafters' commitish config) are not correct. I don't even know if GitHub accepts a `target_commitish` which is not a branch but just some sha.
Do you know what I mean? If not please have a look in the code what the `commitish` config is used for right now, it is NOT used to determine anything which release or commit to use. It just gets passed through to add meta data to a GitHub release.

mikeroll:
> Now what you do is you take this config, use it for something completely different. If you create a release with this pull request applied and use the commitish config for your "up-to-commit" concept, this will end up creating a release whose meta data ( = target_commitish property which got set via release drafters' commitish config) are not correct.

Please see the diff - nothing about releases there, `commitish` is only resolved to a sha when searching for commits to include. 
Having `commitish: release-branch` will still set `target_commitish: "release-branch"` on the release because that's what you asked, no magic.

The actual change is this: if I want my release/tag on `release-branch`, might as well search and include commits which are on that same `release-branch` (this PR), and not on the repository's default branch (current behaviour), no? If `release-branch` is the default branch - nobody notices, but if it's not - stuff breaks down as you described in #1061, simply because it's looking in the wrong place.

If I could do `after: "release-branch 0"` in the GraphQL query, I would, but it needs a commit hash so I give it one - of commit that `release-branch` points to. 

> I don't even know if GitHub accepts a `target_commitish` which is not a branch but just some sha.

The [definition](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish) of `commitish` is roughly "anything that points to a specific commit" as per Git, and Github seems to play along with that as far as I looked, and yes `target_commitish` also works perfectly.

mikeroll:
I was actually wrong - the commit search is fine right now, and the PR does not fix anything regarding that. It just adds the possibilty to stop at a certain commit, if that's what's set as `commitish`.

mikeroll:
Closing in favor of #1065

PR #1062, opened 2022-02-08
Sync docker/action.yml with action.yml

mkurz:


jetersen:
I wonder if we could write a test for it? I miss it everytime 😢

mkurz:
Maybe. It's not my priority right now. If I find time yes.

Issue #935, opened 2021-09-02
Only include commits that contain modifications to a path prefix


blast-hardcheese:
I'd like to have multiple configurations of `release-drafter` in order to maintain multiple distinct versions in the same repository.

`release-drafter` does permit configuring a tag prefix, as well as custom template locations, so the only missing component is the ability to choose which commits to include.

From the CLI, I'm looking at something like:

```
user$ git log --oneline -- module/foo/src/main  # This would be tracked in foo-v1.2.3
user$ git log --oneline -- module/bar/src/main  # This would be tracked in bar-v2.3.4
```

I'm happy to do the work here, just opening this issue to announce intent.

blast-hardcheese:
I've got a proof of concept set up, I'll clean up the code and submit a PR after I've used it for a bit in my project

jetersen:
newer version supports multiple `_extends` for configs levels

`filter-by-commitish` should help you

blast-hardcheese:
@jetersen It seems as though `filter-by-commitish` would be used to maintain multiple individual branches, but I'm trying to release multiple multiple modules, colocated in `master` -- am I misunderstanding something here?

jetersen:
So do you release all modules at once in the same release, ie what your looking for is grouping releases inside the draft according to path.

Or would you want each module to be in an individual release making separate release drafts?

blast-hardcheese:
The latter -- Each module is binary compatible with the others, so if the modules are structured as a tree (`core` branching out to leaf modules with highly specific functionality), I should be able to bump the leaf modules without having to release all other modules.

It may be easiest to see how I've got things set up in my project:

- [Releases](https://github.com/guardrail-dev/guardrail/releases)
- [`.github/`](https://github.com/guardrail-dev/guardrail/tree/master/.github) containing configs
- [config section I added to filter PRs containing paths](https://github.com/guardrail-dev/guardrail/blob/master/.github/release-drafter-scala-http4s.yml#L1-L7)

and, if necessary,

- [my fork of `release-drafter` to add the functionality I was after](https://github.com/release-drafter/release-drafter/compare/349214e6b3b889d8b333c012cc61a1f1753baf40...blast-hardcheese:master)

jetersen:
@blast-hardcheese some of these changes looks fine. Perhaps if you could prepare smaller pull requests for the separate bits we can get them into this repo?

jetersen:
[...]

PR #1084, opened 2022-02-25
include-paths: Permit filtering only PRs that modify path prefixes

blast-hardcheese:
#### This PR implements part of #935

---

This introduces a new GraphQL query, requesting all node IDs that contain modifications to the specified file list. That node list is then used to filter the full set of PRs to only the set that contain those node IDs, thus giving us only the PRs that include changes to our paths.

Examples of `include-paths`:

```yaml
- src/main
- README.md
- doc/
```

For a project with ancillary files that need not be considered for a release (an automated job that regenerates binary files on an automated basis, generating lots of noise, for example).

In my case, this is because I have many parallel instances of release-drafter, each running with different configurations, to release different binary artifacts from a single source tree. The project is organized this way for ease of contribution, as well as reducing needless version bumps for consumers (where different source trees contain completely unrelated changesets).

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -2,6 +2,33 @@ const _ = require('lodash')
 const { log } = require('./log')
 const { paginate } = require('./pagination')
 
+const findCommitsWithPathChangesQuery = /* GraphQL */ `
+  query findCommitsWithPathChangesQuery(
+    $name: String!
+    $owner: String!
+    $targetCommitish: String!
+    $since: GitTimestamp
+    $after: String
+    $path: String
+  ) {
+    repository(name: $name, owner: $owner) {
+      object(expression: $targetCommitish) {
+        ... on Commit {
+          history(path: $path, since: $since, after: $after) {
+            pageInfo {
+              hasNextPage
+              endCursor
+            }
+            nodes {
+              id
+            }
+          }
+        }
+      }
+    }
+  }
+`
+
 const findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
   query findCommitsWithAssociatedPullRequests(
     $name: String!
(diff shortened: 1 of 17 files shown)


test/index.test.js
```diff
@@ -2118,6 +2119,59 @@ describe('release-drafter', () => {
...
+        )
+        .reply(200, graphqlPathsMergeCommit)
+
+      nock('https://api.github.com')
+        .get(
+          '/repos/toolmantim/release-drafter-test-project/releases?per_page=100'
+        )
+        .reply(200, [])
+
+      //      nock('https://api.github.com')
```
jetersen, line 2145: uncomment and use `yarn jest --updateSnapshot`

The test would write itself and data would be taken from the two query replies.

lib/commits.js
```diff
@@ -2,6 +2,34 @@ const _ = require('lodash')
 const { log } = require('./log')
 const { paginate } = require('./pagination')
 
+const findCommitsWithPathChangesQuery = ({ includePaths }) => /* GraphQL */ `
```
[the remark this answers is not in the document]
blast-hardcheese: I've done a bit of research into this and don't see a way to do this without inlining.
`history(path: String, ...` includes the following documentation:

> path ([String](https://docs.github.com/en/graphql/reference/scalars#string))
> 
> If non-null, filters history to only show commits touching files under this path.

which doesn't support an array. Attempting to unroll this and do the filter in JS would be possible, but instead of issuing an individual GraphQL query, we'd end up needing to round-trip to the API for each desired `path`.

What could be done is to transform the whole `... on Commit` in the main `findCommitsWithAssociatedPullRequests` into a fragment, parameterized by `path`, where if `includePaths` is an empty array we just pass `null`.

The downside there is that we'd then need to interpolate the resulting array of results ourselves, instead of what the current implementation does, selecting a group of node IDs that can be used to filter the array of returned PRs without altering the order.

I'm by far not an expert on either GraphQL itself, or GitHub's exposed GraphQL API, so I may have missed something and would be quite happy to revisit this approach, string-interpolating into a GraphQL query is undesirable from a number of perspectives, I just didn't see a way to do it without asking for support for this functionality from GitHub themselves.

Thanks for taking a look

test/fixtures/config/config-with-include-paths.yml
```diff
@@ -0,0 +1,6 @@
+template: Placeholder with example. Automatically calculated values based on previous releases are next major=$NEXT_MAJOR_VERSION, minor=$NEXT_MINOR_VERSION, patch=$NEXT_PATCH_VERSION. Manual input version is $INPUT_VERSION.
```
jetersen, line 1: How about including `$CHANGES` into the template, I assume a PR made it into the changes block? 🤔

Properly better to use something like:

```
template: |
  # What's Changed
  $CHANGES
```

blast-hardcheese: That clicked what was missing about my tests, hah! Finally got the snapshot tester working as well, I actually believe this PR is good, other than style or structure nits. The functionality is there, I've tested it in my own test harness.

jetersen: Luckily style and structure is taken care by our js tooling as well 👏 eslint, prettier, lint-staged and husky 👏

jetersen (approved):

jetersen:
@blast-hardcheese Thanks for your contribution! 👏

jetersen:
@blast-hardcheese We forgot to regenerate the schema.json 😄

`yarn generate-schema` should do the trick 😅

blast-hardcheese:
@jetersen Sorry about that! https://github.com/release-drafter/release-drafter/pull/1094 is up

Kledal:
@blast-hardcheese thank you so much for this 👍

blast-hardcheese:
@Kledal I didn't realize others were waiting for this functionality too! Great to hear 🙂

PR #1089, opened 2022-02-28
Filter by `commitish` regardless of the `refs/heads` prefix

mikeroll:
Fixes the problem from https://github.com/release-drafter/release-drafter/pull/1065#issuecomment-1049911908

@robbinjanssen appreciate if you could test this!

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -43,8 +43,14 @@ const findReleases = async ({
 
   log({ context, message: `Found ${releases.length} releases` })
 
+  // `refs/heads/branch` and `branch` are the same thing in this context
+  const headRefRegex = /^refs\/heads\//
+  const targetCommitishName = targetCommitish.replace(headRefRegex, '')
   const filteredReleases = filterByCommitish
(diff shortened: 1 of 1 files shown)


lib/releases.js
```diff
@@ -43,8 +43,14 @@ const findReleases = async ({
 
   log({ context, message: `Found ${releases.length} releases` })
 
+  const headRefRegex = /^refs\/heads\//
   const filteredReleases = filterByCommitish
-    ? releases.filter((r) => targetCommitish.match(`/${r.target_commitish}$`))
+    ? releases.filter(
+        (r) =>
+          // `refs/heads/branch` and `branch` are the same thing in this context
+          targetCommitish.replace(headRefRegex, '') ===
```
jetersen, line 51: This replace can be done once instead of for reach release.

mikeroll: Of course :)

robbinjanssen:
@mikeroll seems to work, tested it locally!  👍

jetersen (approved):

PR #1092, opened 2022-03-03
Tag prefix

blast-hardcheese:
Extending #1084 with the rest of #935

- Include a configuration option to filter the draft tag by a static string prefix

A benefit of this is that if you use a `tag-template: v$RESOLVED_VERSION` and `tag-prefix: v`, prerelease versioning will be supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ You can configure Release Drafter using the following key in your `.github/relea
 | `category-template`        | Optional | The template to use for each category. Use [category template variables](#category-template-variables) to insert values. Default: `"## $TITLE"`.                                   |
 | `name-template`            | Optional | The template for the name of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                             |
 | `tag-template`             | Optional | The template for the tag of the draft release. For example: `"v$NEXT_PATCH_VERSION"`.                                                                                              |
+| `tag-prefix`               | Optional | A known prefix used to filter release tags. For matching tags, this prefix is stripped before attempting to parse the version. Default: `""`                                       |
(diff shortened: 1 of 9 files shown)


jetersen:
feel free to rebase :)

blast-hardcheese:
@jetersen Alright, ready. I'm available if there are any changes required, but I did remember to regenerate schema this time. Thank you in advance!

jetersen (approved):

mkurz:
> A benefit of this is that if you use a `tag-template: v$RESOLVED_VERSION` and `tag-prefix: v`, prerelease versioning will be supported

Can you please explain (and maybe give an example) how prerelease versioning will be supported that way? Maybe I just don't get it...

blast-hardcheese:
@mkurz Heh, I think I need to edit that comment. When using the tag prefix versioning, (eg: `foo-v1.2.3-rc1`), prerelease versioning [can't be parsed](https://github.com/blast-hardcheese/release-drafter/blob/09fa0295085c61ef6fa6d4b06a606b0de680d37a/lib/versions.js#L135).

Compare
```
> semver.parse('abc-v1.2.3-rc1')
null
> semver.parse('1.2.3-rc1')
SemVer {
  major: 1,
  minor: 2,
  patch: 3,
  prerelease: [ 'rc1' ],
...
```

I didn't see any special-casing for stripping the leading `v` out of something like `v1.2.3-rc1`, so I presumed that I'd just not hit that edge case yet, since the fallback after `semver.parse` is `semver.coerce`:

```
> semver.coerce('foo-1.2.3-rc1')
SemVer {
  major: 1,
  minor: 2,
  patch: 3,
  prerelease: [],
...
```

I was operating under the incorrect assumption that _any_ string prefix to the version passed into `semver.parse` would cause `coerce` to be used, which I now see is incorrect.

Sorry for the confusion, certainly it doesn't seem necessary if you're using `vX.Y.Z-rcA`-style versioning.

Issue #1093, opened 2022-03-03
feature: Collapsible categories


robbinjanssen:
In some releases our "dependencies" category is huge because of many JS dependencies. It would be nice if there would be an option to make some categories collapsed by default.

Example config:
```yml
categories:
  - title: "New features"
    labels:
      - "new-feature"
  - title: "Dependency updates"
    collapsed: true
    labels:
      - "dependencies"
```

Example output:

## New features

- pr1 Added a collapsed boolean to category config

##  Dependency updates
<details>
  <summary>4 changes</summary>

  - pr2 Bump dep1 from 6.17.9 to 6.18.0
  - pr3 Bump dep1 from 6.18.0 to 6.18.1
  - pr4 Bump dep1 from 6.18.1 to 6.18.2
  - pr5 Bump dep1 from 6.18.2 to 6.18.3
</details>

Is this something you would be willing to accept as an PR?

jetersen:
Great suggestion!

That would be a good first PR :)

robbinjanssen:
I like your idea, if we take it even further we can make it configurable and set the default to  > 3? 

Any suggestion for the name of the configuration?

```yml
categories:
  - title: "Dependency updates"
    collapsed: true
    collapse_at: 3
    labels:
      - "dependencies"
```

robbinjanssen:
Released in https://github.com/release-drafter/release-drafter/releases/tag/v5.19.0 :)

jetersen:
https://cs.github.com/?scopeName=All+repos&scope=&q=collapse-after+path%3A.github%2Frelease-drafter.yml nice to see active usage in 9 places already 😄

robbinjanssen:
I first noticed it being used in [home-assistant/supervisor](https://github.com/home-assistant/supervisor/releases) which I use (and am a big fan of) myself! Pretty cool 😄  We use it in all our (private) repo's.

I've got another feature in mind, but not sure about it yet. 

I would like to merge these:
- #3010 Bump @babel/core from 7.17.5 to 7.17.6 [@dependabot](https://github.com/dependabot)
- #3020 Bump @babel/core from 7.17.6 to 7.17.7 [@dependabot](https://github.com/dependabot)
- #3030 Bump @babel/core from 7.17.5 to 7.17.8 [@dependabot](https://github.com/dependabot)

into:
- #3010,#3020,#3030 bump @babel/core from 7.17.5 to 7.17.8 [@dependabot](https://github.com/dependabot)

[...]

PR #1095, opened 2022-03-04
Add a `collapse-after` option for categories

robbinjanssen:
This implements a new feature as discussed in #1093 

Add the `collapse-after` integer to a category to make it hide the PRs if there are more than the given int.

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -246,9 +246,11 @@ const generateChangeLog = (mergedPullRequests, config) => {
     })
     const pullRequestString = pullRequestToString(category.pullRequests)
 
-    if (category.pullRequests.length <= category['collapse-after']) {
-      changeLog.push(categoryTitle, '\n\n', pullRequestString)
-    } else {
+    const shouldCollapse =
+      category['collapse-after'] != 0 &&
+      category.pullRequests.length > category['collapse-after']
+
+    if (shouldCollapse) {
       changeLog.push(
         categoryTitle,
         '\n\n',
@@ -260,6 +262,8 @@ const generateChangeLog = (mergedPullRequests, config) => {
         '\n',
         '</details>'
       )
+    } else {
+      changeLog.push(categoryTitle, '\n\n', pullRequestString)
     }
 
     if (index + 1 !== categorizedPullRequests.length) changeLog.push('\n\n')
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -247,7 +247,7 @@ const generateChangeLog = (mergedPullRequests, config) => {
     const pullRequestString = pullRequestToString(category.pullRequests)
 
     const shouldCollapse =
-      category['collapse-after'] != 0 &&
+      category['collapse-after'] !== 0 &&
       category.pullRequests.length > category['collapse-after']
 
     if (shouldCollapse) {
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -178,8 +178,8 @@ const categorizePullRequests = (pullRequests, config) => {
     .filter(getFilterIncludedPullRequests(includeLabels))
     .filter((pullRequest) => filterUncategorizedPullRequests(pullRequest))
 
-  categorizedPullRequests.map((category) => {
-    filteredPullRequests.map((pullRequest) => {
+  for (const category of categorizedPullRequests) {
+    for (const pullRequest of filteredPullRequests) {
       // lets categorize some pull request based on labels
       // note that having the same label in multiple categories
       // then it is intended to "duplicate" the pull request into each category
@@ -187,8 +187,8 @@ const categorizePullRequests = (pullRequests, config) => {
       if (labels.some((label) => category.labels.includes(label.name))) {
         category.pullRequests.push(pullRequest)
       }
-    })
-  })
+    }
+  }
 
   return [uncategorizedPullRequests, categorizedPullRequests]
 }
@@ -237,11 +237,10 @@ const generateChangeLog = (mergedPullRequests, config) => {
     changeLog.push(pullRequestToString(uncategorizedPullRequests), '\n\n')
   }
 
-  categorizedPullRequests.map((category, index) => {
+  for (const [index, category] of categorizedPullRequests.entries()) {
     if (category.pullRequests.length === 0) {
-      return category
+      continue
     }
-
     const categoryTitle = template(config['category-template'], {
       $TITLE: category.title,
     })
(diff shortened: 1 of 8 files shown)


jetersen:
@robbinjanssen Please use `yarn install` and `yarn prettier && yarn lint --fix && yarn build`

using `npm` causes these large changeset in `dist/index.js`

lib/releases.js
```diff
@@ -237,17 +237,36 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
+      continue
+    }
+    const categoryTitle = template(config['category-template'], {
+      $TITLE: category.title,
+    })
+    const pullRequestString = pullRequestToString(category.pullRequests)
+
+    if (
+      !category.collapse ||
+      (category.collapse && category.pullRequests.length < 4)
```
jetersen, line 251: If you think this needs to be configurable, feel free to add it :)

Please follow the current config naming with `word-another-word`

lib/releases.js
```diff
@@ -237,17 +237,36 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
+    ) {
+      changeLog.push(categoryTitle, '\n\n', pullRequestString)
+    } else {
       changeLog.push(
-        template(config['category-template'], { $TITLE: category.title }),
+        categoryTitle,
+        '\n',
```
jetersen, line 257: Heading should have double newline.
```suggestion
        '\n\n',
```

lib/releases.js
```diff
@@ -237,17 +237,36 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
+        '<details>',
+        '\n',
+        `<summary>${category.pullRequests.length} changes</summary>`,
         '\n\n',
```
jetersen, line 261: This should also stay for better rendering inside markdown between summary and content

lib/releases.js
robbinjanssen: We could, but I have no idea what to name it 🙈  Any suggestions?

jetersen:
@rnorth @bsideup @kiview This might be interesting for you :) I see you used `details` on release notes for https://github.com/testcontainers/testcontainers-java

If you have any input that would be appreciated 😅

lib/releases.js
jetersen: `collapse-at` is fine it will go together with `collapse`. The readme would have to document whether it is greater than or less than that is used.
The default may be greater than 3?

rnorth:
Wow this is great! I've been doing this by hand and it's such a pain. Thank you!

The only input I'd offer would be around `collapse-at`. I definitely think this value being configurable makes a lot of sense. 

I wonder if having _both_ `collapse` and `collapse-at` is potentially confusing though. The behaviour can differ depending on whether one/both/other are set, and the user has to mentally internalise some boolean logic to know what will happen.

Would it be too radical to suggest having _only_ `collapse-at`?

If not set, the behaviour would be to never collapse (use `MAX_VALUE` as the default). If set, `collapse` would be implicitly true when the length exceeds the value.

jetersen:
Only having `collapse-at` makes a lot of sense 😄

robbinjanssen:
Great suggestions! collapse-at definitely makes sense, i'll try to update the pr as soon as i got some spare time!

jetersen:
one more suggestion, perhaps `collapse-after` makes more sense :) So `collapse-after: 3` would mean `pullRequests.length > 3` would mean it collapses once it hits 4 😅

robbinjanssen:
@jetersen done! Let me know if you want any more changes :-)

lib/schema.js
```diff
@@ -111,6 +111,10 @@ const schema = (context) => {
           Joi.object()
             .keys({
               title: Joi.string().required(),
+              'collapse-after': Joi.number()
+                .integer()
+                .min(0)
+                .default(Number.MAX_VALUE),
```
jetersen, line 117:
```suggestion
                .default(0),
```

lib/releases.js
```diff
@@ -237,17 +237,33 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
+    if (category.pullRequests.length <= category['collapse-after']) {
+      changeLog.push(categoryTitle, '\n\n', pullRequestString)
+    } else {
       changeLog.push(
-        template(config['category-template'], { $TITLE: category.title }),
+        categoryTitle,
+        '\n\n',
+        '<details>',
+        '\n',
+        `<summary>${category.pullRequests.length} changes</summary>`,
```
jetersen, line 257: I think it would be best to swap the if and else statement around. According to my suggestion with using 0 for collapse-after.

So you could stay

```javascript
const shouldCollapse = collapseAt != 0 && category.pullRequests.length > collapseAt

if (shouldCollapse) {
  changeLog.push(
    categoryTitle,
    '\n\n',
    '<details>',
    '\n',
    `<summary>${category.pullRequests.length} changes</summary>`,
  )
} else {
  changeLog.push(categoryTitle, '\n\n', pullRequestString)
}
```

robbinjanssen: Sure I can swap the `if`, but the problem with `0` as default value is that you cannot collapse all items. What if I want a category to always be collapsed?

jetersen: If you want it to always collapse, you use `collapse-after: 1` it would not collapse for one but for two PRs. I think that is a sensible behavior. Why collapse at exactly one item, you might as well just show that one item.

robbinjanssen: In your scenario the default value should be `-1`, then if you define `collapse-after: 0` all items will be collapsed

jetersen: Also you would have to fix the plural of `number changes` in the summary.

robbinjanssen: Valid point, let me update the PR :-)

jetersen: Note the rendering of the following markdown.

One shows that it does not make sense to always collapse if you only have 1 PR as it would take up the same size as two PRs
Two double news lines after summary matters for the rendering of the list 😄

### Deps

- Test

### Deps

<details>
<summary>1 changes</summary>

- Test

</details>

### Deps

<details>
<summary>1 changes</summary>
- Test
</details>

### Deps

<details>
<summary>2 changes</summary>

- Test
- Test

</details>

```
### Deps

- Test

### Deps

<details>
<summary>1 changes</summary>

- Test

</details>

### Deps

<details>
<summary>2 changes</summary>

- Test
- Test

</details>
```

jetersen:
41b8bf1abc4729929c6e25f512a6f36f210dc7f5 great addition 😄

lib/releases.js
robbinjanssen: Yep, you're right :-) Checkout the latest code, i've also taken out the category title and pushed it to the changelog before determining the pull request string.

jetersen:
@rnorth do you mind reviewing it?

jetersen (approved):

jetersen:
@rnorth any concerns or do you think I can go ahead and merge? 😄

If your too busy just let me know.

rnorth:
> @rnorth any concerns or do you think I can go ahead and merge?
> 
> 
> 
> If your too busy just let me know 😅 

Sorry I'm off sick at the moment. I'm sure if you've reviewed then it's going to be fine. Thank you!

jetersen:
@rnorth thanks and get well.

jetersen:
@robbinjanssen thank you for your contribution.

PR #1101, opened 2022-03-15
Interrupt create/update release if no commits available to categorize

blast-hardcheese:
My original fork had an early `return` to interrupt the rest of `drafter`, but I hadn't thought it through when I removed it when cleaning up my upstream contributions.

That early `return` was used to prevent drafting empty releases or empty draft releases. Since I've switched from my branch back to 5.19.0, I'm now getting those empty draft releases again.

My one question is whether

```javascript
      if (runnerIsActions()) {
        setActionOutput(createOrUpdateReleaseResponse, releaseInfo)
      }
```
needs to be run always, or only during create/update. I believe it only makes sense to run during create/update, but I'm open to feedback.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,25 +200,32 @@ module.exports = (app, { getRouter }) => {
     })
 
     let createOrUpdateReleaseResponse
-    if (!draftRelease) {
-      log({ context, message: 'Creating new release' })
-      createOrUpdateReleaseResponse = await createRelease({
+    if (commits.length === 0) {
+      log({
         context,
-        releaseInfo,
-        config,
+        message: 'Not creating a release, no commits to categorize',
       })
     } else {
-      log({ context, message: 'Updating existing release' })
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
Yikes -- I didn't realize there was a legitimate use case for drafting an empty release, could I get some clarification on this before I revisit what I'm trying to do here?

jetersen:
@blast-hardcheese I guess not. I don't see the use case. The test might have been made to simply pass code coverage or something 😅

setActionOutput, you could test whether it is necessary by creating action that uses the output in some way to see how it behaves.

I suspect is it not necessary to set.

jetersen:
@blast-hardcheese I have actually seen cases by jenkinsci and jenkins-infra who expect a `next` or `edge` draft to exist 😅
So they actually expect empty cases.

blast-hardcheese:
Oh no 😧

Alright, that makes sense, all good. Thanks for the follow up!

PR #1202, opened 2022-08-15
Add missing inputs to action config

gilbertsoft:
Providing header or footer leads to a warning and the inputs are ignored
by the action. These two inputs are explicitly mentioned at
https://github.com/release-drafter/release-drafter#action-inputs. This
patch fixes the behavior to finally comply with the docs.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -43,6 +43,16 @@ inputs:
       The object that the release should be created to point to.
     required: false
     default: ''
+  header:
+    description: |
+      A string that would be added before the template body.
+    required: false
+    default: ''
+  footer:
+    description: |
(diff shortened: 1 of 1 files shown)


jetersen (approved):

jetersen:
Sorry, my bad for always forgetting action.yaml

I'll try and include tests for in v6 👏

jetersen:
Released in https://github.com/release-drafter/release-drafter/releases/tag/v5.20.1

gilbertsoft:
Yeah, tests are always good for these kind of stuff :) Thanks for the quick release!

PR #1220, opened 2022-09-07
fetch 100 labels for pull requests instead of 10

matoubidou:
# Issue

* we are using your great action in a setup with quite a lot of labels (one repository with multiple teams using [prow](https://github.com/kubernetes/test-infra/tree/master/prow))
* some PRs are not drafted as only the first 10 labels of each PR are fetched

# Solution

* fetch more labels of the pull requests in question

>Note: I am unsure what might be a good number (our setup normally comes with 10-20 labels). Would 100 be a performance issue?

Thanks in advance!

<sub>Mathias Zeller <mathias.zeller@mercedes-benz.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))</sub>

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -74,7 +74,7 @@ const findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
                   }
                   mergedAt
                   isCrossRepository
-                  labels(first: 10) {
+                  labels(first: 100) {
                     nodes {
                       name
                     }


jetersen:
Instead of 100 maybe do some investigation one of your repos and count the average and max number of labels ?

I think 15 should suffice, however I noticed one PR https://github.com/kubernetes/test-infra/pull/24882 with 76 labels... 😰 If we are at level stopping it from being more.

I don't think we want pagination for labels and since we are only pulling out the name. 100 Maybe be enough. But I would still be interested in some data.

jetersen:
Did some quick powershell and gh api graphql magic:

Got this:
```
Results for kubernetes/test-infra
Average: 6.24016714771895
Maximum: 78
Minimum: 0
Median: 6
Standard Deviation: 3.75587349867424
```

```powershell
$owner = "kubernetes"
$name = "test-infra"

$prsQuery = @"
query(`$endCursor: String) {
  repository(owner: "$owner", name: "$name") {
    pullRequests(first: 100, after: `$endCursor) {
      pageInfo {
        hasNextPage
        endCursor
      }
      nodes {
        labels {
          totalCount
        }
      }
    }
  }
}
"@

# use jq to parse the JSON into a JSON array which is easier to flatten than objects.
$test = Write-Output $prsQuery | gh api graphql --paginate -F query=@- --jq ".data.repository.pullRequests.nodes"
# paginate returns an list of json arrays, so we need to flatten it
$b = @($test | ForEach-Object { ConvertFrom-Json $_ } | ForEach-Object {$_})
$totalCounts = $b.labels.totalCount
$measure = $totalCounts | Measure-Object -Average -Maximum -Minimum -StandardDeviation
$average = $measure.Average
$maximum = $measure.Maximum
$minimum = $measure.Minimum
$standardDeviation = $measure.StandardDeviation

$sortedNumbers = @($totalCounts | Sort-Object)
if ($numberSeries.Count % 2) {
    # Odd, pick the middle
    $median = $sortedNumbers[($sortedNumbers.Count / 2) - 1]
} else {
    # Even, average the middle two
    $median = ($sortedNumbers[($sortedNumbers.Count / 2)] + $sortedNumbers[($sortedNumbers.Count / 2) - 1]) / 2
} 

Write-Output "Results for $owner/$name"
Write-Output "Average: $average"
Write-Output "Maximum: $maximum"
Write-Output "Minimum: $minimum"
Write-Output "Median: $median"
Write-Output "Standard Deviation: $standardDeviation"
```

maigl:
I'm working with @matoubidou, and I want to share some more data.
I don't use powershell :) but some quick linux magic gave me:

```
gh api --paginate -X GET repos/myorg/myrepo/pulls -f q='state:all' | jq '.[] | .labels | length' |jq -s '{minimum:min,maximum:max,average:(add/length),median:(sort|if length%2==1 then.[length/2|floor]else[.[length/2-1,length/2]]|add/2 end)}'
```
```
{
  "minimum": 0,
  "maximum": 12,
  "average": 5.662337662337662,
  "median": 6
}

so yes, we don`t need 100 all the time .. but 15 seems too close to max, imho.

**Comment by jetersen** [comment]
Would you mind pointing your GitHub workflow to your pull request origin repo and branch to verify it can run this query before merging 😅

**Comment by maigl** [comment]
> Would you mind pointing your GitHub workflow to your pull request origin repo and branch to verify it can run this query before merging sweat_smile

We tested it with 100 and it seems to work find for us.

**Review by jetersen — APPROVED** [review]

**Comment by matoubidou** [comment]
To be honest I do not get the problem **Tests / build** action is reporting. I ran

```
yarn prettier && yarn lint --fix && yarn build
```

before commiting. Could it be a problem with different yarn versions? I have been using

```
yarn version v1.22.19
```

I'll try to understand the issue in more depth.

jetersen:
@matoubidou another issue can be CRLF vs LF. Even though git attributes is set to enforce LF.

jetersen:
Released in https://github.com/release-drafter/release-drafter/releases/tag/v5.21.0

Issue #1135, opened 2022-05-05
Release drafter should ignore prerelease releases when drafting changelog


blast-hardcheese:
Somewhat related to #614

I ran into an issue where someone had manually released a draft as `prerelease`, but the subsequently generated draft did not include the commits from the last stable release.

Presumably this is not intentional, but before I start working on this I'd like to make sure.

Thanks!

jetersen:
Do not assume 😅

I don't think anyone had considered the use case.

So the release drafter should ignore pre-releases when drafting a non pre-release. However when drafting a pre-release then the release drafter should include pre-releases ?

Right hopefully we are thinking the same thing ☺️

blast-hardcheese:
> So the release drafter should ignore pre-releases when drafting a non pre-release. However when drafting a pre-release then the release drafter should include pre-releases ?

For the first case, non pre-release drafts should always be since the last non pre-release.

I could go either way on the pre-release (either total accumulating or incremental changesets).

jetersen:
> I could go either way on the pre-release (either total accumulating or incremental changesets).

True, I am leaning more to total cause that is how the non pre-release would potentially look like.

blast-hardcheese:
I could see the other way as well, so it may be desirable for someone to add a configuration parameter if they need something like this, but I think for now it'll be good enough for all releases, prerelease or not, to do the full draft of all notes from the last non pre-release.

I can't jump on this immediately, but thanks for confirming the direction 🙂

hcoohb:
That would be the last feature to allow me to name release-drafter an absolute must and life-saver!!

At the moment I copy and merge all of the pre-release changelog for the next non-pre-release, but it is a bit time consuming.
This would be a terrific improvement!

ssbarnea:
While release drafter is a wonderful tool, that bug alone is an epic PITA that makes me avoid making pre-releases, just because once I do, I need to do a lot of manual work to rebuild the changelog for the release.

Apparently https://github.com/paperless-ngx/release-drafter fork works, so I am wondering if I should switch to the fork, or there is still hope to get the changes backported here?

jetersen:
PR are always welcome 🤗

qcasey:
Our fork at https://github.com/paperless-ngx/release-drafter bluntly ignores prereleases without any configuration to toggle this behavior. I'm not sure how many people's workflows rely on **not** ignoring prereleases so haven't considered bringing it here upstream.

[...]

PR #1240, opened 2022-10-02
Only use last full release when drafting

ssbarnea:
Also disregards pre-releases in addition to drafts when drafting.

Fixes: #1135

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -57,7 +57,7 @@ const findReleases = async ({
     ? commitishFilteredReleases.filter((r) => r.tag_name.startsWith(tagPrefix))
     : commitishFilteredReleases
   const sortedPublishedReleases = sortReleases(
-    filteredReleases.filter((r) => !r.draft)
+    filteredReleases.filter((r) => !r.draft && !r.prerelease)
   )
   const draftRelease = filteredReleases.find((r) => r.draft)
   const lastRelease =


jetersen (approved):

ssbarnea:
On the other hand, I think I got something wrong because I switched to use release-drafter from the for and the changelog was still based on the last pre-release instead of taking only releases under consideration: https://github.com/ansible/ansible-lint/releases

jetersen:
This should be `||` and not `&&`?

jetersen:
@ssbarnea does solves your problem, with my suggested change?

ssbarnea:
@jetersen Testing is right now... i will reply with result.

ssbarnea:
@jetersen Sadly something does not work even with this new setting. I would have expected to add the new merged change on top of the already existing ones, not to start from zero.

lib/releases.js
```diff
@@ -57,7 +57,7 @@ const findReleases = async ({
     ? commitishFilteredReleases.filter((r) => r.tag_name.startsWith(tagPrefix))
     : commitishFilteredReleases
   const sortedPublishedReleases = sortReleases(
-    filteredReleases.filter((r) => !r.draft)
+    filteredReleases.filter((r) => !r.draft || !r.prerelease)
```
blast-hardcheese, line 60: I can't see how this could be anything but `&&`, unless it was `filterNot`

draft | prerelease | candidate
--- | --- | ---
false | false | true
true | false | true
false | true | true
true | true | false

Only in the case that what's being evaluated is a draft for a prerelease does it not factor in. Presumably we want to only consider that which are not draft AND not prerelease.

I am on mobile though, so perhaps missing context, but `&&` seems like it would work, you're saying it didn't?

ssbarnea:
@jetersen I can confirm that with the new changes it works exactly as expected. Maybe we can get it in now?

didix21:
Wondering, how we can use the previous behavior to this change? Is configurable?

ssbarnea:
AFAIK, Is not configurable and I personally do not think it should. A full release is always relative to previous full release. You cannot ask users of reeased version to read prerelease notes.

didix21:
Well, I think it depends on the platform, in our case we use release drafter for generating new iOS releases. The thing is we create a prerelease and we upload it to Apple Store. While we're waiting to the Apple reviewers verdict, we still want to keep developing + when you set the release up as prerelease, the tag is already generated. So the next tag is release-drafter computing is the same because now it can't detect prereleases.

robbinjanssen:
This has broken our release flow as well.. I think we do need to make this configurable.

robbinjanssen:
I've created a pull request where you can config this behaviour. The default functionality will be the same as introduced by this PR.

PR #1255, opened 2022-10-24
fix(release): strip prefix before comparing version

neilime:
In documentation, it says 
> A known prefix used to filter release tags. For matching tags, this prefix is stripped before attempting to parse the version. Default: ""

However, while sorting version to retrieve the latest release, the prefix is taking into account and leads to sort regarding `created_at`, not semver

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -5,11 +5,17 @@ const { getVersionInfo } = require('./versions')
 const { template } = require('./template')
 const { log } = require('./log')
 
-const sortReleases = (releases) => {
+const sortReleases = (releases, tagPrefix) => {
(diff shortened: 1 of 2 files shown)


DuMaM:
Thanks for this PR,
Do you think it can resolve this issue?
https://github.com/release-drafter/release-drafter/issues/1269

neilime:
@DuMaM Yes I think so

neilime:
@jetersen need your approval to run the workflow, and your review. Thanks

neilime:
@jetersen  :up:  thanks

jetersen (approved):

PR #1302, opened 2023-01-20
Add `include-pre-releases` configuration option

### LINKED ISSUE #1291: Pre-Release Changelogs - New Draft Releases has the same stuff like the pre release changelog
Hello!

I got a little problem:
image
image
in these images, the changelogs are pretty normal, because its not a pre release.

However if the Release is a Pre-Release, the Draft Release is copying the changelogs from the pre release in the draft.

Is there a way to "disable" this?

Here are another images to show what i mean:
image
image

### LINKED ISSUE #1293: Introduce flag to consider pre-release version when drafting new release
The changes in #1240, which removed the latest pre-release version from consideration when drafting a new release has introduced problems for my worklow. In my workflow use case, I create a new pre-release which does not get released right away. This is resulting in the new draft release being named the same as the pending pre-release.

Example:
v1.2 (Latest Released Version)
v1.3 (Pre-release pending release)

While v1.3 still has not been released, a new draft release is created, but since v1.2 is the latest release, and v1.3 prerelease is now ignored, a new draft release named v1.3 is created, rather than a version v1.4.

The ask is that a config flag is created which will allow for consideration of pre-releases when resolving a new version, and allows release-drafter to function as it did prior to Release v5.22.

robbinjanssen:
The change introduced in #1240 broke the release flow for some people. By adding a configuration option to `include-pre-releases` we can make this work for everybody :-) The default is `false` so the behaviour introduced in #1240 will be the default behaviour.

tl;dr - If you want to allow pre-releases to count as releases, add `include-pre-releases: true` to your config

Closes #1291
Closes #1293

---
When merging, please squash :-)


robbinjanssen:
@jetersen ready for review :)

lib/releases.js
```diff
@@ -57,7 +58,9 @@ const findReleases = async ({
     ? commitishFilteredReleases.filter((r) => r.tag_name.startsWith(tagPrefix))
     : commitishFilteredReleases
   const sortedPublishedReleases = sortReleases(
-    filteredReleases.filter((r) => !r.draft && !r.prerelease)
+    filteredReleases.filter(
+      (r) => !r.draft && (!excludePreReleases || !r.prerelease)
```
blast-hardcheese, line 62: Logic checks out, but it doesn't read right.
```
!r.draft && (!r.prerelease || includePreReleases)
```
would be preferable.

I'm quite concerned about the amount of boolean logic in this function already which is why I'm even giving this a comment. It probably needs to get reflowed entirely, but that's out of scope here.

blast-hardcheese (commented):
Two maintainability considerations, but lgtm other than that

lib/releases.js
robbinjanssen: I've changed the configuration to `include-pre-releases` which makes more sense :) Thanks for your input!

blast-hardcheese (approved):
Looks great, thank you!

robbinjanssen:
@jetersen sorry to poke you, but can you have a look at this PR and decide whether or not you want this at all?

jetersen (approved):
LGTM

jetersen:
@robbinjanssen Thanks for the PR, sorry for not getting back to you sooner.
@blast-hardcheese Thanks for the review, greatly appreciated 👏

PR #1303, opened 2023-01-22
feat: add prerelease increment behavior

neilime:
Allows to increment the prerelease number (https://semver.org/#spec-item-9).
It does NOT change ANY of the actual behavior
It introduces a new config variable or input "prerelease-identifier"

The prerelease number is increment ONLY when:
- prerelease is true
- prerelease-identifier is not empty


neilime:
Hi @jetersen, any chance to review this PR, when you'll have time?

Thanks

jetersen:
@neilime I think we are missing an update to action.yaml 🤔

neilime:
> @neilime I think we are missing an update to [action.yml](https://github.com/release-drafter/release-drafter/blob/master/action.yml) thinking

@jetersen Which omission are you referring to?

jetersen:
`pre-release-identifier` ? Would be useful to have in action as input option?

instead of only inside `config.yml`

jetersen:
Reason I ask I could forsee people wanting the flexibility to determine the prerelease idenfier based on data and scripts in a GitHub action 🤔

neilime:
> Reason I ask I could foresee people wanting the flexibility to determine the prerelease idenfier based on data and scripts in a GitHub action thinking

Yes sure, I'm on it

wadevries:
@neilime thanks for this change, it almost fixes my use case where I want to maintain separate draft releases for next release candidate and next final release. I've made a PR https://github.com/neilime/release-drafter/pull/1 based on your fork.

jaap3:
I've added the pre-release-identifier input to our fork, https://github.com/leukeleu/release-drafter/commit/e607143f76df11606933158d3e72fc6ef2afa6e3 feel free to cherry pick that commit.

jetersen:
> Minor nit: the naming of `pre-release-identifier` is inconsistent with `prerelease`. Would be nice if `pre-release`/`prerelease` was spelled in a consistent manner.

I concur we should attempt to be consistent here.

neilime:
> > Minor nit: the naming of `pre-release-identifier` is inconsistent with `prerelease`. Would be nice if `pre-release`/`prerelease` was spelled in a consistent manner.
> 
> I concur we should attempt to be consistent here.

FYI I've based the name to be consistent with `include-pre-releases` configuration key

neilime:
@jetersen I've made a separate commit to handle naming of "prerelease-identifier" and input

justinmchase (approved):

jetersen (approved):

PR #1348, opened 2023-07-11
feat: add latest input

o-mago:
Creates a new input `latest` to allow publish or update published releases with the github API parameter `make_latest`:
https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release
Three values are possible:
- `true`
- `false`
- `legacy`*

It will only be applied if the release is not a draft or a prerelease
Nowadays all releases are published as `latest` by default

>*legacy specifies that the latest release should be determined based on the release creation date and higher semantic version

fixes #1348


o-mago:
@jetersen could you review it, please? 😁

jetersen (approved):
LGTM 👍

PR #1365, opened 2023-10-24
feat: add template input

neilime:


neilime:
Hi @jetersen, it's ready for review. Thanks

jetersen:
Since the default is there it would potentially mean it takes precedence over the file 😅 Not sure that is ideal

neilime:
> Since the default is there it would potentially mean it takes precedence over the file 😅 Not sure that is ideal

Oh I see, if I remove the default value, will it be good for you?

PR #1376, opened 2024-01-03
Fix: Correctly mention bot accounts in release notes

### LINKED ISSUE #60: Bot users don’t render quite right
It seems that bot users are inlined as usual, and count as contributors, which results in a bit of unusual rendering:

>* 7: Update dependency emoji-regex to v7.0.1 by @renovate[bot]
>
>Thanks to @renovate-bot, @renovate[bot] and @ticky!

jamietanna:
As noted in #60, bot accounts are currently being mis-mentioned, where
instead of linking through to `dependabot[bot]` it'll tag a user
`@dependabot` which isn't the same, and can result in unwanted
notifications.

To address this, we can retrieve the type of user that has authored a
Pull Request, looking at the `__typename` to determine if it's a `Bot`
or a `User` account, and provide a Markdown link to the Bot account's
app page.

Unfortunately this may result in a name such as:

@[dependabot[bot]](https://github.com/apps/dependabot)

But this is a better result than tagging the wrong account.

Closes #60.

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -111,7 +111,13 @@ const contributorsSentence = ({ commits, pullRequests, config }) => {
       pullRequest.author &&
       !excludeContributors.includes(pullRequest.author.login)
     ) {
-      contributors.add(`@${pullRequest.author.login}`)
+      if (pullRequest.author.__typename === 'Bot') {
+        contributors.add(
+          `[${pullRequest.author.login}[bot]](${pullRequest.author.url})`
+        )
+      } else {
+        contributors.add(`@${pullRequest.author.login}`)
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


justinmchase (approved):

lib/releases.js
```diff
@@ -235,17 +241,26 @@ const generateChangeLog = (mergedPullRequests, config) => {
...
     pullRequests
-      .map((pullRequest) =>
-        template(config['change-template'], {
+      .map((pullRequest) => {
+        var pullAuthor = 'ghost'
+        if (pullRequest.author) {
+          pullAuthor =
+            pullRequest.author.__typename &&
+            pullRequest.author.__typename === 'Bot'
+              ? `[${pullRequest.author.login}[bot]](${pullRequest.author.url})`
```
CFenner, line 250:
```suggestion
              ? `[@${pullRequest.author.login}[bot]](${pullRequest.author.url})`
```

lib/releases.js
```diff
@@ -111,7 +111,13 @@ const contributorsSentence = ({ commits, pullRequests, config }) => {
       pullRequest.author &&
       !excludeContributors.includes(pullRequest.author.login)
     ) {
-      contributors.add(`@${pullRequest.author.login}`)
+      if (pullRequest.author.__typename === 'Bot') {
+        contributors.add(
+          `[${pullRequest.author.login}[bot]](${pullRequest.author.url})`
```
CFenner, line 116:
```suggestion
          `[@${pullRequest.author.login}[bot]](${pullRequest.author.url})`
```

lib/releases.js
```diff
@@ -111,7 +111,13 @@ const contributorsSentence = ({ commits, pullRequests, config }) => {
       pullRequest.author &&
       !excludeContributors.includes(pullRequest.author.login)
     ) {
-      contributors.add(`@${pullRequest.author.login}`)
+      if (pullRequest.author.__typename === 'Bot') {
```
CFenner, line 114: Is the author url set also for human users? Then we could omit the else case.

CFenner (approved):
👍

lib/releases.js
jetersen: URL is always set.

jamietanna:
Thank you 🙇

jetersen (approved):

jetersen:
@jamietanna Thank you for the contribution :clap:

PR #1385, opened 2024-03-26
Update only drafts with the same prerelease status

jaap3:
Do **not** update *final* release drafts with when drafting a *prerelease* and vice versa

Given the following releases:

Name | Draft | Prerelease
-- | -- | --
v3.4.0 |   |  
v3.5.0-rc.4 |   | ✅
v3.5.0 | ✅  |  
v3.5.0-rc.5 | ✅ | ✅

when running release-drafter with this config:

```
prerelease: true
prerelease-identifier: rc
```

I'd expect the draft of `v3.5.0-rc.5` to be updated instead of `v3.5.0`.

This allows me to maintain a draft for the next prerelease, listing the changes since the previous prerelease, and a draft for the next final release, listing the changes since the previous final release.

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -68,7 +68,9 @@ const findReleases = async ({
     ),
     tagPrefix
   )
-  const draftRelease = filteredReleases.find((r) => r.draft)
+  const draftRelease = filteredReleases.find(
+    (r) => r.draft && r.prerelease === includePreReleases
+  )
   const lastRelease = sortedSelectedReleases[sortedSelectedReleases.length - 1]
 
   if (draftRelease) {
--- a/test/releases.test.js
+++ b/test/releases.test.js
@@ -318,6 +318,7 @@ describe('releases', () => {
       const { draftRelease } = await findReleases({
         context,
         targetCommitish: 'refs/heads/master',
+        includePreReleases: false,
         tagPrefix: '',
       })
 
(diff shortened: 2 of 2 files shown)


nimjor:
Any chance we can get this merged?  I would also love to see this functionality, and in fact it is the behavior I had expected when getting started with release-drafter.  I would essentially like to be able to run release-drafter twice for every PR merged.  One manages regular release drafts the "normal" way, while the other immediately publishes a new incremented `-rc` pre-release.  As it is today, the second run overwrites the first.  I can work around this, I think, by using a `tag-prefix` in the configuration for the "prerelease-drafter", but it seems redundant when there is already an expected "tag-suffix" (the prerelease-identifier "rc") to distinguish these tags.

Basically this is a condensed tl;dr of what I'm trying to do:

```yaml
name: Release Drafter
on:
  push:
    branches: [main]
  pull_request:
    types: [opened, reopened, synchronize]
jobs:
  draft-release:
    runs-on: ubuntu-latest
    permissions:
      contents: write
      pull-requests: write
    steps:
      - uses: release-drafter/release-drafter@v6
        with:
          config-name: release-drafter.yaml
          commitish: ${{ github.base_ref || github.ref }}
        env:
  prerelease:
    runs-on: ubuntu-latest
    permissions:
      contents: write
      pull-requests: write
    steps:
      - uses: release-drafter/release-drafter@v6
        with:
          config-name: prerelease-drafter.yaml
          commitish: ${{ github.base_ref || github.ref }}
          prerelease: true
          publish: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
          prerelease-identifier: rc
          header: |
            > [!WARNING]
            > This is a release candidate and as such is not recommended for production use.

        env:
```

jetersen (approved):
LGTM

mvestola:
These changes might have broken fetching existing draft releases. See https://github.com/release-drafter/release-drafter/issues/1425

lib/releases.js
```diff
@@ -68,7 +68,9 @@ const findReleases = async ({
     ),
     tagPrefix
   )
-  const draftRelease = filteredReleases.find((r) => r.draft)
+  const draftRelease = filteredReleases.find(
+    (r) => r.draft && r.prerelease === includePreReleases
```
mvestola, line 72: This evaluates always as false if you have `include-pre-releases: true`  in release-drafter.yml, thus release drafter never finds any draft releases which is not how it should work IMO. Assuming that `includePreReleases` is read from the aforementioned property. See more:  https://github.com/release-drafter/release-drafter/issues/1425

jetersen: Ya, we have variable we can check if semver had prerelease in it, so we should use that.

PR #1390, opened 2024-05-08
Add GHA tag, name, & version to the release body template object

jmuzina:
## What's changed
Adds support for `template` replacement of the following [action input variables](https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs) in `.github/release-drafter.yml`:
| Variable    | Description      |
| ------------- | ------------- |
| `$INPUT_TAG` | `input.tag` from GHA workflow |
| `$INPUT_NAME` | `input.name` from GHA workflow |
| `$INPUT_VERSION` | `input.version` from GHA workflow |

## Reason / Use case
I am using the `input.version` action input to pass in a specific version for the drafted release without using `$NEXT_PATCH_VERSION`, `$NEXT_MINOR_VERSION`, `$NEXT_MAJOR_VERSION`, `$RESOLVED_VERSION` etc. I want to also be able to use this inputted value in the body of the release draft, but noticed it is not supported in any of the template variable sections here, and not passed into `template()` in `generateReleaseInfo()` ([link](https://github.com/release-drafter/release-drafter/blob/77cfd858cefea596b8ac500960670555b15f43bb/lib/releases.js#L365)).


jetersen:
You can use both `$RESOLVED_VERSION` or `$INPUT_VERSION`
`$RESOLVED_VERSION` can be overwritten by `$INPUT_VERSION` but also can also be resolved by the semver increment operation chosen (major, minor, patch)

PR #1400, opened 2024-07-01
feat: default branch

albasyir:
currently, GitHub changed the default branch name from `master` to `main`

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -5,7 +5,7 @@
   "type": "object",
   "properties": {
     "references": {
-      "default": ["master"],
+      "default": ["main"],
       "type": "array",
       "items": {
         "type": "string"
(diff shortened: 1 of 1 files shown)


schema.json
```diff
@@ -5,7 +5,7 @@
   "type": "object",
   "properties": {
     "references": {
-      "default": ["master"],
```
blast-hardcheese, line 8: Probably keep both to avoid breaking all the preexisting repos on GitHub

blast-hardcheese:
Related https://github.com/release-drafter/release-drafter/pull/1079

schema.json
albasyir: dione broo

albasyir:
@blast-hardcheese feedback has been followed

jetersen:
This is generated, so not the correct place to fix it.

Issue #1382, opened 2024-03-09
Missing pull requests in the draft


bonanitech:
Hi guys! This is my first (and hopefully the last 😃) time opening an issue in this fantastic project.

I don't know if I did something wrong, but for some reason, two PRs are not present in the draft of my next release. Any new PRs are automatically included in the draft without issues. Only those two aren't.

The missing PRs are https://github.com/node-red-contrib-themes/theme-collection/pull/368 and https://github.com/node-red-contrib-themes/theme-collection/pull/369.

How can I get that fixed?

Thanks.

<img width="1602" alt="image" src="https://github.com/release-drafter/release-drafter/assets/29807944/a4342b51-b5b4-42e2-800f-8f59c12e263a">

bonanitech:
After some tests in a private repository, I got the impression that the problem was that some labels had a blank space in the name.

https://github.com/node-red-contrib-themes/theme-collection/pull/398

Boy, was I wrong! I forgot to rename the label in that repo, which made me think Release Drafter stopped working because of the spaces in the labels. 🤦🏼‍♂️ So I renamed the label and re-ran the workflow. Everything worked as expected.

Unfortunately, the issue persists in my public repo. I'm wondering if the PRs below broke the Release Drafter for me. Nothing before them appears in the draft.

https://github.com/node-red-contrib-themes/theme-collection/pull/370

bonanitech:
I feel so dumb about my previous comment.

Reading the debug logs, I found this:

`2024-04-17T14:35:58.6831506Z node-red-contrib-themes/theme-collection: Fetching parent commits of refs/heads/dev since 2024-03-01T20:19:50Z`.

That's the date and time of the latest tag creation (https://github.com/node-red-contrib-themes/theme-collection/releases/tag/3.1.11), and the missing pull requests are from an earlier date.

Will I have to add the missing PRs myself to the release notes, or is there something I can change in the settings so they are included automatically?

bonanitech:
I drafted a new release using Github's feature and the missing PRs are listed there.

<img width="1602" alt="image" src="https://github.com/release-drafter/release-drafter/assets/29807944/c46b28df-80de-473b-b70a-38b39f6ca00e">

I don't know what I did to make Release Drafter not list them.

bonanitech:
Sadly, there were no responses to this issue. I removed Release Drafter for now and will write myself the notes for the next release.

Issue #1260, opened 2022-11-05
5.26.X is missing PR's from the release changelog


shadycuz:
Not sure if the change was with `5.26.0` or `5.26.1`, I suspect it was `5.26.1`. 

Instead of showing all the features that was released, it only shows a single bug fix. If you look at my [merged](https://github.com/DontShaveTheYak/cf2tf/pulls?q=is%3Apr+is%3Amerged+sort%3Aupdated-desc+) pull requests. You can see that a number of PR's were merged. Including an update to `release-drafter`.

I did my release exactly like I normally do but the changelog is wrong:

> ![image](https://user-images.githubusercontent.com/17090999/200111948-a141df4b-e30c-45a9-b9a4-c9c9e6bfd026.png)

My [project](https://github.com/DontShaveTheYak/cf2tf) is open source so all the Github actions logs and settings should be available to you.

cchanche:
Hey @shadycuz , has this issue ever occured again recently ? Thanks

jetersen:
This is should fixed in [v7.3.0](https://github.com/release-drafter/release-drafter/releases/tag/v7.3.0) by https://github.com/release-drafter/release-drafter/pull/1604 https://github.com/release-drafter/release-drafter/pull/1570

Issue #871, opened 2021-05-21
Drafts with No Changes


krispharper:
We just installed this and it's working great on some of our repos, but on others, after merging a PR, we get a draft release that just says, "No Changes."

![image](https://user-images.githubusercontent.com/1270239/119197074-350d9080-ba55-11eb-9b7d-95acd2b0356b.png)

Our config in `.github` looks like this

```
change-template: '* $TITLE (#$NUMBER)'
template: |
  $CHANGES
```

Do we have something misconfigured?

krispharper:
Actually just noticed that it appears it's always missing the latest PR in the changes. I.e., if we merge one PR it says "No changes." If we then merge another PR, it will list the first PR, but not the second, etc.

Manuel-Naix:
That is the same behaviour I see in my project. The latest PR merge triggers the action to newly draft the release but the PR merged is not considered and misses in the draft. Only the next merge ensures that the PR appears in the release draft but then again, the most recent misses.

JimNero009:
Same for some of our projects -- not all repos, just some of them, have exactly this kind of 'n-1' behaviour.

krispharper:
Yes, also noticed that it doesn't affect every repo. I'd say maybe half. But I haven't noticed a pattern as to which ones are affected.

c-emil:
I noticed the same issue when I pushed a commit straight to the master branch without a PR. It updated the draft release that contained all the PRs that were merged previously with "No Changes" body.

jetersen:
I have never experienced the n-1 maybe it is because your using the github app?

Have you tried switching to the github action?

c-emil:
@jetersen 
> thats the intended effect. Release drafter is only able to properly categorize pull requests thanks to labels

I think we misunderstood each other. The problem is that it wipes out the body of the draft release completely and adds "No changes" in the body if I push a commit straight to the master branch. What I'd expect is that either it doesn't change anything in the draft release since it only adds changes from the PR commits, or it adds those commits in the "Other changes" (a preferable solution which was also suggested in the issues you referenced). None of this happens.

estahn:
I'm facing a similar issue. Would be great if there is no draft created if there are no changes. I'm using `publish: true` and it started to create empty releases.

As for @c-emil issue, the content of `release-drafter` could be placed in tags similar to `<!-- release-drafter:being !--> ... <!-- release-drafter:end !-->`, which would aid in replacing content partially.

ksolie:
[...]

PR #1408, opened 2024-10-06
feat: Add support for ref comparison

jose-rojas0:
This project is great ❤️ 

I've found myself using it more and more, however I've found that it does not work as I'd expect for some git workflows, specifically when using "merge commits" to the branch where `release-drafter` is configured.

This is happening because `release-drafter` uses the date of the last release to list commits to draft a new release, however if commits were merged with an earlier date they get ignored!

This PR adds a configuration option to support such workflows in a backwards-compatible way.

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -180,7 +180,12 @@ const findCommitsWithAssociatedPullRequests = async ({
       context.octokit.graphql,
       diffCommitsWithAssociatedPullRequestsQuery,
       { ...variables, baseRef: `refs/tags/${lastRelease.tag_name}` },
-      dataPath
+      [
+        'repository',
+        'ref',
+        'compare',
+        'commits',
+      ]
     )
     allCommits = _.get(data, [
       'repository',
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -92,6 +92,68 @@ const findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
   }
 `
 
+const diffCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
+  query diffCommitsWithAssociatedPullRequests(
+    $name: String!
+    $owner: String!
+    $baseRef: String!
+    $targetCommitish: String!
+    $withPullRequestBody: Boolean!
+    $withPullRequestURL: Boolean!
+    $withBaseRefName: Boolean!
+    $withHeadRefName: Boolean!
+  ) {
+    repository(name: $name, owner: $owner) {
+      ref(qualifiedName: $baseRef) {
+        compare(headRef: $targetCommitish) {
+          commits(first: 100) {
+            totalCount
+            pageInfo {
+              hasNextPage
+              endCursor
+            }
+            nodes {
+              id
+              committedDate
+              message
+              author {
+                name
+                user {
+                  login
+                }
+              }
+              associatedPullRequests(first: 5) {
+                nodes {
+                  title
+                  number
+                  url @include(if: $withPullRequestURL)
+                  body @include(if: $withPullRequestBody)
+                  author {
+                    login
+                  }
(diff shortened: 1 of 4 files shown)


lib/commits.js
```diff
@@ -92,6 +92,68 @@ const findCommitsWithAssociatedPullRequestsQuery = /* GraphQL */ `
...
+    $baseRef: String!
+    $targetCommitish: String!
+    $withPullRequestBody: Boolean!
+    $withPullRequestURL: Boolean!
+    $withBaseRefName: Boolean!
+    $withHeadRefName: Boolean!
+  ) {
+    repository(name: $name, owner: $owner) {
+      ref(qualifiedName: $baseRef) {
+        compare(headRef: $targetCommitish) {
```
jarojasm95, line 108: the goal here is to get something like the result of this rest endpoint:
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits

jetersen:
Wouldn't ref comparison not always be the better option over dates? I assume it is related to #1382 #1260 :thinking:

kylejPomelo:
this feature would be useful to my use case. We're managing different types of releases from a monorepo, and whenever one type releases, the commits of the next release, or either type, are calculated since that release. 

I'd like to use this and manually derive the ref on our end and provide that to release-drafter if possible

jetersen:
@jarojasm95 I don't think we need the option to turn this off. I believe would provide a better result then using dates as in we should always use the refs instead of dates. To determine commits that should be included. I would love to see this PR merged.

ScumbagScalawag:
Maybe I can add a little clarity on direction for this feature of "ref comparisons", specifically release notes that can show the diff between two known commits/tags, like `previous_tag_name` in the GitHub API's [generate release notes API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release). 

The [automatically-generated release notes'](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes) `release.yaml` file correctly calculates new Pull Requests for the release because it lists the pull requests between commit `a` and `b` by using something like `git log --graph` (which itself implies the `--topo-order` flag) as opposed to the current implementation, which I suspect is calculating the "new" PRs by date instead of topological order; something like `git log --date-order`. 

The correct solution, which would ***potentially*** also solve issues like #1382 and #1260, would be to make sure the diff to calculate PR's between releases is always determined by topological order of git history. I've tested this logic out in a (private) repo that has Pull requests that were new to a "release" branch, but older than the last release + release notes and using the topological history (or `--graph`) is the only way for me to validate the auto-generated release notes' listed PRs. When I do regular git log and look at the commit range it fails to include a series of PRs that I know are actually new to that branch/build/release notes. 

Not sure if this is *exactly* what's going on, but it's one of the problems I've found with calculating PR diffs for the sake of another complex release train I'm managing. I can elaborate more if needed, but I have a suspicion that the topological ordering is necessary not just to feature match GitHub's `previous_tag_name`, but also to fix these issues of missing PRs

cchanche:
Sorry for the delay, this is an interesting PR and debate

Using the last release date is indeed maybe not the best, and I concur with the idea taht it should be replaced with soemthing more accurate (for reasons depicted above).

In https://github.com/release-drafter/release-drafter/issues/871#issuecomment-3944566391 we also discussed that the github API `associatedPullRequest` is not super reliable, and that having a pure `git` CLI implementation of retrieving PRs in the git history would be nice.

I would love if @ScumbagScalawag you could elaborate on how you achieved PR retrieval with [`git` and the `--topo-order`](https://git-scm.com/docs/git-rev-list#_commit_ordering) flag ! Things get confusing pretty fast in there...

cchanche:
implemented in #1570

PR #1409, opened 2024-10-06
feat: Add support for commits without pull requests

jose-rojas0:


jetersen:
I feel like if we want to add this we should at least offer a way of adding label support via commit message's like `feat(desktop):` or `bugfix:` :thinking:

alies-dev:
Great PR!

>I feel like if we want to add this we should at least offer a way of adding label support via commit message's like feat(desktop): or bugfix: 🤔

I think it can be as one of the next steps to evolve this idea. The core functionality is already there

jetersen:
Closing, we have moved the codebase to Typescript. Feel free to recreate it on the newer codebase. Also work is happening in #1466

Issue #469, opened 2020-03-31
Monorepo support


josemarluedke:
First, thank you for this project. I really like it and I have been using it in every project that I work on.

This is a feature request that probably is a bit controversial, but I wanted to submit anyway.

When working in projects that are monorepo, the changelog usually makes sense to mention what packages the changes have affected. This is very common in JavaScript projects for example.

There is [lerna-changelog](https://github.com/lerna/lerna-changelog) that has support for this, for reference.

It would be awesome if we could have something like this with release-drafter. I understand that this project is not specific to a language and that's why I think it is a bit controversial. However, many other languages have the same concept of packages and monorepos.

Thank you.

gavinlove:
You can use [actions/labeler](https://github.com/actions/labeler) to automatically add a label with the correct package then add those labels to the categories list in the release drafter config.

mxj4:
monorepo almost works if I run this action multiple times, each time with a different configuration, the only issue is when detecting last release: the last release from a different package is also included in search. If we can have a git tag prefix option in the configuration, then we can do better filtering when detecting the last release and exclude other packages.

jetersen:
I believe @blast-hardcheese managed to get it working pretty well.

https://github.com/guardrail-dev/guardrail/blob/master/.github/workflows/release-drafter.yml it ant pretty but it works 👍

blast-hardcheese:
That reminds me, I still need to apply your advice for config merging, instead of generating via bash script, it would be a better example for those trying to do this

Blacksmoke16:
@jetersen Thanks for that link! I played around a bit with that approach, but not sure it'll fit my use case unfortunately.

In my case I don't have an actual monorepo so I think there would be issues with the drafts including _every_ PR all the time given there would be no release to filter by. Unless you can filter by commit hash? But even then would be kinda a meh flow needing to update that whenever a version is released in a child repo.

After thinking about it for a bit, I think the crux of my issue is wanting to create the release in a different repo than which the PR is merged while still filtering to specific paths for each child repo. I.e. this way the child repo's latest version/releases list is used and is only updated with PRs that affected it.

Totally understandable if this isn't something that is going to be supported. Just going through and trying to figure out my options to make it so I don't need to keep track of all this manually anymore :sweat_smile:.

blast-hardcheese:
[...]

PR #1423, opened 2025-01-19
feat: Add Support for Injecting Configuration Directly into Release Drafter

ReneWerner87:
This pull request introduces a new feature to the Release Drafter GitHub Action that allows users to inject configuration dynamically during workflow execution.

# Key Changes

1. **New `config` Input**:
   - Users can now provide the release configuration directly as a string using the `config` input.
   - This eliminates the need for configuration files to be stored in the repository or committed during workflow execution.
   - The `config` input takes precedence over the `config-name` input if both are provided.

2. **Use Case**:
   - Ideal for workflows where configurations are dynamically generated or vary depending on runtime conditions.
   - Avoids unnecessary commits of temporary or environment-specific configuration files. e.g. (https://github.com/gofiber/storage/commit/2aa6a93b98c53e37c446c72f8eb9ef06d6b6b8e4)

3. **Example**:
   - Added documentation with a clear example of how to use the new `config` input to inject configurations directly.

# Benefits

- Enhances flexibility in using the Release Drafter Action.
- Simplifies workflows by removing the need for file-based configurations in some cases.


ReneWerner87:
tested the concept
https://github.com/gofiber/storage/actions/runs/12856319049/job/35842867376

ReneWerner87:
friendly ping ?

gaby:
@jetersen @toolmantim @TimonVS Any chance of getting a review for this PR?

It heavily simplifies the usage of release-drafter for mono-repos.

gaby:
Hello @jetersen @toolmantim @TimonVS Any chance of getting a review for this PR? We have been trying to reach out for 2 months with no answer.

Thanks!

say8425:
This would make it easier to support monorepo. Currently, I [have to create n number of config files](https://github.com/release-drafter/release-drafter/issues/469#issuecomment-1122499757) like the code below to use the monorepo feature, which is inconvenient. 
https://github.com/guardrail-dev/guardrail/blob/master/.github/workflows/release-drafter.yml

PR doesn't seem to have any issues, is there anything in particular that needs to be reviewed further?

ReneWerner87:
@jetersen @toolmantim @TimonVS ping

mdelapenya:
@jetersen friendly ping! I knew about this and I'm interested in adopting this pattern in the testcontainers-go monorepo 🙏 

Please let me know if you need any kind of support with it

jrbeilke:
@und-miller what do you think about this feature/PR?

und-miller:
> @und-miller what do you think about this feature/PR?

I would want feedback from @jetersen before merging new features as I am not as familiar with the repo.

gaby:
@jrbeilke @und-miller @jetersen We been running this for months in the GoFiber org using a fork. Having this merged would be awesome

action.yml
```diff
@@ -13,10 +13,10 @@ inputs:
       The config should still be located inside `.github` as that's where we are looking for config files.
     required: false
     default: 'release-drafter.yml'
-  name:
+  config:
```
czuares, line 16: @ReneWerner87 Is it necessary for the `name` input to be removed? It looks like both `name` and `config` are supported and this change only exists in the non-docker version.

```suggestion
  name:
    description: |
      The name that will be used in the GitHub release that's created or updated.
      This will override any `name-template` specified in your `release-drafter.yml` if defined.
    required: false
  config:
```

ReneWerner87: @czuares  not sure
image

What you mean, in the full change you can only see that the config property has been changed.

image

I'm not sure why the range doesn't include all the changes.

czuares: Unless I am missing something obvious, I don't see `name` in the list of inputs for the `action.yml` since it appears it was replaced with `config`.

The [docker/action.yml](https://github.com/ReneWerner87/release-drafter/blob/6dec4ceb1fb86b6514f11a2e7a39e1dedce709d0/docker/action.yml#L16-L24) contains both `name` and `config` inputs.

I also see a warning in my editor when referencing `name` using the HEAD commit as well.

Image

gaby:
@cchanche Thoughts on getting this merged?

cchanche:
Hey thank you for pointing this out ! Been quite busy on other stuff lately

On the surface, I think this is great. I will try to take a look asap and let you know.

cchanche:
So introducing such a feature indeed opens quite a bunch of possibilities, and I would love for us to nail this as simply and elegantly as possible. 

Until now, the community has been issuing PRs for individual inputs they wanted to dynamically configure, and this has led to having a subset of params in the action's input - somewhat not very cleanly implemented. This PR superseds the need for all of these - but we need to continue to support them for this to be non-breaking.

I would love to take advantage of the incentive, and clean-up how we do things. Here is a breakdown of things that bug me : 

- current action inputs are not being validated like the config file is (using Joi)
  - i get that `config:` content is being validated in the PR, but I mean to validate the rest
- the config file(s) is (are) currently being fetched by probot using `context.config`, which uses octokit's [`rest.repos.getContent`](https://octokit.github.io/rest.js/v18/#repos). That means we are looking for a file on the default branch, **and not on the current file system at runtime**.
- I would like this `config:` to be **merged** with the config file, instead of plainly replacing
- merging inputs and the config file currently requires _not very clean_ sets of _ifs_ and _elses_
  - Joi should be able to help with that. I know zod can. Migrating to zod is an option

If this is to be merged, I would also like for us to asses exactly what it means for users to specify inputs like this :

```yaml
- name: Use Release Drafter with Injected Config
  uses: release-drafter/release-drafter@v6
  with:
    config: |
      latest: true
      commitish: ${{ context.ref_name }}
      version: v${{ steps.version.outputs.major }}
```

... instead of the current :

```yaml
- name: Use Release Drafter with Injected Config
  uses: release-drafter/release-drafter@v6
  with:
    latest: true
    commitish: ${{ context.ref_name }}
    version: v${{ steps.version.outputs.major }}
```

Questions I have right off about this new `config: |` syntax are : 

- how does it impact user's ability to use variable expansion ?
- how does it impact user's ability to use github action expressions ?
- what is the merging order/priority ? We have :
  - the config string as an input
  - the config file
  - individual configs as inputs (if still relevant)

jetersen:
This `config: |` is bad and it is not strongly typed and no schema parser can parse it except at runtime so it becomes hard as the one managing the action workflow it becomes less ideal.
```yaml
config: |
```

I am in favor of `with:` action input supporting all our config flags but some features are not supported like nested objects so in fact it should only be limited options and rest have to come from a config file. We could change how we fetch the file. Maybe we can simply adopt a pattern of check if the file exist locally and if not assume we should use the the [octokit config context](https://github.com/probot/octokit-plugin-config)

The other thing we could consider is those that have nested objects we could consider using the string multiline yaml but again going down a dark road at least that's how it feels like to me.
```yaml
with:
  categories: |
    - title: '🚀 Features'
      labels:
        - 'feature'
        - 'enhancement'
    - title: '🐛 Bug Fixes'
      labels:
        - 'fix'
        - 'bugfix'
        - 'bug'
    - title: '🧰 Maintenance'
      label: 'chore'
```
But even just writing this was cursed, cause indents :cry:

cchanche:
I guess if ppl are able to dynamically create such strings, they might aswell write it down into a file. Us supporting fetching the file from the runner is maybe the quickest fix here.

I did not think about those nested fields, and writing them down manually does not look like fun indeed 😅 

And then also about validating the current few options params that are available through inputs, maybe that can wait #1476

cchanche:
OP's github workflow with current PR looks like this : 

```yaml
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6
      - name: Generate dynamic config from template
        id: generate-config
        run: |
          folder="${{ matrix.package }}"
          sed "s|{{FOLDER}}|$folder|g" .github/release-drafter-template.yml > .github/release-drafter-$folder.yml
          echo "config<<EOF" >> $GITHUB_OUTPUT
          cat .github/release-drafter-$folder.yml >> $GITHUB_OUTPUT
          echo "EOF" >> $GITHUB_OUTPUT
      - name: Use dynamic release-drafter configuration
        uses: ReneWerner87/release-drafter@6dec4ceb1fb86b6514f11a2e7a39e1dedce709d0
        with:
          config: ${{ steps.generate-config.outputs.config }}
```


I think step `Generate dynamic config from template` could easily be rewritten to redirect to a file instead of `$GITHUB_OUTPUT` (given adding support for fetching files locally) : 

```yaml
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6
      - name: Generate dynamic config from template
        run: sed "s|{{FOLDER}}|$folder|g" .github/release-drafter-template.yml > .github/release-drafter-${{ matrix.package }}.yml
      - name: Use dynamic release-drafter configuration
        uses: ReneWerner87/release-drafter@6dec4ceb1fb86b6514f11a2e7a39e1dedce709d0
        with:
          config-name: .github/release-drafter-${{ matrix.package }}.yml
```

Would that sound good to you @gaby @ReneWerner87 ?

ReneWerner87:
Yes, that sounds okay.
The only thing that's important to me is that I can create it dynamically and inject it.
It doesn't matter whether I go through a temporarily created file or inject it directly.

Normally, that doesn't make any difference from my point of view.
Or is it difficult to treat dynamic content as YAML content?

jetersen:
Ya quick fix is properly allow checking if the config file is available locally and use that. Again if it's easy to generate and/or use tools like yq/sed to patch it with tools available in github runners. I feel like adding `config` input might be viable if we rely on other github actions steps to generate it but our github action becomes more complex action to support config file, `config` input and all other actions input :sweat:

cchanche:
This will be part of [Next major version](https://github.com/release-drafter/release-drafter/pull/1475). Please read [this discussion](https://github.com/release-drafter/release-drafter/pull/1475#discussion_r2737793512) for details.

cchanche:
superseded by #1475

PR #1429, opened 2025-02-06
Fix: Move changelog empty check after categorization for accuracy

Bledai:
Previously, the changelog check was based on `mergedPullRequests.length`, which could be misleading since some PRs might be filtered out or uncategorized. This change ensures the check happens after categorization, making sure only meaningful PRs are considered.

- Moved the "no changes" check after PR categorization
- Improved accuracy by counting only categorized and uncategorized PRs
- Optimized changelog generation by preventing unnecessary processing

Fixes potential issues where a release might be marked as having changes when none remain after filtering.

### DIFF
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -219,12 +219,20 @@ const categorizePullRequests = (pullRequests, config) => {
 }
 
 const generateChangeLog = (mergedPullRequests, config) => {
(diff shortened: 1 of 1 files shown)


cchanche:
Very nice catch ! Thank you @Bledai, sorry for the delay

PR #1442, opened 2025-05-23
feat: add support for updating tagged releases

borislavr:


cchanche (commented):
Bu using the `tag` input, you update and/or create a release that is associated with this tag.

The update is performed regardless of whether the release is already published, or a draft. (from what I understand in your code)

This does not look to be in line with what we want to achieve, neither with the tag input nor release-drafter as a whole.

Please clarify what your exact use-case is @borislavr 🙏

borislavr:
Hi @cchanche 
**My use case was as follows:** For publishing and releasing Helm charts, we used [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action), which automatically creates a release in GitHub but does not generate release notes. Since we use **release-drafter** everywhere to generate release notes according to a unified template, we need to recreate the release notes for releases created by chart-releaser-action.

**P.S.:** Later we abandoned chart-releaser-action and implemented our own Charts release mechanism, so the additional functionality from this PR is no longer needed actually.

cchanche:
Ah I see, that's interesting ! Thank you for sharing.

But if it's not needed anymore, I'd like to close this (if you don't mind) - branch would need to be rebased anyway.

If anyone ever has a similar request, I think the corresponding config should be more explicit about about release-drafter becoming a plain release-editor instead.

As a workaround, I think there may also be a way to achieve this using rease-drafter's output in dry-run mode, and then using another action (or plain API calls) to edit past published releases.

PR #1444, opened 2025-06-04
feat: allow always collapsing a category

mhanberg:
This patch allows setting the `collapse-after` key to 0 in order to
always collapse a category. This is useful in cases where you have
entries in multiple categories, and you want to by default hide the
"duplicate" categories.


jetersen:
@mhanberg seems like a good feature to have, maybe your willing to update it after typescript conversion?

mhanberg:
Can do

mhanberg:
@jetersen should be good for review now.

cchanche (approved):

PR #1463, opened 2025-11-25
Implements a personalization of the pagination limit in the GraphQL API.

### LINKED ISSUE #1419: GraphQL API timeout/error when fetching parent commits with a large history
When querying parent commits for a branch with a large history (1500+ commits between releases), the GraphQL API consistently returned an error and failed to execute the query. This issue persisted for several days until a manual release was generated. It appears that the large number of commits between releases triggered a timeout or internal error.

image

image

In this github doc  https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28 is mentioned that the api must respond in 10 seconds before it throws a timeout, so I was hopping that there was some configuration to avoid the timeout while using the action

NunesErick:
This PR implements customizable pagination for GitHub's GraphQL API to address issues caused by commits with large histories.

Fix #1419

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -73,6 +73,11 @@ inputs:
       A boolean indicating whether the autolabeler mode is disabled.
     required: false
     default: ''
+  pagination-limit:
+    description: |
+      The maximum number of items to fetch per request when listing pull requests or other paginated data.
+    required: false
+    default: '100'
 outputs:
   id:
     description: The ID of the release that was created or updated.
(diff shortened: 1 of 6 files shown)


cchanche:
Merged in #1470, available in [v6.2.0](https://github.com/release-drafter/release-drafter/releases/tag/v6.2.0)

PR #1379, opened 2024-01-29
Update Node.js to 20

massongit:
fix https://github.com/release-drafter/release-drafter/issues/1378

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

Node.js 16 actions are deprecated.
Therefore, I update Node.js used in this Action to 20.

PR #1465, opened 2025-12-03
chore: bump node to v24

### LINKED ISSUE #1460: Node20 will reach end-of-life (EOL) in April of 2026
> Node20 will reach end-of-life (EOL) in April of 2026. As a result we have started the deprecation process of Node20 for GitHub Actions. We plan to migrate all actions to run on Node24 in the fall of 2025.

See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

>> Node24 is incompatible with macOS 13.4 and lower versions.
Node 24 does not have official support for ARM32, so self-hosted runners on ARM32 will no longer be supported after Node 20 deprecation.
To find out more about the OS versions we support and self-hosted runner architectures, please read our documentation.

https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runsusing-for-javascript-actions

cchanche:
Fixes #1460 

Did not quite take the time to understand why Node 22 was not suggested by Github in their [depreciation notice](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) @v1v provided in the original issue, but I figured node 24 might also do the trick just fine.

Here is Node's release calendar (for reference): 

image

/cc @und-miller @oleg-nenashev

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,8 @@
         "prettier": "2.7.1"
       },
       "engines": {
-        "node": ">=20.0.0"
+        "node": ">=24.0.0",
+        "npm": ">=11.5.2"
       }
     },
     "node_modules/@actions/core": {
@@ -113,6 +114,7 @@
       "resolved": "https://registry.yarnpkg.com/@babel/core/-/core-7.26.0.tgz",
       "integrity": "sha1-14tgI8yPMRTM8EnrIZYT90p0e0A= sha512-i1SLeK+DzNnQ3LL/CswPCa/E5u4lh1k6IAEphON8F+cXt0t9euTshDru0q7/IqMa1PMPz5RnHuHscF8/ZJsStg==",
       "dev": true,
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.0",
@@ -1227,6 +1229,7 @@
       "version": "3.6.0",
       "resolved": "https://registry.yarnpkg.com/@octokit/core/-/core-3.6.0.tgz",
       "integrity": "sha1-M3bLnzAI2bPREDcNkOCh/NX+YIU= sha512-7RKRKuA4xTjMhY+eG3jthb3hlZCsOwg3rztWh75Xc+ShDWOfDDATWbeZpAHBNRpm4Tv9WgBMOy1zEJYXG6NJ7Q==",
+      "peer": true,
       "dependencies": {
         "@octokit/auth-token": "^2.4.4",
         "@octokit/graphql": "^4.5.8",
@@ -2151,6 +2154,7 @@
       "resolved": "https://registry.yarnpkg.com/acorn/-/acorn-8.14.0.tgz",
       "integrity": "sha1-Bj4scMrF+09kZ/CxEVLgTGgnlbA= sha512-cl669nCJTZBsL97OF4kUQm5g5hC2uihk0NxY3WENAC0TYdILVkAyHymAntgxGkl7K+t0cXIrH5siy5S4XkFycA==",
       "dev": true,
+      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.209.6/containers/javascript-node/.devcontainer/base.Dockerfile
 
 # [Choice] Node.js version (use -bullseye variants on local arm64/Apple Silicon): 16, 14, 12, 16-bullseye, 14-bullseye, 12-bullseye, 16-buster, 14-buster, 12-buster
-ARG VARIANT="16-bullseye"
+ARG VARIANT="24-bullseye"
```
v1v, line 4: wonder if this should be changed to another debian version, for instance debian 13 named `trixie`:

https://www.debian.org/releases/

image

.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
```
v1v: wonder if this version could be bumped to:

```suggestion
        uses: actions/setup-node@v6
```

.github/workflows/action-build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-node@v3
```
v1v: wonder if this version could be bumped to:

```suggestion
        uses: actions/setup-node@v6
```

.github/workflows/tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
```
v1v: wonder if this version could be bumped to:

```suggestion
        uses: actions/setup-node@v6
```

v1v (commented):
LGTM, just left a couple of comments

.devcontainer/Dockerfile
cchanche: You are absolutely right !

I made a quick tour of the VS Code specific config files (including `.devcontainer/`) and did a quick cleanup of some clutter and deprecated params.

Take a look at https://github.com/release-drafter/release-drafter/pull/1467

.github/workflows/release.yml
cchanche: Same here opened a dedicated pr : https://github.com/release-drafter/release-drafter/pull/1468

.github/workflows/action-build.yml
cchanche: Same here opened a dedicated pr : https://github.com/release-drafter/release-drafter/pull/1468

.github/workflows/tests.yml
cchanche: Same here opened a dedicated pr : https://github.com/release-drafter/release-drafter/pull/1468

cchanche:
So basically, the version of node our action runs on is determined by :

```yaml
# action.yml
runs:
  using: 'node20' # changed to node24
```

... but that does not determine what node version we are building it with - meaning the change above should be enough.

However, I think not aligning both of these is a mistake, and some work has already been done here about this.

In #1480 , we briefly discussed how :
1. using node v24 could allow us to have a npm version compatible with oidc out of the box (>= 11.5.2)
2. pushing `using: 'node24'` should or should'nt be a breaking change for our users.


Item 1. seems easy enough : simply merge back master here, remove `npm install -g npm@11.5.2`, and look at what `actions/setup-node@v6` with `node-version: 24` provides out of the box (npm version)

Item 2. is less easily sorted-out. My take is that we should make this a breaking change. In #1379 (update do node 20), @jetersen pointed-out that some GH Enterprise workflows may break if this had not been a breaking change. I think this has to do with the fact that some github users (mostly GHE users I guess) use custom github-actions runners. The public runner introduced it's support for node 24 mid 2025 in [v2.327.1](https://github.com/actions/runner/releases/tag/v2.327.1). It was immediately available for users using the public runners, but some GHE users may not have updated their custom runners (or whatever they are doing, maybe they use the public one, only using a fixed tag they would not have bumped yet)

cchanche:
I think this is ready. I would like to make it a breaking change. LMK @jetersen

cchanche:
Superseded by #1475.

PR #1467, opened 2025-12-07
chore: upgrade various vscode specific settings

cchanche:
quick cleanup of some clutter and deprecated params


jetersen:
Seems outdated?

Issue #1469, opened 2025-12-09
Better error handling - action step does not fail despite error


cchanche:
Make sure action fails whenever an error is thrown.

The exact locations errors are thrown (and do not fail the action) is yet to be determined. There are multiple, so I figured there might be a single `catch` to discard somewhere.

Examples are: 
 
- #1450 
  - `Error: Validation Failed`
- #1335 
  - another `Error: Validation Failed`
- #1313 
  - `Error: Request failed due to following response errors`
- #1301 
  - `tag is already exist`
- https://github.com/release-drafter/release-drafter/issues/1450#issuecomment-3210582238
  - `502 Bad Gateway`
- #1413
  - insufficient token permission
- #1350
  - insufficient token permission

cchanche:
fixed in #1475

Issue #1450, opened 2025-07-18
GitHub action status should be marked as failed in case of errors


slawekjaranowski:
Example we have in logs:

```
Error: Validation Failed: {"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"}
```

release draft is not created by action but status is success

slawekjaranowski:
similar as:
 - #1335

slawekjaranowski:
next one:
 - #1313
 - #1301

tdekoning93:
Same problem;

Release Drafter fails to create draft release, logs 502 Bad Gateway, but the workflow step returns with exit code 0 and does not fail the job.

DemesneGH:
> Same problem;
> 
> Release Drafter fails to create draft release, logs 502 Bad Gateway, but the workflow step returns with exit code 0 and does not fail the job.

Same 502 Bad Gateway:
```
{
  name: 'HttpError',
  id: '17127680285',
  status: 502,
  response: {
    url: 'https://api.github.com/graphql',
    status: 502,
    headers: {
      'content-length': '150',
      'content-type': 'text/html',
      date: 'Thu, 21 Aug 2025 13:13:52 GMT',
      server: 'github.com',
      vary: 'Accept-Encoding, Accept, X-Requested-With',
      'x-github-request-id': ...
    },
    data: '<html>\r\n' +
      '<head><title>502 Bad Gateway</title></head>\r\n' +
      '<body>\r\n' +
      '<center><h1>502 Bad Gateway</h1></center>\r\n' +
      '<hr><center>nginx</center>\r\n' +
      '</body>\r\n' +
      '</html>\r\n'
  },
...
```

Seems the similar issue: https://github.com/googleapis/release-please/issues/2577

Any fix for this? Thanks!

radirobi97:
Hey @jetersen @jaap3 sorry for tagging you directly,  but any plans to fix this? We can see that the project is lacking of active contributors at the moment.

dx777:
So basically release drafter is unusable as of right now ? 2 days trying to implement it and always get 502 error

amitkma:
Any workaround to fix it?

gjvoosten:
[...]

Issue #1335, opened 2023-04-10
Validation errors in action and no draft created but action reports success


jdforsythe:
I'm seeing odd behavior from configurations that have been working for many months across over a hundred repositories.

We have seen duplicate draft releases created, one by the "release-drafter" user and another by a "github-actions" user, both for the same version number. I have been deleting one of them and publishing the other, and then when we merge more PRs, there are duplicates again.

Then we saw a release *deleted after it was published*. There was a discrepancy in the version number when we looked at the releases. The latest release was v1.1.1 but we knew we had published a v1.1.2 (the tag exists and it was published to NPM, which can only happen when we publish a release). The newest draft release from release-drafter was showing as v1.1.2, so I simply updated the tag number and name of the release to reflect v1.1.3 and published it. The workflow ran again when the draft was published, which I don't understand - it's only configured to run on push to master and the 3 pull request types, opened, reopened, and synchronize - and it overwrote the *published* release with a new draft release.

This explains the issue with the missing release, but I don't understand why this action would be running when the draft gets updated from draft to published - the config shows it should not:

```yml
name: Release Drafter

on:
  push:
    branches:
      - master

  ## required for auto labeler
  pull_request:
    types: [opened, reopened, synchronize]

jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
      - uses: release-drafter/release-drafter@v5
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Digging into what's happening, I triggered the action to run with a new PR and noted several errors in the logs:

```
Run release-drafter/release-drafter@v5
"pull_request_target.opened" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)
"pull_request_target.reopened" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)
"pull_request_target.synchronize" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)
"pull_request_target.edited" is not a known webhook name (https://developer.github.com/v3/activity/events/types/)
```

```
***ORG_NAME***/***REPO***: Found 8 releases
{ name: 'event', id: '4660822018' }
***ORG_NAME***/***REPO***: No draft release found
{ name: 'event', id: '4660822018' }
***ORG_NAME***/***REPO***: Last release: v1.1.3
{ name: 'event', id: '4660822018' }
[...]

Issue #1313, opened 2023-02-15
Step does not fail on query failure


ozydingo:
Using `release-drafter/release-drafter@v5`

We encountered a failure in one of the internal steps of Draft Releaser. This failure caused our outputs to be empty, which breaks downstream steps. I expect this failure would cause the job to fail, but it did not.

To be clear, I am primarily concerned here that the job did not fail, and only secondarily concern with what the actual error was.

Log excerpt:

```
3playmedia-dev/threeplaymedia_app3: Found 42 releases
{ name: 'event', id: '4185073928' }
3playmedia-dev/threeplaymedia_app3: Draft release: 0.41.0
{ name: 'event', id: '4185073928' }
3playmedia-dev/threeplaymedia_app3: Last release: 0.40.0
{ name: 'event', id: '4185073928' }
3playmedia-dev/threeplaymedia_app3: Fetching parent commits of f76250b617f26cf72f50bb4af666f5a1fd93b060 since 2023-02-[14](https://github.com/3playmedia-dev/threeplaymedia_app3/actions/runs/4185073928/jobs/7252147220#step:2:15)T16:42:47Z
{ name: 'event', id: '4185073928' }
Error: Request failed due to following response errors:
 - Something went wrong while executing your query. Please include `5403:99CD:1CB71FC:3AB44D1:63ECF4EE` when reporting this issue.

...

Error: GraphqlResponseError: Request failed due to following response errors:
     - Something went wrong while executing your query. Please include `5403:99CD:1CB71FC:3AB44D1:63ECF4EE` when reporting this issue.
        at /home/runner/work/_actions/release-drafter/release-drafter/v5/dist/index.js:6088:13
        at async paginate (/home/runner/work/_actions/release-drafter/release-drafter/v5/dist/index.js:142845:14)
        at async findCommitsWithAssociatedPullRequests (/home/runner/work/_actions/release-drafter/release-drafter/v5/dist/index.js:142646:12)
        at async drafter (/home/runner/work/_actions/release-drafter/release-drafter/v5/dist/index.js:142388:7)
        at async Promise.all (index 0)

...

  stack: 'AggregateError: \n' +
    '    GraphqlResponseError: Request failed due to following response errors:\n' +
    '     - Something went wrong while executing your query. Please include `5403:99CD:1CB71FC:3AB44D1:63ECF4EE` when reporting this issue.\n' +
[...]

PR #1475, opened 2026-01-12
feat: new major version

### LINKED ISSUE #335: Deprecating the Release Drafter GitHub App (not Actions!)
Hey all, we're planning to deprecate the Release Drafter GitHub App in favor of the [Release Drafter GitHub Action](https://github.com/marketplace/actions/release-drafter). This move will help us focus on making improvements to the Action and reduce time spent on maintaining two platforms.

The plan is to keep the GitHub App running for at least another year. During this period we'll keep updating it with security patches and bug fixes (only for breaking bugs), but it won't receive any new features.

Please let us know if you have any concerns about deprecating the GitHub App.

### LINKED ISSUE #336: Option to generate template/config from a script
It looks like `release-drafter.yml` is currently loaded using probot-config, but it would be great if it could also be loaded from a local checkout from within actions. Then we could generate custom templates using shell scripts or other actions to feed into Release Drafter, and don't necessarily need to implement a huge number of dynamic features in Release Drafter itself. As it is, the config file must be static and exist on the main branch.

### LINKED ISSUE #1014: use config.get directly instead of probot's getConfig call
Could be improved even more by using: https://github.com/probot/octokit-plugin-config/blob/a613c0efd2873fb64a3a87f79365ec1951fd29ad/src/index.ts#L13 directly, as we would be able to specify a branch to get the config from. Actually this function is even better as it will attempt to use `.github` repo without using `_extends`

_Originally posted by @jetersen in https://github.com/release-drafter/release-drafter/issues/365#issuecomment-1013868496_

### LINKED ISSUE #1274: Add GitHub token to input via `${{github.token}}`
GitHub Actions can get a token via `github` context like `actions/checkout`. 
https://github.com/actions/checkout/blob/bf085276cecdb0cc76fbbe0687a5a0e786646936/action.yml
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Users can get rid of `${{secrets.GITHUB_TOKENS}}`.

### LINKED ISSUE #1334: Error: Configuration file .github/release-drafter.yml is not found. The configuration file must reside in your default branch.
I am struggling to get the `release-drafter` action to work for a private repo within a GitHub Enterprise account. I am running this action as part of a custom GHA workflow. I am not sure where to go from here. I have also tested removing the `Set GHE_HOST` job step but the action fails with an `Error: Validation Failed: {"resource":"Release","code":"invalid","field":"target_commitish"}` error and a `422` HTTP status code when calling the `https://api.github.com/repos/[MY_ORG]/[MY_REPO]/releases` endpoint.

**Steps I have taken so far:**
* Added a `.github/release-drafter.yml` file to the repository and merged the file into the `main` (default) branch.
```
name-template: 'Test - v$RESOLVED_VERSION'
tag-template: 'v$RESOLVED_VERSION-test'
categories:
  - title: '🚀 Major/Breaking Changes'
    labels:
      - '(MAJOR)'
  - title: '🐛 Minor Changes/Bug Fixes'
    labels:
      - '(MINOR)'
  - title: '🧰 Maintenance Patches'
    label: '(PATCH)'
change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
template: |
  ## Changes

  $CHANGES

```

* Created a new `test` branch and added the following steps to my GHA CI workflow.
```
      # Create/update a GitHub Release
      # Note: Setting the `GHE_HOST` variable is required for GitHub Enterprise accounts.
      - name: Set GHE_HOST
        run: |
          echo "GITHUB_SERVER_URL = '${GITHUB_SERVER_URL}'"
          echo "GHE_HOST=${GITHUB_SERVER_URL##https:\/\/}" >> $GITHUB_ENV

      - name: Update GitHub Release
        with:
          config-name: release-drafter.yml
          name: '${{ steps.create-tag-suffix.outputs.tag_suffix }} - v${{ steps.apply-semantic-version.outputs.version }}'
          prerelease: ${{ github.event_name == 'pull_request' && true || false }}
          tag: ${{ steps.apply-semantic-version.outputs.version_tag }}
          version: 'v${{ steps.apply-semantic-version.outputs.version }}'
        env:
```

* Added the following permissions to the job that runs the above steps.
```
    permissions:
      contents: write
      pull-requests: read
```

* Created a pull request for the `test` branch (targeting the `main` branch) to trigger the CI workflow to run.

**Expected Results**
* The CI workflow should run and a new GitHub Release should be created.

**Actual Results**
* The CI workflow runs and the following error is reported in the GHA workflow log.
```
Warning: ololabs-playground/keith-gha-actions-test: Invalid config file
{
  name: 'event',
  id: '463073[16](https://github.com/ololabs-playground/keith-gha-actions-test/actions/runs/4630731678/jobs/8192704466#step:11:17)78',
  stack: 'Error: Configuration file .github/release-drafter.yml is not found. The configuration file must reside in your default branch.\n' +
    '    at getConfig (/runner/_work/_actions/release-drafter/release-drafter/v5/dist/index.js:1427[19](https://github.com/ololabs-playground/keith-gha-actions-test/actions/runs/4630731678/jobs/8192704466#step:11:20):13)\n' +
    '    at async drafter (/runner/_work/_actions/release-drafter/release-drafter/v5/dist/index.js:142347:[20](https://github.com/ololabs-playground/keith-gha-actions-test/actions/runs/4630731678/jobs/8192704466#step:11:21))\n' +
    '    at async Promise.all (index 1)',
  type: 'Error'
}
Error: Invalid config file
```

### LINKED ISSUE #1338: Allow inputing config-name file contents
If my workflow requires multiple release-drafter configs that are dependent on the contents of the repo, I want to dynamically supply those as a `config-name` parameter. This will help not having to create/delete files in the `.github` directory as new configs are necessary. It would nice to input the config file as text or be able to write the config during the workflow run.

### LINKED ISSUE #1401: Add dry-run option to v6
## Description

I was hoping setting `disable-releaser: true` would operate like a `dry-run`, however it appears to completely disable the action so that no output is generated. I noticed that in v6 this input was removed all together. Instead, I was thinking a `dry-run` option would be useful in order to use the action outputs for other use-cases without generating a release.

My specific use-case involves using the action twice, in tandem, in order to set the release to a custom tag that includes the git hash, which isn't available in the release template.

### LINKED ISSUE #1413: Lack of permissions to write does not fail the execution
I suspect that this might be a regression with v6 but if release drafter is unable to create the release notes due to lack of permissions, it should fail the execution, but now it does succeed.

Here is the proof https://github.com/ansible/ansible-dev-tools/actions/runs/11592274941/job/32273792197

"Error: Resource not accessible by integration" is clearly visible in the output but the job reported success.

### LINKED ISSUE #1460: Node20 will reach end-of-life (EOL) in April of 2026

### LINKED ISSUE #1476: Next major version - help needed 🙏
I'd like this issue to be home for the next major release - v7 ! 🥳

This version will take full advantage of [actions/typescript-action](https://github.com/actions/typescript-action). This, we think, will improve maintainability and stability in some regards.

The detailed list of actual end-user changes still needs to be written down properly. I have created a [CHANGELOG](https://github.com/release-drafter/release-drafter/blob/refactor/use-typescript-template/CHANGELOG.md) file to track those when we find them along the way during refactor.

A similar major refactor was started by @jetersen back in 2022 in #1173 (issue) and #1204 (draft PR)

In the recent [Call for maintainers](https://github.com/release-drafter/release-drafter/issues/1455), we briefly discussed next major improvements, and I thought switching to `actions/typescript-action` would be a great first step towards getting to implement them all.

👉 I you feel like @jetersen's #1204 be our starting point, and have already been through it's code, please share !
👉 If you've read through all the above and felt like you have another good idea, please share !
👉 I you feel like contributing, please open a PR targeting branch [`refactor/use-typescript-template`](https://github.com/release-drafter/release-drafter/tree/refactor/use-typescript-template)

cchanche:
# Major Release `v7` :rocket:

This release focuses on a modernized codebase to ease maintenance, enabling future long-awaited features.

Major takeaways are :

- sources rewritten to typescript
- drop the Probot framework in favor of [actions/typescript-action](https://github.com/actions/typescript-action)
- drop node20 support in favor of node24

Read below for a detailed list of features and breaking changes.

## Features

- Read your config from a local file using the `file:` scheme for the `config-name:` input. This enables dynamically-generated configs. Read more in the [Configuration Loading](https://github.com/release-drafter/release-drafter/blob/master/docs/configuration-loading.md) article

- Introduce the `token` input, replaces the (need for) `GITHUB_TOKEN` env variable. Defaults to `${{ github.token }}`

- Introduce `dry-run: true` mode. Avoids write operations and logs payload content instead. Works for drafter & autolabeler. Enabled from action inputs.

- No need for custom `GHE_HOST` variable for Github Enterprise Server runners anymore. The correct api and gql endpoints will be detected using `GITHUB_API_URL` and `GITHUB_GRAPHQL_URL`. The runner registration script on your self-hosted runner should already configure those. (i.e. `./config.sh --url https://your-ghes-instance.com`). See [`actions/github@2.2.0`](https://github.com/actions/toolkit/blob/main/packages/github/RELEASES.md#220), [variable reference](https://docs.github.com/en/actions/reference/workflows-and-actions/variables).

- When one of the manual version overrides (`version`, `tag`, `name`) is provided, version computations now consistently use that input-derived version as the reference for `$NEXT_*` and `$RESOLVED_VERSION` values.

- Introduced `$NEXT_PRERELEASE_VERSION_PRERELEASE` for directly accessing the prerelease suffix component (for example `-beta.0`) when templating.

## BREAKING Changes

- drop support for node 20. Your runners will require the `node24` runtime. See GitHub [depreciation notice](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) to reconfigure your runners if needed.

- move Autolabeler to a dedicated action. The corresponding `disable-releaser` and `disable-autolabeler` were removed.

  ```yaml
  steps:
    # runs drafter
    - uses: release-drafter/release-drafter@latest
    # runs autolabeler
    - uses: release-drafter/release-drafter/autolabeler@latest
  ```

- Drop the `include-pre-releases` config. Use the `prerelease` and `prerelease-identifier` instead. Read [#1515](https://github.com/release-drafter/release-drafter/pull/1515) for details

- Disabled use of the `references` config. Use [`on` conditions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#on) instead :

  ```yaml
  # .github/workflows/release-drafter.yaml

  on:
    push:
      # Sequence of patterns matched against refs/heads
      branches:
        - main
        - "mona/octocat"
        - "releases/**"
      # Sequence of patterns matched against refs/tags
      tags:
        - v2
        - v1.*
  ```

- Config syntax of autolabeler made use of Joi's `array.single()` syntax, which is not supported anymore. The new implementation of config validation runs on zod, which does not support a similar feature natively. The configuration fields that relied on `array.single()` are :
  - `autolabeler[].branch`
  - `autolabeler[].title`
  - `autolabeler[].files`
  - `autolabeler[].body`
  - `categories[].labels` (you may have used `label` instead, still supported)
  - `version-resolver.major.labels`
  - `version-resolver.minor.labels`
  - `version-resolver.patch.labels`

  To migrate, do the following :

  ```yaml
  # .github/release-drafter.yaml

  # Before
  autolabeler:
    - label: bug
      files: "src/bugs/*"

  # After
  autolabeler:
    - label: bug
      files:
        - "src/bugs/*"
  ```

- Action inputs are now validated against a zod schema. This may raise some unexpected errors for some of your already-existing inputs.

- removed option `"legacy"` for the `publish` input/config. Use `"false"` instead.

- following the removal of probot, we replaced the [Probot Config](https://github.com/probot/probot-config) plugin with custom-made syntax and capabilities. Key takeaways are :
  - fallback to your org/user `.github` repo does not work anymore. You will need to be explicit about it. (ex: `config-name: .github:/configs/release-drafter/main.yaml`)
  - the `_extends` config-keyword now uses the same syntax as the `config-name:` input. Read more in the [Configuration Loading](https://github.com/release-drafter/release-drafter/blob/master/docs/configuration-loading.md) article

- If you were using a specific access token using `env.GITHUB_TOKEN: $MY_CUSTOM_TOKEN`, migrate to the new `token:` input

  ```yaml
  jobs:
    steps:
      # Before
      - uses: release-drafter/release-drafter@v6
        env:

      # After
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
  ```

- `$INPUT_VERSION` is no longer available as a template variable. Use `$RESOLVED_VERSION` instead.

- When one of the manual version overrides (`version`, `tag`, `name`) is provided, `$NEXT_MAJOR_VERSION`, `$NEXT_MINOR_VERSION`, and `$NEXT_PATCH_VERSION` are now calculated from that input-derived version, rather than from the previous release.

- For first-release scenarios (no previous release and no parseable version input), default next-version values are now accurately derived from `v0.1.0`.

- `$NEXT_PRERELEASE_VERSION` now follows the configured `version-template` rendering behavior instead of acting like a prerelease-fragment-only value. If you need only the prerelease suffix part, use `$NEXT_PRERELEASE_VERSION_PRERELEASE`.

- json _meta-schema_ for our `schema.json` moved away from version `draft-07` in favor of version `draft/2020-12`. You may need to adapt your validation pipelines (see: #1575)

---

My intent for this new version is to :

- [x] Introduce node 24 (runtime & development)
- [x] Transition to Typescript
- [x] Transition to ESM
- [x] Split drafter and autolabeler into two separate actions (same repo)
- [x] Drop Probot and drop support for the (legacy) app
- [x] Read config from runner instead of / on top of default branch
- [x] Port core v6 feature
- [x] Config loading strategy
  - [x] Support v6 loading (probot config plugin)
  - [x] Config from runner
- [x] Port every test
  - Coverage : 92.06% !
- [x] JSON schema from configs
- [x] CI/CD
- [x] Cleanup docs
- [x] Make sure yaml inputs align with schemas

closes #1476
closes #1401
fixes #335
fixes #1338
fixes #1460
fixes #1334
fixes #1274
fixes #1014
fixes #336
fixes #1413
superseds #1423
helps with #1450
helps with #1469
helps with #1335
helps with #1313

### DIFF
--- a/dist/chunks/common.js
+++ b/dist/chunks/common.js
@@ -32,7 +32,7 @@ var __exportAll = (all, no_symbols) => {
 	return target;
 };
 var __copyProps = (to, from, except, desc) => {
-	if (from && typeof from === "object" || typeof from === "funcion") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
+	if (from && typeof from === "object" || typeof from === "function") for (var keys = __getOwnPropNames(from), i = 0, n = keys.length, key; i < n; i++) {
 		key = keys[i];
 		if (!__hasOwnProp.call(to, key) && key !== except) __defProp(to, key, {
 			get: ((k) => from[k]).bind(null, key),


jetersen:
I think we should look into separating the autolabeler into a separate action.
You can have multiple actions in the same git repo:
See this approach: https://github.com/GitTools/actions (note gitversion and gitreleasemanager directory)

cchanche:
Ah very nice, I'll look into this ! Thank you @jetersen 

I was also wondering how to best separate concerns there, but had not yet considered having two separate actions. If this is going to be a breaking update, we may also go all the way. I think this is a good idea.

I'll look into gitTools ! I also like the monorepo approach with yarn or pnpm - helps sharing code easier. Maybe gitTools also implements something similar I still need to read the details

cchanche:
Ok so what I understand from what GitTools does is : 

- a single repo hosting multiple actions
- clever build configs (using Vite) that outputs `main.js` files for each action
- each action is in a dedicated sub-folder you have to specify when using the action in a workflow
  ex: `gitversion/execute/action.yml` with `gitversion/execute/main.mjs` can be used as :
  ```yaml
  steps:
    - uses: gittools/actions/gitversion/execute@v4.2.0
  ```
- no need for dedicated monorepo tools (like lerna, nx or pnpm/yarn etc...) - the build-system does the heavy lifting
- the repo's releases are the available versions for each action (here, `v4.2.0`)

cchanche:
I think I like it ! Configuring the build system will be the hard part, but this is a one-off job anyway.

About how users would reference our action, I suggest the following : 

```yaml
steps:
  # targets root `action.yaml` - runs `drafter/main.js`
  - uses: release-drafter/release-drafter@v4.2.0
  # targets `autolabeler/action.yaml` - runs `autolabeler/main.js`
  - uses: release-drafter/release-drafter/autolabeler@v4.2.0
  # [optional] also runs `drafter/main.js`
  - uses: release-drafter/release-drafter/drafter@v4.2.0
```

cchanche:
Damn, so that was interesting !

Had a LOT of trouble making Jest work with the new _module_ way of doing things in vite+typescript. Turns out vitest works like a charm and solves quite a lot of weird hacks I had to do. Vitest's API is similar if not the same, but mocks behave a bit differently. I'll have to see if porting the rest of the tests is not too painful, but I think I've done the most difficult part already (mocking the context & the config)

For the bundling part, I went for `preserveModules: true` in rollup's config. This preserves our source's structure in `dist`, and so are both of our entrypoints. Meaning `src/actions/drafter/run.ts` ends up in `dist/drafter/run.js` with it's content being the bundled version of the original file - nothing more, nothing less.

Without preserving modules, tracking the files themselves in the output was manageable, but they would contain whatever Rollup decided was best to include according to it's bundling algorithm. Since our entrypoints are nothing but js modules with side-effects, this turned-out to be a single bundle with two side effects, no matter whether you tried executing `dist/drafter.js` or `dist/autolabeler.js`

`preserveModules: true` keeps each module's side effects in their corresponding _module_, which is what we want.

GitTools went for another approach I was not too keen on implementing. Using `preserveModules: false` and `manualChunk: () => {...}`, they organize their `dist` folder how they please and drastically reduces the number of files in `dist`. `dist` stays an actual bundle with some clever chunks. However, they did not bother with their entrypoints being bundled together with the runtime code... They manually maintain production entrypoints in git. For [instance](https://github.com/GitTools/actions/blob/main/gitreleasemanager/discard/main.mjs) : 

```yaml
// gitreleasemanager/discard/action.yml - (versioned)
runs:
  using: node20
  main: main.mjs
```
```js
// gitreleasemanager/discard/main.mjs - (versioned)
import { run } from '../../dist/tools/lib.mjs' // have to pray this stays valid...
await run('github', 'gitreleasemanager', 'discard')
```

With `preserveModules: true` here is how it turned-out for us :

```yaml
// drafter/action.yml - (versioned)
runs:
  using: node24
  main: ../dist/drafter/run.js # side-effect, guaranteed by preserveModules: true
```
```js
// dist/drafter/run.js - (generated)
import { run } from "../src/actions/drafter/runner.js"; // 'src' here is under dist/src
await run();
//# sourceMappingURL=run.js.map
```

src/common/load-config-file.ts
```diff
@@ -0,0 +1,24 @@
+import path from 'path'
+import { existsSync, readFileSync } from 'fs'
+
+export const loadConfigFile = (configFilename: string): string => {
```
jetersen: This removes an important feature and use case of octokit config plugin, the ability to load your config from `yourorg/.github/.github/release-drafter.yml` and many orgs are using this approach. One such example is this: https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.yml

cchanche: Oh, I am guessing you mean the ability to fetch from the default branch (through the network) ?

I had the feeling fetching the file from the runner (@actions/checkout required) would satisfy most use cases.

It also enables live-testing changes to the config on a separate branch, which I thought was nice too !

Do you wish to have the _fetching on the default branch_ as a default or as a fallback ? (as discussed on that other issue I need to find the number of)

cchanche: Oh right sorry just reading your comment again !

You mean some orgs have a special `.github` repo (=/= folder) with some configs

Keeping that feature is indeed interesting for them

cchanche: We may want to grab [this function from the plugin](https://github.com/probot/octokit-plugin-config/blob/main/src/util/get-config-files.ts) somehow 

I may be able to pull it off through a dependency. Since rollup tree-shakes everything for us, that should not clutter the build too much with probot stuff

Otherwise I'll implement it manually that's not that hard either.

Thank you for pointing this out !

jetersen: Yes, I meant the special `.github` repo :)

Properly makes sense to use [`actions/github`](https://github.com/actions/toolkit/tree/main/packages/github#actionsgithub) and there you can add the octokit plugin.
Later if we want to support other runtimes (cli) we just have to hydrate the octokit client ourself instead of using the `actions/github` or simply support taking a hydrated octokit client with plugins in `src/core` and then have `src/cli` and `src/action` create a hydrated octokit client with plugins.
Sorry I am using loose terms here for the release drafter action and potential CLI.

cchanche: No dw it's pretty clear

Hoisting runtime-dependant stuff somewhere (ex : src/core) will be required for the cli indeed

Looking at all imports from `@actions/*` sums it up pretty nicely : 

- The octokit client
- The context object
- action inputs and outputs
- logging utility

I think I can already picture how each of these could be wrapped in some way, such as other runtimes may be supported !

Now about config loading, taking a deeper look at [the probot plugin](https://github.com/probot/octokit-plugin-config), It looks like it relies only on a simple octokit client. Lucky for us, it does not bother instanciating the client by himself, and requires us to give it to him.

Simply matching v6 config loading, and doing it properly with hoisting this GitHub stuff in a corner somewhere, looks easy enough🤞


Now there is a thing that bugs me about this new config-loading strategy I would have liked to introduce (as suggested in this PR already).

Loading the config from the runner is easy enough. But now that I see the full feature set of the plugin (ie `.github` repo & `_extends`), I am unsure how to go about inserting the _local-file-system-loading_ thing.

To clarify, let's say there's the `default-branch`, the `.github repo` and the new `local` strategies.

- What should be the default strategy ?
- What strategy should be the fallback to another ?
- If `local` is added, should it also be added to the available strategies supported by the `_extends` feature ?

I'll try to think about it and find a nice solution to this.

Whatever the answers to these questions end up being, I believe they will require us to rewrite the plugin to insert our new `local` strategy stuff.

cchanche: So, I thought about it, and I may have been overthinking this 😅 

> - What should be the default strategy ?
> - What strategy should be the fallback to another ?

The config-loading order may simply be : `local` > `default-branch:./.github` > `[owner]/.github` repo

I think this is the simplest implementation, because we can rely on the plugin without having to change anything for the fallback probot loading style.

It may also be the least invasive way of introducing the `local` config-loading capability, despite inserting it as the default.

Users that do not want to adopt `local` :

- may not have their repo checked-out in their workflows anyway, since it is not required in `v6`
- if they do (and don't want to adpot new strategy), they can just `$ rm .github/realease-drafter.yml` the file beforehand at runtime
- if they cannot touch the runner's file-system (maybe workflow include a `git diff --porcelain` step like ours), they'll just have to rename / move the config-file in their repo. 
  - `rename / move` the file may be breaking if you _absolutely want_ to use `default-branch:./.github` over `local` (ex: to benefit from the `_extends` stuff)

Users that want to use this new strategy :

- most probably don't use the `[owner]/.github` repo strategy, and have the config-file at hand in their runners - they'll just need to add `@actions/checkout`


Additional features that may require some attention if we want to be 100% inclusive of all possible scenarios :

- add the `_extends` capability to `local`ly loaded configs
- add an input to disable the `local` strategy

I believe these may come later !

jetersen: the plugin already takes care of the _extends and if file is present in the default branch. Properly it would be easier to implement it ourself to ensure it integrates nicely with local + extends or default branch + extends because local excludes needing to fetch default branch imho, but it might not exclude asking it to use `_extends: .github` or `_extends: my-other-org/repo` :thinking:

cchanche: Yes, agreed, the local fs loading kind of supersedes the default branch capability...

What could also be nice, if we implement this ourselves (which does not look that hard), is to come up with a syntax for specifying a target config file, that would also be the syntax to use in the `_extends` keyword.

Maybe start with something like `[owner]/[repo]:[filepath]@[ref]`, and then : 
- omit only the `@[ref]` and you'll target the default branch
- omit only the `[owner]/[repo]:` and you'll get your own repo
- omit both `[owner]/[repo]:` and `@[ref]`, and we'll fetch the thing locally
    - get from runner if specified in `config-name:` input
    - get from same context as current file if specified as `_extends`

To be more generic,  `[owner]/[repo]:` and `@[ref]` are just overrides of the current runtime context variables `github.context.repo` and `github.context.ref` - with the exception of the github-actions runtime where we fetch from the runner if nothing is overriden.
 
This, however, changes the current behavior : people using their `[org]/.github` repo will need to be explicit about it

cchanche: @jetersen 👇  here is what i came up with, tell me what you think boss 


image

<sup>generated by [RR - Railroad Diagram Generator](https://www.bottlecaps.de/rr/ui)</sup>

full spec written as (empty) test-suites are in [these commits](https://github.com/release-drafter/release-drafter/pull/1475/changes/256f3c63328863b933fc4e96e2adfd97bff4fc84..859cfb4a54a48a8ddadd8d789f55a6335e13daee) - which is convenient if you feel like commenting !

full implementation of the loader is underway - I feel good about this

jetersen: Would prefer we assume it is the same org when they only type `repo` so we keep the same behavior as it is currently. Check the [`_extends`](https://github.com/probot/octokit-plugin-config#the-_extends-key) docs

cchanche: Yup that's what I had in mind ! I have read how probot-config does _extend, and that was precisely what inspired me to craft this syntax.

I have pushed my implementation, ~but I still need to~ :
- [x] write a nice readme entry
- [x] finish writing tests for the whole thing
- [x] write a breaking-change changelog entry
- [x] plug new implementation to the rest of the app
- [x] rewrite config-loading mocks for non-config-loading related tests


In the process, I figured that being explicit about where you want to fetch your config from (remote or local) greatly simplified things in the code.

Here is what I came up with (syntax-wise) : 

image




- `config-name: release-drafter.yaml`
  - assumed scheme : `github:`
  - assumed repo, owner & ref : the same as the runtime
  - parsed path : `.github/release-drafter.yaml`
- `config-name: .github:/config/custom.json`
  - assumed scheme : `github:`
  - parsed repo : `.github`
  - parsed path : `config/custom.json`
  - assumed owner : same as runtime
  - assumed ref : none since repo is different - octokit will fetch from default branch
- `config-name: file:../src/configs/rd-generated.yml`
  - parsed scheme: `file:`
  - parsed path: `src/configs/rd-generated.yml`
- `_extends: ./common.yaml`
  - assumed scheme: `github:`
  - assumed repo, owner & ref : same as parent
  - parsed path : `[parent dirname]/common.yaml`

src/actions/autolabeler/main.ts
```diff
@@ -0,0 +1,112 @@
...
+export const main = async (params: { config: Config }) => {
+  /**
+   * TODO :
+   *
+   * 1. [new] check event is 'pull_request' - was handled by probot using app.on()
+   * 2. get PR's details. I think @actions/github's context is fine
+   * 3. find changed files. @actions/github's octokit instance may be useful
+   * 4. create Set() of labels to add based on config's autolabeler params
+   * 5. add labels. @actions/github's octokit instance may be useful
+   */
```
jetersen: Is this relevant still?

schema.json
jetersen, line 1: This will break schemastore https://github.com/SchemaStore/schemastore/blob/9c361a8af71cc4c43a80269ecb20e0e5ce313f42/src/api/json/catalog.json#L7100 😅

cchanche: Oh yes damn, did not know this was a thing ! I'll relocate the schema to the root folder.

Thank you for noticing !

docs/configuration-loading.md
bcheidemann, line 1: Since this is a breaking change anyway, would it make sense to rename `_extends` to `extends`?

This would open up the possibility of maintaining backwards compatibility by handling `_extends` the same way as it is currently in v6, and would also allow retrofitting the new `extends` syntax to v6, making for a smoother transition.

cchanche: That is indeed a nice idea ! There would be some work to be done to adapt the probot config plugin to only trigger for that `_extends` key, but nothing impossible.

However, not using the probot plugin to load the first file will be breaking no matter what. Maybe the solution is to use the probot config plugin _full-stack_ on the `config-name:` input (like before), and introduce a new `config:` or `config-file:` input for this new configuration loading strategy.

Also, I came to like the idea of having that `_` prefix for this key, since it elegantly (IMO) identifies `_extends` as not being a configuration parameter like the others.

I saw you saying this felt like a private class-attribute/method to you, and was a bit confusing. Maybe that's also what strict Object Oriented hackers feel, and finding another better-suited name may actually be interesting to do here.

I like to think, however, that Probot already had this debate on their side, and us inheriting from (copying) that is the safest approach...

cchanche: Oh, and maybe I should be more specific inside this migration docs. The new syntax here for configuration loading is really not that far off the old one (Probot's one).

I believe this won't be a breaking change for 90%+ use-cases

bcheidemann: I can see the logic behind using the _ prefix to signal that this isn't a normal config option. And in any case, sticking with established conventions is generally a good default.

Regarding whether this implies that it's some kind of private option, I don't think it's a big issue as long as it's documented along side the other fields in the config file. That would pretty clearly signals it's fine to use.

cchanche:
@jetersen have you had the time to further review ? How do you feel about this ?

I have been using this on my projects with no particular issues, and I am confident this is stable. I simply still don't know if you are ok with me making such (major) changes on the repo.

Even if this is not stable, I'll be around to fix things anyway

jetersen:
I'll do a deep dive this weekend 😬

jetersen:
Ended up being sick over the weekend, so hopefully this weekend :pray:

cchanche:
No worries at all @jetersen! Take all the time you need. Hope you feel better soon!

cchanche:
- [x] rebase on `v6.4.0` for:
 - revert: deprecate include-pre-releases (#1523) @cchanche
 - feat: add include-pre-releases to action inputs (#1525) @cchanche
 - feat: Add support to excluded-paths (#1522) @guimorg

dist/index2.js
jetersen, line 1: index2.js seems sus 😅 Is this a mistake?

autolabeler/action.yml
```diff
@@ -0,0 +1,23 @@
+name: 'Release Drafter - Auto Labeler'
+description: 'Automatically labels pull requests based on your configuration.'
+author: release-drafter
+branding:
+  icon: edit-2
+  color: orange
+runs:
+  using: node24
+  main: dist/autolabeler/run.js
```
jetersen: drafter/action.yaml has `..`

```suggestion
  main: ../dist/autolabeler/run.js
```

eslint.config.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+    languageOptions: {
+      globals: {
+        ...globals.node,
+        Atomics: 'readonly',
+        SharedArrayBuffer: 'readonly'
+      },
+
+      parser: tsParser,
+      ecmaVersion: 2023,
```
jetersen: nodejs v24 can use ecmaVersion 2024 👍
```suggestion
      ecmaVersion: 2024,
```

package.json
```diff
@@ -1,69 +1,78 @@
...
-    "lodash": "4.17.23",
-    "probot": "12.2.5",
-    "regex-parser": "2.2.11",
-    "semver": "7.3.7"
+    "@actions/core": "^1.11.1",
+    "@actions/github": "^9.0.0",
+    "compare-versions": "^4.1.3",
+    "escape-string-regexp": "^5.0.0",
+    "ignore": "^7.0.5",
+    "lodash": "^4.17.21",
```
jetersen: I wonder if we still need lodash? maybe we can clean it up in another PR at least worth exploring.

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "newLine": "lf",
+    "noImplicitAny": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "./dist",
+    "pretty": true,
+    "strict": true,
+    "strictNullChecks": true,
+    "target": "ES2022",
+    "module": "ES2022",
```
jetersen: same as earlier ES2024 seems like the more approriate target

```suggestion
    "lib": ["ES2024"],
    "types": ["node", "vite/client"],
    "newLine": "lf",
    "noImplicitAny": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "outDir": "./dist",
    "pretty": true,
    "strict": true,
    "strictNullChecks": true,
    "target": "ES2024",
    "module": "ES2024",
```

src/common/config/README.md
```diff
@@ -0,0 +1,73 @@
+# Configuration loading
+
+TODO
```
jetersen: 🤫

src/actions/drafter/config/merge-input-and-config.ts
```diff
@@ -0,0 +1,147 @@
+import * as core from '@actions/core'
+import isBoolean from 'lodash/isBoolean.js'
```
jetersen: 😰

```suggestion
import { isBoolean } from 'lodash'
```

src/actions/drafter/lib/build-release-payload/get-version-info.ts
```diff
@@ -0,0 +1,369 @@
...
+import * as semver from 'semver'
+import { renderTemplate } from './render-template'
+import { Config, ExclusiveInput } from '../../config'
+
+type Release = Exclude<
+  Awaited<ReturnType<typeof findPreviousReleases>>['lastRelease'],
+  undefined
+>
+
+// TODO is this not supposed to be $MAJOR.$MINOR.$PATCH$PRERELEASE ?
```
jetersen: 🤫

jetersen (changes requested):
ONLY two TODO maybe cleanup or fix the todos? 
Then there is one import mistake which should be fixed :pray:

jetersen:
I would suggest fixing the last stuff and fix the merge conflict / rebase.
Properly hold of merging #1524 if it is even needed after going forward with the v7 release :clap:

dist/index2.js
jetersen: This is because our dependency. My main issue with this is the vite build it feels wrong. I think we can simplify a lot more here.

cchanche: I'll look for something ! Do you feel like we should have a single bundle file (or fewer) ?

I also came to think our bundle should be side-effect-free, especially since it is published to npm as a package.

Additionally, if we are to release a CLI (or more generally, a generic runtime) someday, this will definitely be needed.

Bottom line is, I changed my mind about what I said [here](https://github.com/release-drafter/release-drafter/pull/1475#issuecomment-3791774150) : 

> GitTools went for another approach I was not too keen on implementing.

Let's change that and do the same thing. Thank you for pointing it out !

autolabeler/action.yml
cchanche: Damn, nice catch ! Just realized I did not even try to use `autolabeler` in any way yet...

My bad, and again, nice catch

eslint.config.mjs
cchanche: Definitely !

package.json
cchanche: I agree, I would also like to clean up the regex dependency above because I think a saw a new native thing that looks like it does the same thing.

I'll scope this to another PR and work on it 😃

tsconfig.json
cchanche: Ah yes, again ! I'll fix this

src/common/config/README.md
cchanche: Ah very nice catch ! This whole RAEDME was moved to docs/configuration-loading.md, forgot to delete...

src/actions/drafter/lib/build-release-payload/get-version-info.ts
cchanche: Yes so I wanted to avoid breaking anything when porting the code...

Now that tests are all set-up, I believe it is indeed a good time to address this ! 

I'll look into it

package.json
jetersen: I might be cooking something on this and the whole vite build :sweat_smile:

cchanche:
> I would suggest fixing the last stuff and fix the merge conflict / rebase. Properly hold of merging #1524 if it is even needed after going forward with the v7 release 👏

Thank you for looking into it @jetersen ! As commented above, I'll take care of fixing your suggestions, and then proceed to merge master.

About #1524 , I'll see if a `v6.5.0` or `v6.4.1` is necessary for some other matter in the meantime (of me squaring-up `v7`), and simply add #1524 to the release.

Otherwise, I hope `v6.4.0` can be the last pure js release 🥳

dist/index2.js
jetersen: I did some vite bundle visualization stuff and for instance named imports for Zod was a solid improvement among others, I'll open a PR with my changes for the vite build improvments and bundle size

jetersen: https://github.com/release-drafter/release-drafter/pull/1526

jetersen: I don't think there is much of a point to do npm release now that we removed the probot part. We should properly stop releasing a package till we create a probot or cli after v7 :thinking:

cchanche: > I don't think there is much of a point to do npm release now that we removed the probot part. We should properly stop releasing a package till we create a probot or cli after v7 🤔

100% Agree on this ! 

I thought for a moment you wanted to keep it for the _Weekly Download_ thing, because there's no other real metric with GHA

autolabeler/action.yml
cchanche: Fixed in #1526

eslint.config.mjs
cchanche: Fixed in #1526

package.json
cchanche: Fixed in #1526

tsconfig.json
cchanche: Fixed in #1526

dist/index2.js
jetersen: regarding your point of side effect free, that is hard when we depend on actions/core and actions/http-client and actions/github which all have side effects.
Also we the fact that many of our dependencies are still cjs is an issue.

Either we find dependency alternative or try and reduce the side effects.
For instance if we were to create a CLI I would want it to be 100% ESM.

cchanche: How do CJS and side effects relate? As you have probably guessed already, I am not an expert when it comes to bundling. Is rollup (rolldown) not wrapping everything in esm for us already?

`@action/*` deps are both 
 1. runtime-specific (they kind of expect to be run in GHA)
 2. platform dependant (they expect to be run for Github remotes)
 2. produce side effects
 
To address 1. & 2., I thought we should end up creating abstraction for them. The easy one would be the logger, but there's the octokit instance, the context etc... I am thinking something like :

```ts
const logger = new Logger({runtime: 'GHA', platform: 'github'}) // proceeds to use @action/core
```

Would this alleviate their side effects ?

Anyway, I think this should be for future work

jetersen: Definitely future work, it was mainly a raising it as a concern I would have. Yes using octokit directly would allow us to avoid these issues as it is already ESM and no side effects. It doesn't directly affect these PR changes for now.

src/actions/drafter/lib/build-release-payload/get-version-info.ts
cchanche: As mentioned in https://github.com/release-drafter/release-drafter/pull/1475/commits/fb9621ac2eebfa3f80ad5eadda96bbe3d67887ad 's message, this looks more complicated than expected to me

Since this may end up being a breaking change, I'd like to do it here. As of writing, I believe the breaking changes may only involve edge use-cases of some variables, sometimes even undocumented ones... but still _breaking_.

Still on it

src/tests/drafter/drafter.e2e.test.ts
```diff
@@ -1207,6 +1207,296 @@ describe('drafter e2e', () => {
...
+          await mockContext('push')
+          mocks.config.mockReturnValue('config-with-component-helpers-default')
+          const scope = nockGetAndPostReleases({
+            fetchedReleases: ['release']
+          })
+          const gqlScope = mockGraphqlQuery({
+            payload: 'graphql-commits-merge-commit'
+          })
+          await runDrafter()
+          expect(mocks.postReleaseBody.mock.lastCall).toMatchInlineSnapshot(`
```
jetersen, line 1223: To update snapshots: `vitest -u` @cchanche 😅

cchanche: I want to review the faulty snapshots one by one first to make sure I have a viable solution

I have not had the time to go through all of them 😉

cchanche: Done in [304a69f](https://github.com/release-drafter/release-drafter/pull/1475/commits/304a69f36261a838f558729c2e093b72ebf4d44e)

src/actions/drafter/lib/build-release-payload/get-version-info.ts
cchanche: See https://github.com/release-drafter/release-drafter/compare/43c692bf1924957f1b21647ee18b2b5c6203de95...79d3033af2340df2231a4368bdf72b07432743f8

cchanche: Updated PR description with resulting breaking changes

pbrousseau-sdx:
I suppose I am doing a bit of beta-testing.  I have an action set up on PRs:
```
jobs:
  autolabeler:
    permissions:
      contents: read
      pull-requests: write
    if: ${{ github.event_name == 'pull_request' }}
    runs-on: ubuntu-latest
    steps:
      - uses: release-drafter/release-drafter/autolabeler@refactor/use-typescript-template
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
```
Updating the `release-drafter` is the only change in the PR.

Trying this out, however, I get:
```
Run release-drafter/release-drafter/autolabeler@refactor/use-typescript-template
  with:
    token: ***
    config-name: release-drafter.yml
file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:49
	throw Error("Calling `require` for \"" + x + "\" in an environment that doesn't expose the `require` function. See https://rolldown.rs/in-depth/bundling-cjs#require-external-modules for more details.");
	      ^

Error: Calling `require` for "node:net" in an environment that doesn't expose the `require` function. See https://rolldown.rs/in-depth/bundling-cjs#require-external-modules for more details.
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:49:8
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:219:2
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:23:48
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:420:19
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:23:48
    at file:///home/runner/work/_actions/release-drafter/release-drafter/refactor/use-typescript-template/dist/chunks/common.js:15622:45
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.13.1
```

Perhaps I am missing something?

jetersen:
@pbrousseau-sdx thanks for checking it out, definitely missed the fact that we did not have type module in the package.json. It should fix it by setting the extension to .mjs when vite builds :sweat_smile:

cchanche:
@pbrousseau-sdx I've added smoke tests in CI here to showcase your problem - we will address this shortly ! 

This required a small feature around `dry-run` mode I've added

- [x] link issues related to feature-request of `dry-run`

cchanche:
All looks good @pbrousseau-sdx !

Issues fixed thanks to your vigilance : 

- use `platform: "node"` (instead of `browser`) in Vite to allow requiring some modules (01d9e01bedf0fd83451886b7c2fedbfd9e4c71c9)
- defer defining defaults to the github token, such as `${{ github.token }}` from inputs has a chance to be the default (c372ef2216b1ee88b80c51a2faeb3d0d1a8746ac)
- force our build output to rely on actual process.env at runtime instead of empty object (0b2be7fb66ac85e2b249318176a9172c22025ae3)

pbrousseau-sdx:
Can confirm; no longer seeing the above error; thanks!

jetersen:
I missed the config change, I really think extending from `.github` org repo by default was one of best features :sweat_smile: I think many would be sad to see it gone.

PR #1478, opened 2026-01-19
build: replace ncc with rollup

cchanche:
I work on a few different machines, 3 of which are windows-based.

Because of this, I was unable to pass the tests in CI because of ncc's https://github.com/vercel/ncc/issues/638 and https://github.com/vercel/ncc/issues/992 issues.

There probably was a way around that, but I figured using another build system was not that hard, and also way overdue. See [trends](https://npmtrends.com/@vercel/ncc-vs-esbuild-vs-parcel-vs-rollup).

I went with rollup since that is the bundler used by https://github.com/actions/typescript-action

EDIT: initial assumptions about the actual issue were incorrect. read below.

### DIFF
--- a/.github/no-unstaged-files.sh
+++ b/.github/no-unstaged-files.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 
-git diff
+# The following makes git "touch" the files at leas one,
+# and applies the contents of .gitattributes the same way
+# a local developer would.
+# This helps alleviate line-endigs inconsistencies our tools
+# and dependencies might produce.
+git add .
+git reset
+
 if [[ "$(git status --porcelain)" != "" ]]; then
   git status -u all
   echo "::error::💥 Unstaged changes detected. Locally try running: yarn prettier && yarn lint --fix && yarn build"


cchanche:
Ok so that's a funny one ! 

Found another problem with rollup using a different moduleId for the ./index.js file in the bundle on my machine. This was because I have the repo checked-out in a worktree, which is basically a directory that has an arbitrary name. Mine is `master` -> rollup would output a `requireMaster` function in the build somewhere, and the CI would do the same but call it `requireReleaseDrafter`.

But now I keep having this crlf problem can't figure-out why

cchanche:
Damn ok so for some reason, our `dist/index.js` files contained mixed line endings (CRLFs **AND** LFs).

I have not been able to see why that is... Probably because my node_modules is full of CRLFs (found out using `$ find . -not -type d -exec file "{}" ";" | grep CRLF`) - but rollup is supposed to output a clean LF-terminated file so 🤷 

To see those line endings in the output i did : `od -c dist/index.js | grep '\\r' | head`

When I would `git add dist/index.js`, git would _"touch"_ the file, and apply whatever config my repo has about line endings. For realease-drafter, it applies what's inside `.gitattribute`, meaning cleaning up CRLFs for LFs. The file would then disappear from the added changes because the _clean-LF_ version of it was identical from the one at HEAD -> I could not commit the CRLF + LF file, and that is a good thing.

If one would run a `git diff` **before** adding file using `git add`, you should see :

```
> git diff
warning: in the working copy of 'dist/index.js', CRLF will be replaced by LF the next time Git touches it
```

... which means git has found some CRLFs in the working tree, and will delete them.

The same warning appeared in CI if I added a `git diff` on top of `no-unstaged-files.sh` - which means the underlying OS or IDE or whatever on my machine is not responsible for this (hence my assumption about node_modules).

To fix this, I added the following on top of the `no-unstaged-files.sh` : 

```
git add .
git reset
```

This makes git "touch" the files at least once, and cleans up line-endings.

The `git reset` is there such as the subsequent `git status` shows files as _unstaged_, and does not confuse whoever would have forgotten to stage them.

cchanche:
Unsure about this being the same problem we have with `@vercel/ncc` on master. Still need to investigate. The actual issue I was having may have been related to the moduleIDs [as stated here](https://github.com/release-drafter/release-drafter/pull/1478#issuecomment-3768966037). This would be an easy fix

If it is the same problem, I might just drop this branch (safer)

cchanche:
Turns-out : 
- rollup does not solve the CRLF problems we have - vercel's `ncc` might not be to blame here
- my fix with the `git add && git reset` was actually already implemented [here](https://github.com/release-drafter/release-drafter/commit/1b86f47b595f348972735dfb4002d375aa60981d), and I simply forgot I had removed it when building rollup configs
- the issue I was originally having was caused by my local `node_modules` folder being full of unwanted dependencies after doing (too much) `checkout`s from one branch to another. The bundler (webpack for `ncc`, or `rollup`) would add some primitives in the final bundle that were not added in CI.

This branch is probably ready to be merged, but does not fix or improve anything IMHO. I'll just close and hope #1476 comes to life quickly with new bundler.

Issue #1479, opened 2026-01-19
build: restore publishing to npmj ?


cchanche:
I think [our NPMJS_TOKEN was revoked](https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/)

[I disabled publishing for now](https://github.com/release-drafter/release-drafter/commit/60cac2b845b56526905acdb9de6d6c6cb4e199f5), and thought maybe since the app is no longer supported, the corresponding npmjs repo could be discontinued ? Does it serve any other purpose (or any at all) ? LMK @jetersen

I you'd like to publish to npmjs, I can take over tokens and stuff if you or @TimonVS or @toolmantim would prefer. They changed how it works I would need to look into it.

jetersen:
We should just disable tokens and use OIDC auth.

Which I have done. The action should be updated to the following:
https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow

```yaml
permissions:
  id-token: write  # Required for OIDC

...

      - name: yarn publish
        run: yarn publish --access public
```

reduced for brevity 

<img width="777" height="591" alt="Image" src="https://github.com/user-attachments/assets/cb15801c-6069-4c13-9d18-59b182d8c734" />

jetersen:
not sure if yarn supports oidc but it uses npm underneath to my undestanding so hopefully. Otherwise pnpm seems to have support so 😅 

Ended up updating the action https://github.com/release-drafter/release-drafter/commit/9928b9297cb7677b5b52f905a4b1e5759e2960b0

cchanche:
Ah very nice ! Thank you for this

cchanche:
I refactored the `release.yml` file to a manually-triggered workflow for the sake of publishing v6.1.1 and testing oidc. (see [commit](https://github.com/release-drafter/release-drafter/commit/af10e976934a6312e5fe7889fb2034cd669f10c6))

The [corresponding workflow](https://github.com/release-drafter/release-drafter/actions/runs/21202101921/job/60989822674) run failed :(

```log
error Couldn't publish package: "https://registry.npmjs.org/release-drafter-github-app: Not found"
info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.
Error: Process completed with exit code 1.
```

Plan is to rewrite the workflow back to the original once v6.1.1 was published to npmjs. The original `release.yml` is only triggered on tag pushes.

[...]

PR #1480, opened 2026-01-21
build: support OIDC for npmjs publishing

cchanche:
closes: #1479 

NPMJS requires OIDC authentication now.

migrated repo to vanilla `npm` using [`synp`](https://www.npmjs.com/package/synp)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,9 +2,8 @@ name: Release
 
 on:
   push:
-    branches:
-      - master
-  pull_request:
+    tags:
+      - v*.*.*
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


cchanche:
Build output is exactly the same as before, that's a good sign 👍

cchanche:
Drafted because the current workflows cannot be merged as-is

cchanche:
Alright so support for oidc was introduced in [`npm@v11.5.0`](https://github.com/npm/cli/releases/tag/v11.5.0).

`setup-node` in ci installs the following : 

```
Environment details
  node: v20.19.6
  npm: 10.8.2
  yarn: 1.22.22
```

I'll look into bumping `npm` (and enforcing it in CI at least)

cchanche:
Alright, everything looks good ! The current release workflow published to npmjs like it's nothing. [Here is the logs](https://github.com/release-drafter/release-drafter/actions/runs/21206086642/job/61002811960?pr=1480).

I'll revert this workflow to trigger only on tag pushes now

.github/actions/common-setup/action.yml
```diff
@@ -0,0 +1,19 @@
...
+      with:
+        node-version: 20
+
+    - name: Install npm
+      shell: bash
+      run: npm install -g npm@11.5.2
+
+    - name: Install dependencies
+      shell: bash
+      run: npm ci
```
jetersen, line 19: Just bump the node version to 24 the active LTS? v20 is going to be EOL in april.

jetersen: Also this :sweat:
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

cchanche: Yeah i have though of this, discussed it in #1465 already, and thought this might be a breaking change like [v6.0.0](https://github.com/release-drafter/release-drafter/releases/tag/v6.0.0) was... though I am unsure I see what's really breaking

cchanche: Ah, [found it](https://github.com/release-drafter/release-drafter/pull/1379#issuecomment-1928009747).

I'll take a quick look at depreciation notices from github about node 20 and see if we need a breaking change or not. keeping this open in the meantime

jetersen: GitHub actions recommends bumping major version on nodejs version change in the action.yml it is barried somewhere in one of the https://github.com/actions repo

https://github.com/release-drafter/release-drafter/blob/9928b9297cb7677b5b52f905a4b1e5759e2960b0/action.yml#L4

cchanche: Yes, altering action.yml is defenitely the way to go.
Let's discuss this in #1565, I think it is best to drop node 20 quickly

Issue #1509, opened 2026-02-12
when running a job with `prerelease: true` and `publish: true`, after running a job with `prerelease: false` and `publish: false`, the draft release is removed by the prerelease job


bcheidemann:
# Reproduction

```yml
name: Release
'on':
  push:
    branches:
      - main
  pull_request:
    types:
      - opened
      - reopened
      - synchronize
permissions:
  contents: read
jobs:
  UpdateReleaseDraft:
    runs-on: ubuntu-latest
    timeout-minutes: 20
    permissions:
      contents: write
    steps:
      - name: Draft next release
        uses: release-drafter/release-drafter@v5
        with:
          commitish: main
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  PublishBetaRelease:
    runs-on: ubuntu-latest
    timeout-minutes: 20
    if: github.event_name == 'push'
    permissions:
      contents: write
    # Added for reproducability - ideally the jobs would run concurrently but this leads to unpredictable behavior depending on the order in which the finish.
    needs:
      - UpdateReleaseDraft
    steps:
      - name: Publish beta release
        uses: release-drafter/release-drafter@v5
        with:
          commitish: main
          prerelease: true
          prerelease-identifier: beta
          publish: true
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

# Expected Behavior

A draft release (e.g. 1.2.3) should be created
A prerelease (e.g. 1.2.3-beta.0) should be created

# Actual Behavior

The draft release appears to be removed by the `PublishBetaRelease` step
The prerelease is created

cchanche:
Hmm, I am unsure your expected behavior is the intended one.

Here is what (I believe) is happening here : 

Your pull request events draft (and update a draft-) release, until someone pushes on main.

When a push on main happens, draft is updated one last time, and published right after. 

(I am unsure `prerelease: true` has anything to do with your actual issue, unless you tested without it and behavior is different.)

What you refer as

> The draft release appears to be removed

... is but `PublishBetaRelease` applying it's `publish: true`, meaning publishing whatever the run deemed as the last release-draft - which corresponds to whatever you drafted in `UpdateReleaseDraft`.

Now, I am pretty sure there is a solution in release-drafter for your particular workflow to work, and I think it may involve using the `include-pre-releases` param correctly.

[...]

PR #1515, opened 2026-02-25
feat: deprecate include-pre-releases in favor of prerelease

### LINKED ISSUE #1425: Release drafter v6.1.0 does not find existing draft releases and thus creates multiple draft releases
Just updated to release drafter v6.1.0 from  v6.0.0 and noticed that it started to create multiple draft releases. So if I already had a draft release titled as "Release 1.1.0", merging a new PR to main branch triggered release drafter which should update this "Release 1.1.0" draft release. However, with v6.1.0, release drafter can not find this existing draft release and thus it creates another draft release with the same name  "Release 1.1.0". This ends up to having several  "Release 1.1.0"  draft releases in the repository rather than what would be expected is to have just one  "Release 1.1.0" draft release which release drafter should update until an actual non-draft release is made. This issue happens with multiple repositories. I am using GitHub Enterprise Server 3.13.8.

Here is example of my release drafter GHA step:
```
  - name: Run release drafter
	id: release-drafter
	uses: release-drafter/release-drafter@v6
	env:
	  GITHUB_TOKEN: ${{ secrets.token }}
	with:
	  disable-releaser: false
	  disable-autolabeler: true
```

GHA job output when using release-drafter v6.0.0:
```
my-org/my-repo: Found 125 releases
{ name: 'event', id: '3494448' }
my-org/my-repo: Draft release: untagged-b37987ae3db4d75cab9d
{ name: 'event', id: '3494448' }
my-org/my-repo: Last release (including prerelease): v1.19.7
{ name: 'event', id: '3494448' }
my-org/my-repo: Fetching parent commits of refs/heads/master since 2025-01-21T09:17:22Z
{ name: 'event', id: '3494448' }
my-org/my-repo: Updating existing release
{ name: 'event', id: '3494448' }        
```
So note that how it found the `Draft release: untagged-b37987ae3db4d75cab9d` and thus it updates existing draft release.

GHA job output when using release-drafter v6.1.0:
```
my-org/my-repo: Found 125 releases
{ name: 'event', id: '3494235' }
my-org/my-repo: No draft release found
{ name: 'event', id: '3494235' }
my-org/my-repo: Last release (including prerelease): v1.19.7
{ name: 'event', id: '3494235' }
my-org/my-repo: Fetching parent commits of refs/heads/master since 2025-01-21T09:17:22Z
{ name: 'event', id: '3494235' }
my-org/my-repo: Creating new release
{ name: 'event', id: '3494235' }
```

So note that how it did not find any draft release even though there for sure was a draft release and thus it creates new draft release.

So to me it seems that something is broken in the release-drafter v6.1.0. It just does not seem to be able to find an existing draft release. I would guess that it is this one which have broken this: https://github.com/release-drafter/release-drafter/pull/1385

christopherklint97:
## Summary

Fixes #1425

The `findReleases` function filtered draft releases using `r.prerelease === includePreReleases`, which is incorrect because `include-pre-releases` controls whether prereleases count when searching for the **last release** — not whether only prerelease drafts should be matched.

This caused users with `include-pre-releases: true` in their config to never find their existing non-prerelease drafts (`false === true` is always `false`), creating duplicate draft releases on every push.

### Fix

Use the same filter pattern already used for last release detection on the [line above](https://github.com/release-drafter/release-drafter/blob/master/lib/releases.js#L67):

```diff
  const draftRelease = filteredReleases.find(
-   (r) => r.draft && r.prerelease === includePreReleases
+   (r) => r.draft && (!r.prerelease || includePreReleases)
  )
```

This gives the expected semantics as described in [this comment](https://github.com/release-drafter/release-drafter/issues/1425#issuecomment-3804985110):

- `includePreReleases = false` → only match non-prerelease drafts
- `includePreReleases = true` → match any draft (prerelease or not)

No changes to `index.js` — the fix is entirely within `lib/releases.js`.

### Test plan

- [x] All existing tests pass (174/174)
- [x] New test: `includePreReleases: false` with only a prerelease draft → returns undefined (correctly skipped)
- [x] New test: `includePreReleases: true` with non-prerelease draft → finds the draft (the bug scenario)
- [x] New test: `includePreReleases: true` with both draft types → finds the first matching draft
- [x] New test: `includePreReleases` not provided → finds non-prerelease draft (default behavior)
- [x] Prettier, ESLint, and `ncc build` all pass

### DIFF
--- a/lib/commits.js
+++ b/lib/commits.js
@@ -129,13 +129,29 @@ const findCommitsWithAssociatedPullRequests = async ({
     : config['initial-commits-since']
   const validationResult = Joi.date().iso().validate(since)
 
-  core.debug(' since value: ' + since)
-  core.debug(
-    ' since value validation.value: ' +
-      validationResult.value +
-      ' error: ' +
-      validationResult.error
-  )
+  log({
+    context,
+    message: `since value: ${since}`,
+    debug: true,
+  })
+  log({
+    context,
+    message: `since value validation.value: ${JSON.stringify(
+      validationResult,
+      null,
+      2
+    )}`,
+    debug: true,
+  })
+  log({
+    context,
+    message: `since value validation.error: ${JSON.stringify(
+      validationResult.error,
+      null,
+      2
+    )}`,
+    debug: true,
+  })
 
   // The validation result contains either an error or the validated value
   if (validationResult.error) {
--- a/lib/log.js
+++ b/lib/log.js
@@ -1,9 +1,13 @@
-const log = ({ context, message, error }) => {
+const log = ({ context, message, error, debug }) => {
   const repo = context.payload.repository
   const prefix = repo ? `${repo.full_name}: ` : ''
   const logString = `${prefix}${message}`
   if (error) {
     context.log.warn(error, logString)
+  } else if (debug) {
+    typeof debug === 'object'
+      ? context.log.debug({ ...debug }, logString)
+      : context.log.debug({}, logString)
   } else {
     context.log.info(logString)
   }
--- a/lib/releases.js
+++ b/lib/releases.js
@@ -1,6 +1,5 @@
 const compareVersions = require('compare-versions')
 const regexEscape = require('escape-string-regexp')
-const core = require('@actions/core')
 
 const { getVersionInfo } = require('./versions')
 const { template } = require('./template')
@@ -342,7 +341,8 @@ const generateChangeLog = (mergedPullRequests, config) => {
 const resolveVersionKeyIncrement = (
   mergedPullRequests,
   config,
-  isPreRelease
+  isPreRelease,
+  context
 ) => {
   const priorityMap = {
     patch: 1,
@@ -358,7 +358,11 @@ const resolveVersionKeyIncrement = (
       .flat()
   )
 
-  core.debug('labelToKeyMap: ' + JSON.stringify(labelToKeyMap))
+  log({
+    context,
+    message: `labelToKeyMap`,
+    debug: labelToKeyMap,
+  })
 
   const keys = mergedPullRequests
     .filter(getFilterExcludedPullRequests(config['exclude-labels']))
@@ -366,7 +370,11 @@ const resolveVersionKeyIncrement = (
     .flatMap((pr) => pr.labels.nodes.map((node) => labelToKeyMap[node.name]))
     .filter(Boolean)
 
-  core.debug('keys: ' + JSON.stringify(keys))
+  log({
+    context,
+    message: `keys`,
+    debug: keys,
+  })
 
   const keyPriorities = keys.map((key) => priorityMap[key])
   const priority = Math.max(...keyPriorities)
@@ -374,7 +382,11 @@ const resolveVersionKeyIncrement = (
     (key) => priorityMap[key] === priority
   )
 
-  core.debug('versionKey: ' + versionKey)
+  log({
+    context,
+    message: `versionKey`,
+    debug: versionKey,
+  })
 
   const versionKeyIncrement = versionKey || config['version-resolver'].default
 
@@ -424,10 +436,15 @@ const generateReleaseInfo = ({
   const versionKeyIncrement = resolveVersionKeyIncrement(
     mergedPullRequests,
     config,
-    isPreRelease
+    isPreRelease,
+    context
   )
 
-  core.debug('versionKeyIncrement: ' + versionKeyIncrement)
+  log({
+    context,
+    message: `versionKeyIncrement`,
+    debug: versionKeyIncrement,
+  })
 
   const versionInfo = getVersionInfo(
     lastRelease,
@@ -440,7 +457,11 @@ const generateReleaseInfo = ({
     config['prerelease-identifier']
   )
 
-  core.debug('versionInfo: ' + JSON.stringify(versionInfo, null, 2))
+  log({
+    context,
+    message: `versionInfo`,
+    debug: versionInfo,
+  })
 
   if (versionInfo) {
     body = template(body, versionInfo)
@@ -452,7 +473,11 @@ const generateReleaseInfo = ({
     tag = template(tag, versionInfo)
   }
 
-  core.debug('tag: ' + tag)
+  log({
+    context,
+    message: `tag: ${tag}`,
+    debug: true,
+  })
 
   if (name === undefined) {
     name = versionInfo
@@ -462,7 +487,11 @@ const generateReleaseInfo = ({
     name = template(name, versionInfo)
   }
 
-  core.debug('name: ' + name)
+  log({
+    context,
+    message: `name: ${name}`,
+    debug: true,
+  })
 
   // Tags are not supported as `target_commitish` by Github API.
   // GITHUB_REF or the ref from webhook start with `refs/tags/`, so we handle
--- a/package-lock.json
+++ b/package-lock.json
@@ -3823,6 +3823,21 @@
       "resolved": "https://registry.yarnpkg.com/fs.realpath/-/fs.realpath-1.0.0.tgz",
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8= sha512-OO0pH2lK6a0hZnAdau5ItzHPI6pUlvI7jMVnxUQRtw4owF2wk8lOSabtGDCTP4Ggrg2MbGnWO9X8K1t4+fGMDw=="
     },
+    "node_modules/fsevents": {
+      "version": "2.3.3",
+      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.3.tgz",
+      "integrity": "sha512-5xoDfX+fL7faATnagmWPpbFtwh/R77WmMMqqHGS65C3vvB0YHrgF+B1YmZ3441tMj5n63k0212XNoJwzlhffQw==",
+      "dev": true,
+      "hasInstallScript": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": "^8.16.0 || ^10.6.0 || >=11.0.0"
+      }
+    },
     "node_modules/function-bind": {
       "version": "1.1.2",
       "resolved": "https://registry.yarnpkg.com/function-bind/-/function-bind-1.1.2.tgz",


mmaaras (approved):

test/releases.test.js
```diff
@@ -367,9 +410,9 @@ describe('releases', () => {
       })
 
       expect(draftRelease).toEqual({
-        tag_name: 'v1.0.2-rc.1',
+        tag_name: 'v1.0.0',
         draft: true,
-        prerelease: true,
+        prerelease: false,
```
cchanche: Suite mentions "returning prerelease draft", but it expects a release with `{ prerelease: false }`

cchanche (changes requested):
Thank you for contributing @christopherklint97  !

This is indeed more in line whith what the name of the input suggests, but as mentioned [here](https://github.com/release-drafter/release-drafter/issues/1425#issuecomment-3804985110), the real problem was not what `include-pre-release` does, but rather how it is named.

Therefore, I'd like to know more about your particular workflow @christopherklint97 !

I'd like to be proven wrong, but the way I see prerelease workflows (see: #1509), there should be no need a hybrid _fetch releases AND prepreleases_.

If there is - which is strongly suggested by the name of the input and by everything people did with it pre-[`v6.1.0`](https://github.com/release-drafter/release-drafter/releases/tag/v6.1.0) (see #1385) - I would suggest adding both options instead of switching back-and forth from one mode to the other. This PR is essentially reverting #1385 I believe.

For instance :

```yaml
filter-by-prerelease: 'only' | 'include' | 'exclude' # defaults to 'exclude'
include-pre-release: boolean # legacy support - enables filter-by-prereleas's 'include' mode. Default : 'exclude' (false)
```

christopherklint97:
> Thank you for contributing @christopherklint97 !
> 
> This is indeed more in line with what the name of the input suggests, but as mentioned [here](https://github.com/release-drafter/release-drafter/issues/1425#issuecomment-3804985110), the real problem was not what `include-pre-release` does, but rather how it is named.
> 
> Therefore, I'd like to know more about your particular workflow @christopherklint97 !
> 
> I'd like to be proven wrong, but the way I see prerelease workflows (see: #1509), there should be no need for a hybrid _fetch releases AND prepreleases_.
> 
> If there is - which is strongly suggested by the name of the input and by everything people did with it pre-[`v6.1.0`](https://github.com/release-drafter/release-drafter/releases/tag/v6.1.0) (see #1385) - I would suggest adding both options instead of switching back-and forth from one mode to the other (this PR reverts #1385).
> 
> For instance :
> 
> ```yaml
> filter-by-prerelease: 'only' | 'include' | 'exclude' # defaults to 'exclude'
> include-pre-release: boolean # legacy support - enables filter-by-prereleas's 'include' mode. Default : 'exclude' (false)
> ```

Thanks for the thorough review @cchanche!

Good catch on the test name mismatch, went ahead and fixed that one.  😅 

After digging deeper into the history that you mentioned (#1385, #1425) I believe the root issue is that `include-pre-releases` was being used for two different things:

1. **Which published releases to scan** when finding `lastRelease` (the changelog baseline)
2. **Which draft to update** when finding `draftRelease`

These feel like separate concerns. The scan question is about *reading history*, but the draft question is really "find the draft that matches what I'm *about to create*", which is driven by `config.prerelease`, not by how we scan past releases.

So instead of introducing a new `filter-by-prerelease` enum right now, I've split the two concerns:

- **`lastRelease` scan** — still uses `includePreReleases` (unchanged)
- **`draftRelease` matching** — now uses a new `isPreRelease` parameter (from `config.prerelease`) with a fallback:
  1. First, find a draft whose `prerelease` status matches what we're creating
  2. Fall back to a non-prerelease draft if no match exists (handles the first prerelease run)

I believe this solves the original #1385 problem (don't overwrite the wrong draft when both types exist) without the regression from #1425, and without changing the `include-pre-releases` contract.

If there's a real need for an `'only'` scan mode, that could be added later as a separate `filter-by-prerelease` option but I think it should only affect the `lastRelease` scan, not draft matching.

Happy to hear your thoughts on this approach!

cchanche:
Ah very nice analysis @christopherklint97, thank you !

You are correct, I made the (quick) assumption that draft and non-draft past releases were fetched using the same `r.prerelease === includePreReleases` condition. I should have paid more attention to the code.

Thinking about it, the fact that past (non-draft) releases are filtered using `(!r.prerelease || includePreReleases)` is perfectly fine. `only` scan mode (as you put it nicely) can't seem to fit any use case I can think of now that I see the details. If there is a full release in your history of published releases, you will most probably want release drafter to see it, no matter whether you used `include-pre-releases` and/or `prerelease`. I was wrong.

I also like the idea of bringing the `prerelease` config into the mix. I like to think this could actually be the only config there is about this, such as you could simply infer "hey, let me do prerelease stuff now", and that's it. The thing behaves as expected.

Truth however, is that everyone does not always have the exact same expectations out of release drafter doing prerelease stuff... That's why I would have liked to know more about your exact workflow ! I am not a prerelease user myself, things get pretty confusing sometimes...

So now, about what we should do with how we find draft releases... As an output of `findPreviousReleases()`, it's role is only to help us decide whether to update (the returned draft) or create a new one (that may also be published right away, thus not ending as a draft - but `publish: true` is out of scope here).

Current capability of `includePreReleases` does two things :

1. it interferes with how we fetch non-draft past releases
2. it is in _exclusive_ mode - either you do or you don't see drafts, but not both

Your change here suggests we stop using `includePreReleases` here, which de-facto eliminates 1., and introduces `prerelease: true/false` as the source of how we see past draft prereleases - and alters 2. : 

```ts
  const draftRelease =
    filteredReleases.find(
      (r) => r.draft && r.prerelease === Boolean(isPreRelease)
    ) || filteredReleases.find((r) => r.draft && !r.prerelease)
```

There are two things happening at the same time here : 

1. An assumption that users that want release-drafer to upsert a prerelease (draft or not) will want that to happen by updating their past prerelease draft
2. Another assumption, that assumes converting a previous draft (full) release to a prerelease if nothing was found during 1.

I think 2. may not be to everyone's liking... I can see how people would maintain two drafts at the same time. One for their full release `v3.5.0`, and one for a candidate for the same full release : `v3.5.0-rc.2`. Let's say you publish (manually or not) `v3.5.0-rc.2` because you want your users to start testing whatever is in `v3.5.0-rc.2`. Now your only draft release is `v3.5.0`. At that point, if you want to draft `v3.5.0-rc.3` ( draft prerelease) with changes since `v3.5.0-rc.2`, assumption 2. will end up converting `v3.5.0` into this `v3.5.0-rc.3` you want instead of simply creating it, and going back to having two drafts at the same time.

Am I missing something ? Is this what you need for your workflow ?

It looks to me like this matches pretty closely your new test suite : 

`it(should fall back to non-prerelease draft when isPreRelease is true but no prerelease draft exists)`


For my part, I think users should be able to have both drafts at the same time, and code may need to simply look like :

```ts
  const draftRelease =
    filteredReleases.find(
      (r) => r.draft && r.prerelease === Boolean(isPreRelease)
    )
```

cchanche:
Now, about this : 

> I also like the idea of bringing the `prerelease` config into the mix. I like to think this could actually be the only config there is about this, such as you could simply infer "hey, let me do prerelease stuff now", and that's it. The thing behaves as expected.

What do you think about the following : 

```ts
  const sortedSelectedReleases = sortReleases({
    releases: filteredReleases.filter(
      (r) => !r.draft && (!r.prerelease || Boolean(isPreRelease))
    ),
    tagPrefix
  })
  const draftRelease =
    filteredReleases.find(
      (r) => r.draft && r.prerelease === Boolean(isPreRelease)
    )
```

... which lets us drop `include-pre-releases` entirely.

I can't get to see how deprecating this would :

- drop the flexibility anyone would need.
- not be waaay easier for people, using only `prerelease: true/false` to do everything

What are use-cases for having `prerelease` different from `include-pre-releases` ? (assuming it is implemented as I suggest in my previous comment)

christopherklint97:
Great points on both counts @cchanche, appreciate the different edge cases to cover.

**On the fallback concern** I agree. The dual-draft scenario you described (`v3.5.0` + `v3.5.0-rc.2`, publish rc.2, then the fallback hijacks `v3.5.0`) is a real problem. My original thinking was that the fallback would help on the *first* prerelease run when no prerelease draft exists yet, but that's not worth the risk of silently converting someone's full release draft. Users in that situation would just get a fresh draft created, which is the correct behavior anyway.

What do you think about dropping the fallback to:

```ts
const draftRelease = filteredReleases.find(
  (r) => r.draft && r.prerelease === Boolean(isPreRelease)
)
```

I believe this is clean and predictable so each draft type stays in its own lane. I would also update the fallback test (`should fall back to non-prerelease draft when isPreRelease is true but no prerelease draft exists`) to expect `undefined` instead, since the correct behavior is now to create a new draft.

---

**On dropping `include-pre-releases` entirely**, the more I think about it it feels like these two concerns align:

- **Creating a prerelease** (`prerelease: true`) → you want prereleases in the `lastRelease` scan so the changelog shows RC-to-RC changes (e.g. what changed since `v1.0.0-rc.1`)
- **Creating a stable release** (`prerelease: false`) → you want only stable releases in the scan so the changelog shows the full picture for users upgrading from the last stable (e.g. everything since `v0.9.0`)

I initially thought there'd be a case for `prerelease: false` + `include-pre-releases: true` (baseline on the latest RC when building a stable release), but that actually produces a weird changelog where users upgrading from `v0.9.0` to `v1.0.0` would only see the tiny delta since rc.2, not the full set of changes.

So `prerelease` really does drive both sides naturally. Your proposed code:

```ts
const sortedSelectedReleases = sortReleases({
  releases: filteredReleases.filter(
    (r) => !r.draft && (!r.prerelease || Boolean(isPreRelease))
  ),
  tagPrefix,
})
const draftRelease = filteredReleases.find(
  (r) => r.draft && r.prerelease === Boolean(isPreRelease)
)
```

I'm on board with this. The only thing to be mindful of is backwards compatibility for users currently relying on `include-pre-releases`. I think we could either:

1. Keep `include-pre-releases` as a supported input but mark it deprecated, and derive the scan behavior from `isPreRelease` by default with a fallback to `includePreReleases` when explicitly set
2. Do a clean break and just switch to `isPreRelease` — `prerelease-identifier` users already imply `prerelease: true` in practice so they wouldn't be affected

Happy to go either way so what's your preference on the migration path?

cchanche:
Damn, you are way better at explaining how this works than I'll ever be 😅 

I am 100% onboard with everything you said.

About migration, I believe current `v6` could deprecate `include-pre-releases` without breaking anything, and be entirely removed for `v7` (in #1475)

LMK if you want me to finalize things in this current PR for `v6` !

christopherklint97:
Haha thanks it took me awhile to wrap my head around what they actual use cases are 😅 

Sounds good, you can go ahead and finalize! Thanks for the help 🙌🏾

cchanche (approved):

jtraglia:
> I initially thought there'd be a case for `prerelease: false` + `include-pre-releases: true` (baseline on the latest RC when building a stable release), but that actually produces a weird changelog where users upgrading from `v0.9.0` to `v1.0.0` would only see the tiny delta since rc.2, not the full set of changes.

@cchanche Could there be an option to keep the current functionality? Our repo would be affected by this change and I'm not sure how we'd deal with it with a single `prerelease` option. We use pre-releases to signal that the latest specifications are under development. I'd prefer if the stable release didn't contain a duplicate list of contributions.

* https://github.com/ethereum/consensus-specs/releases

cchanche:
Ah thank you for reaching-out @jtraglia 

I guess you already tried `v6.3.0` then, and that your desired workflow prevents you from upgrading.

I'll immediately take a look at your repo

jtraglia:
Thanks! And we haven't tried making a release with v6.3.0 yet. We're just updating dependencies. I noticed that it would cause issues when reading the release notes. Nothing is urgent.

* https://github.com/ethereum/consensus-specs/pull/4983

cchanche:
So, to be clear, what has been removed is the two following cases :

1. `prerelease: false + include-pre-releases: true`
2. `prerelease: true + include-pre-releases: false`

In `v6.2.0` :

1. meant : upsert a stable release (draft), but when looking for : 
  - a previous published release, convert prereleases to stable releases
  - a previous draft, look only for prereleases
  
  This means the upsert draft will be stable, with :
    - changes since the last prerelease, or stable release if none were found
    - converts the last prerelease draft to a stable draft (the upserted one)
  
  Both of the above bullet points looked like no one would want to us.
  
2. meant : upsert a prerelease (draft), but when looking for :
  - a previous published release, ignore prereleases
  - a previous draft, look only for stable drafts
  
  This means the upsert draft will be a prerelease, with :
    - changes since the last stable release
    - converts the last stable draft to a prerelease draft (the upserted one)
  
  Both of the above bullet points looked like no one would want to us.
  

@jtraglia, does any of the 1. and 2. workflow above look like something you would want to do ?

In https://github.com/ethereum/consensus-specs/releases, it does indeed look like release `v1.6.0` only includes changes since the last prerelease (instead of intended stable release). The `release-drafter@6.2.0` workflow that would have created `v1.6.0` could have used case 1. `prerelease: false + include-pre-releases: true` as far as I understand.

If so, it would incorrectly have identified a potential previous stable draft for `v1.6.0` had it existed beforehand

> 1. meant : upsert a stable release (draft), but when looking for a previous draft, look only for prereleases


So what you need IMO, is something to create a stable release (draft or not) with changes since the last prerelease. Am I correct ? This indeed is not in line with what we've done here for `v6.3.0`. See [Prerelease Workflow](https://github.com/release-drafter/release-drafter?tab=readme-ov-file#prerelease-workflow) : 

> `update_full_release_draft` will pile-up changes since v3.5.0 inside a draft for v3.5.1 (or v3.6.0 or v4.0.0, depending on your config)

jtraglia:
> @jtraglia, does any of the 1. and 2. workflow above look like something you would want to do ?

Yes, we would like to do case 1 (`prerelease: false + include-pre-releases: true`).

This would apply to us: "a previous draft, look only for prereleases." We create the draft [here](https://github.com/ethereum/consensus-specs/blob/c72444226ea61de8dcf992fc88fd93cce9742515/.github/workflows/release.yml#L38), so that we can review the notes while the reference tests are being generated. Then we [publish](https://github.com/ethereum/consensus-specs/blob/c72444226ea61de8dcf992fc88fd93cce9742515/.github/workflows/release.yml#L102) the release here.

(Sorry, I've edited this ^^^ & fixed the links)

> So what you need IMO, is something to create a stable release (draft or not) with changes since the last prerelease. Am I correct ? This indeed is not in line with what we've done here for `v6.3.0`. See [Prerelease Workflow](https://github.com/release-drafter/release-drafter?tab=readme-ov-file#prerelease-workflow) :

Yes, this is correct. I understand that this is not the standard way of doing release notes, but it was a feature that we liked having. If the `include-pre-releases` feature is removed, we would need to either (1) not upgrade to v7 or (2) change how we do releases so that all of our releases are stable releases.

cchanche:
Let me break down what you need to accommodate your workflow (and, to be clear, probably other's too). Critical release-drafter mechanisms are :

- **`[scan published]`** looking for previous published releases (marks the commit we scan change history up-to)
- **`[scan drafts]`** looking for previous draft releases (selects the (potential) draft-release to update)
- **`[stable ?]`** whether the updated/created draft release (that may alwo be published right away) is a preleases or not


In your workflow, you need :

- for publishing stable `v1.6.0` :
  - **`[scan published]`** include prereleases, stable releases may also be scanned _- you want changes since last prerelease if it exists_
  - **`[scan drafts]`** include only stable releases _- you want to update previous `v1.6.0` draft if it exists (not based on it's version name)_
  - **`[stable ?]`** yes

- for publishing prerelease `v1.6.0-alpha.2` :
  - **`[scan published]`** include prereleases, stable releases may also be scanned _- you want changes since last prerelease if it exists_
  - **`[scan drafts]`** include only prereleases _- you want to update previous `v1.6.0-alpha.2` draft if it exists (not based on it's version name)_
  - **`[stable ?]`** no


In `v6.3.0`, `prerelease: true` accommodates your prerelease publishing workflow, but not your stable workflow : 

- for publishing stable `v1.6.0` (with `v6.3.0`'s `prerelease: false`) :
  - ❌**`[scan published]`** ~~include prereleases~~, stable releases may also be scanned _- you want changes since last prerelease if it exists_
  - ✅**`[scan drafts]`** include only stable releases - you want to update previous `v1.6.0` draft if it exists (not based on it's version name)
  - ✅**`[stable ?]`** yes

To fix this, you need to make use of deprecated `include-pre-releases: true` (available in `v6.3.0`), which enables scanning for published  prereleases. Problem here being it is deprecated 😅 

Am I correct @jtraglia ?

jtraglia:
> In `v6.3.0`, `prerelease: true` accommodates your prerelease publishing workflow, but not your stable workflow
> 

> Am I correct @jtraglia ?

Yes, this accurately describes the situation 😄

cchanche:
To clarify, I only took note of your [reply above](https://github.com/release-drafter/release-drafter/pull/1515#issuecomment-4013262138) after publishing my long breakdown above.

Thank you for sharing the exact workflow files, I'll take another look at them to confirm what you need !

> I understand that this is not the standard way of doing release notes

> it was a feature that we liked having

Let's keep the feature then ! I'll take a night's sleep on how to properly revert things, and also how a next major (breaking) version could clarify things.

You can safely consider `include-pre-releases`'s functionality is not deprecated anymore, and `v6.3.0` should be usable without any modification needed. I'll come back here with news when new version comes up 😉

cchanche:
@jtraglia as promised, just released [`v6.4.0`](https://github.com/release-drafter/release-drafter/releases/tag/v6.4.0) should stabilize things for you

jtraglia:
> @jtraglia as promised, just released [`v6.4.0`](https://github.com/release-drafter/release-drafter/releases/tag/v6.4.0) should stabilize things for you

@cchanche thank you! 🙂

PR #1517, opened 2026-02-26
feat: support advanced substitutions in replacers

### LINKED ISSUE #1363: Allow `replacers` to uppercase/lowercase capturing groups
My contributors do not always format their pull request titles appropriately, so I want to implement a replacer that always uppercases the first letter of the changelog entry. I tried this:

```yaml
replacers:
  # Uppercase the first letter of each changelog entry
  - search: '/- ([a-z])/g'
    replace: '- \u$1'
```

The `\u` works for some regex replacers, like in VSCode. However, this does not seem to work. Is there a way I can achieve this? If not, consider this a feature request!

cchanche:
fix #1363

Support more substitution tokens.

For examples (and reference), go to [regex101.com](https://regex101.com/) : 
- select "Function" > "Substitution" in the "Regex Editor" menu (default opened menu)
- In the "Quick Reference" on the right panel, look for "Substitution" at the bottom

Now switch between flavors in the "Regex Editor" menu, and notice how some implementations have more substitution tokens available than native "ECMAScript"

This PR introduces a few of them, based on those supported by VSCode's search & replace feature.

To be fair, Codex did the hard part of porting VSCode's sources.


jetersen:
think your easiest path forward is either drop all commits and cherry pick your last commit or start over :dagger:

cchanche:
> think your easiest path forward is either drop all commits and cherry pick your last commit or start over 🗡️

Yeah I was just in the process of doing that for other PRs and branches I need... rebasing hurts indeed

cchanche:
To be clear : I literally copied VSCode's implementation. Have not put much thoughts to it other than looking for a library before submitting to shameful copy/paste 😅  Open to suggestions if there's a cleaner way - especially if we want to clean-up VSCode's code I have not even looked at

src/tests/drafter/render-template.test.ts
```diff
@@ -5,7 +5,7 @@ describe('render template', () => {
   it('replaces $A with B', () => {
     const output = renderTemplate({ template: '$A', object: { $A: 'B' } })
 
-    expect(output).toEqual('B')
+    expect(output).toMatchInlineSnapshot(`"B"`)
```
jetersen, line 8: Seems sus that is it quoted in the test, but I guess that is match inline snapshot that's doing that.

cchanche: Yes that's bc snapshots can also be objects, booleans, numbers, arrays etc....

cchanche: I think I changed i t to snapshot because the corresponding suite may have failed during implementation, and I realized I could make my life easier if the `expect` was using snapshot

PR #1522, opened 2026-03-06
feat: Add support to excluded-paths

guimorg:
> Add exclude-paths so changelog can skip commits that touch certain paths

## What it does

New config option `exclude-paths`: list of path patterns (same format as include-paths). First drop any commit whose ID is in the excluded set, then, if `include-paths` is non-empty, keep only commits in the included set; otherwise keep all remaining commits. So exclude wins over include: if a path is in both, those commits are left out.

## Why

I have a case where I have a folder that is supposed to be included in one release drafter and be excluded from the main release drafter. With include-only you have to list every path you care about, which is brittle and annoying when the repo has many directories. Exclude-paths will help me to do exactly that, in one file I'll `include-paths: ['my-path']` and on the other one I'll `exclude-paths: ['my-path']`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ You can configure Release Drafter using the following key in your `.github/relea
 | `sort-direction`           | Optional | Sort changelog in ascending or descending order. Can be one of: `ascending`, `descending`. Default: `descending`.                                                                                         |
 | `prerelease`               | Optional | Whether to draft a prerelease, with changes since another prerelease (if applicable). Default `false`.                                                                                                    |
 | `prerelease-identifier`    | Optional | A string indicating an identifier (alpha, beta, rc, etc), to increment the prerelease version. This automatically enables `prerelease` if not already set to `true`. Default `''`.                        |
+| `include-pre-releases`     | Optional | When looking for the last published release to scan changes up-to, include pre-releases. Has no effect if using `prerelease: true` (already enabled). Default `false`.                                    |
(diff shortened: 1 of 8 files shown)


dist/index.js
```diff
@@ -190231,27 +190230,38 @@ const findCommitsWithAssociatedPullRequests = async ({
...
+      )
+
+      const nodes = _.get(pathData, [...dataPath, 'nodes'], [])
 
-    if (!anyChanges) {
-      // Short circuit to avoid blowing GraphQL budget
-      return { commits: [], pullRequests: [] }
+      for (const { id } of nodes) {
+        excludedIds.add(id)
+      }
```
guimorg, line 190281: I think I'll put this back here, much probably.

guimorg: There we go, re-included this bit. My bad from removing it.

README.md
cchanche, line 1: Aah, sorry to bother with that but keeping the formatting on this would be nice. 

Prettier is not configured for markdown (yet), but I'll simply apply what I have locally if you don't mind. Diff should be easier to read then

cchanche: Nevermind, checking-out your branch and formatting locally did nothing... Not sure why diff looks like this, but it's ok I've taken the time to locate your changes, this is perfect.

guimorg:
> Thank you for your contribution @guimorg, very helpful 😃
> 
> This looks very good to me, thank you for simplifying this bit where we actually exclude or include commits from `allCommits` inside the filter. Your approach is easier to maintain.
> 
> A very small detail about the README's formatting and we are good here !

Oh, I'm so happy this was approved and merge! Really happy and proud to contribute with the community! 💪

PR #1526, opened 2026-03-08
refactor: simplify vite build

jetersen:
## Summary

Cleanup PR on top of #1475.

- **Remove lodash**: replaced `isBoolean` with `typeof x === 'boolean'`, replaced `_.get`/`_.has`/`_.set`/`_.update` in `paginate-graphql` with small native helpers, replaced `_.uniqBy` with a `Set`-based filter (preserving first-occurrence semantics). Removes both `lodash` runtime dep and `@types/lodash` dev dep.
- **Update `@actions/core` to v3.0.0**
- **Add `tsx` dev dep**: allows running `src/scripts/json-schema.ts` directly without a prior build step — fixes the `schemas` npm script
- **Named zod imports**: switch from `import z from 'zod'` (default import) to named imports (`import { object, string, ... } from 'zod'`) and `import type * as z from 'zod'` for type-only usage, across all source and test files
- **Simplified bundle structure**: only 2 entry points (`src/actions/drafter/run.ts` and `src/actions/autolabeler/run.ts`); shared code goes into `dist/chunks/`. Previously every source file was its own output file.
- **Vite config cleanup**: removed redundant defaults (`outDir`, `emptyOutDir`, `sourcemap`, `isolate`, `environment`, `exclude`); inlined `external` as a predicate function; added `node:` prefix to all built-in imports in our source files
- **Fix `autolabeler/action.yml` main path**: was pointing to `dist/autolabeler/run.js` (non-existent), corrected to `../dist/actions/autolabeler/run.js`

## Test plan

- [x] `npm run build` produces `dist/actions/drafter/run.js`, `dist/actions/autolabeler/run.js`, and `dist/chunks/*.js`
- [x] `npm run test` passes (259/259)
- [x] `npm run schemas` runs without needing a prior `npm run build`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "prettier-eslint": "^16.4.2",
         "tsx": "^4.21.0",
         "typescript": "^5.9.3",
-        "vite": "^8.0.0-beta.16",
+        "vite": "^8.0.0-beta.18",
         "vite-tsconfig-paths": "^6.0.4",
         "vitest": "^4.0.18"
       },
@@ -2914,9 +2914,9 @@
       "license": "MIT"
     },
     "node_modules/@rolldown/binding-android-arm64": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-android-arm64/-/binding-android-arm64-1.0.0-rc.7.tgz",
-      "integrity": "sha512-/uadfNUaMLFFBGvcIOiq8NnlhvTZTjOyybJaJnhGxD0n9k5vZRJfTaitH5GHnbwmc6T2PC+ZpS1FQH+vXyS/UA==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-android-arm64/-/binding-android-arm64-1.0.0-rc.8.tgz",
+      "integrity": "sha512-5bcmMQDWEfWUq3m79Mcf/kbO6e5Jr6YjKSsA1RnpXR6k73hQ9z1B17+4h93jXpzHvS18p7bQHM1HN/fSd+9zog==",
       "cpu": [
         "arm64"
       ],
@@ -2931,9 +2931,9 @@
       }
     },
     "node_modules/@rolldown/binding-darwin-arm64": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-arm64/-/binding-darwin-arm64-1.0.0-rc.7.tgz",
-      "integrity": "sha512-zokYr1KgRn0hRA89dmgtPj/BmKp9DxgrfAJvOEFfXa8nfYWW2nmgiYIBGpSIAJrEg7Qc/Qznovy6xYwmKh0M8g==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-arm64/-/binding-darwin-arm64-1.0.0-rc.8.tgz",
+      "integrity": "sha512-dcHPd5N4g9w2iiPRJmAvO0fsIWzF2JPr9oSuTjxLL56qu+oML5aMbBMNwWbk58Mt3pc7vYs9CCScwLxdXPdRsg==",
       "cpu": [
         "arm64"
       ],
@@ -2948,9 +2948,9 @@
       }
     },
     "node_modules/@rolldown/binding-darwin-x64": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-x64/-/binding-darwin-x64-1.0.0-rc.7.tgz",
-      "integrity": "sha512-eZFjbmrapCBVgMmuLALH3pmQQQStHFuRhsFceJHk6KISW8CkI2e9OPLp9V4qXksrySQcD8XM8fpvGLs5l5C7LQ==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-darwin-x64/-/binding-darwin-x64-1.0.0-rc.8.tgz",
+      "integrity": "sha512-mw0VzDvoj8AuR761QwpdCFN0sc/jspuc7eRYJetpLWd+XyansUrH3C7IgNw6swBOgQT9zBHNKsVCjzpfGJlhUA==",
       "cpu": [
         "x64"
       ],
@@ -2965,9 +2965,9 @@
       }
     },
     "node_modules/@rolldown/binding-freebsd-x64": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-freebsd-x64/-/binding-freebsd-x64-1.0.0-rc.7.tgz",
-      "integrity": "sha512-xjMrh8Dmu2DNwdY6DZsrF6YPGeesc3PaTlkh8v9cqmkSCNeTxnhX3ErhVnuv1j3n8t2IuuhQIwM9eZDINNEt5Q==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-freebsd-x64/-/binding-freebsd-x64-1.0.0-rc.8.tgz",
+      "integrity": "sha512-xNrRa6mQ9NmMIJBdJtPMPG8Mso0OhM526pDzc/EKnRrIrrkHD1E0Z6tONZRmUeJElfsQ6h44lQQCcDilSNIvSQ==",
       "cpu": [
         "x64"
       ],
@@ -2982,9 +2982,9 @@
       }
     },
     "node_modules/@rolldown/binding-linux-arm-gnueabihf": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-linux-arm-gnueabihf/-/binding-linux-arm-gnueabihf-1.0.0-rc.7.tgz",
-      "integrity": "sha512-mOvftrHiXg4/xFdxJY3T9Wl1/zDAOSlMN8z9an2bXsCwuvv3RdyhYbSMZDuDO52S04w9z7+cBd90lvQSPTAQtw==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-linux-arm-gnueabihf/-/binding-linux-arm-gnueabihf-1.0.0-rc.8.tgz",
+      "integrity": "sha512-WgCKoO6O/rRUwimWfEJDeztwJJmuuX0N2bYLLRxmXDTtCwjToTOqk7Pashl/QpQn3H/jHjx0b5yCMbcTVYVpNg==",
       "cpu": [
         "arm"
       ],
@@ -2999,9 +2999,9 @@
       }
     },
     "node_modules/@rolldown/binding-linux-arm64-gnu": {
-      "version": "1.0.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@rolldown/binding-linux-arm64-gnu/-/binding-linux-arm64-gnu-1.0.0-rc.7.tgz",
-      "integrity": "sha512-TuUkeuEEPRyXMBbJ86NRhAiPNezxHW8merl3Om2HASA9Pl1rI+VZcTtsVQ6v/P0MDIFpSl0k0+tUUze9HIXyEw==",
+      "version": "1.0.0-rc.8",
+      "resolved": "https://registry.npmjs.org/@rolldown/binding-linux-arm64-gnu/-/binding-linux-arm64-gnu-1.0.0-rc.8.tgz",
+      "integrity": "sha512-tOHgTOQa8G4Z3ULj4G3NYOGGJEsqPHR91dT72u63OtVsZ7B6wFJKOx+ZKv+pvwzxWz92/I2ycaqi2/Ll4l+rlg==",
       "cpu": [
         "arm64"
       ],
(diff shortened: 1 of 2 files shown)


jetersen:
@cchanche what do you think about this approach? :sweat_smile:

I took care of fixing the action.yml path for auto label action and also fixed it to ES2024.

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
```diff
@@ -80,16 +80,18 @@ export const findPullRequests = async (params: {
...
+    ...new Map(
+      commits
+        .flatMap((commit) => commit.associatedPullRequests?.nodes ?? [])
+        .filter((pr) => pr != null)
+        .map(
+          (pr) =>
+            [`${pr.baseRepository?.nameWithOwner}#${pr.number}`, pr] as const
+        )
+    ).values()
+  ]
```
jetersen, line 94: Hmm this has been bugged for forked repos for a while but it is not often that release-drafter runs on a forked repo 😅

cchanche: What's bugged about this ? Is there an existing issue I should know of ?

jetersen: Well it was hiding a PR if you note the e2e test change :sweat_smile:

I guess it was always bugged, just nobody cared about this bug, again rarely that we have forked repos with PRs in them.

jetersen:
Decided to pick up rolldown-vite seeing as rolldown is in release candidate and the future :sweat_smile:

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Oh I see, you changed previous behavior of deduplicating PRs based on their number to deduplicating using number + `nameWithOwner`

lgtm !

cchanche: > Well it was hiding a PR if you note the e2e test change 😅

did not get to read your test changes yet

src/common/config/get-config-files.ts
```diff
@@ -60,11 +59,14 @@ export const getConfigFiles = async (
...
-        omit(extendRepoConfig.config, '_extends')
-      )
+    const { fetchedFrom: extendedFrom } = extendRepoConfig
+    const alreadyLoaded = files.find(
+      ({ fetchedFrom: loadedFrom }) =>
+        loadedFrom.scheme === extendedFrom.scheme &&
+        loadedFrom.filepath === extendedFrom.filepath &&
+        loadedFrom.ref === extendedFrom.ref &&
+        loadedFrom.repo.owner === extendedFrom.repo.owner &&
+        loadedFrom.repo.repo === extendedFrom.repo.repo
```
cchanche: What you implemented is basically what the probot config does, but I thought the local file loading I added could break this

Ex :

- start with action input `config-name:` requesting file1 locally (local loading)
- file1 `_extends:` remote file2
- file2 `extends:` remote file1

You have a loop with 3 different `loadedFrom` objects (file1 is loaded locally & remotely)


But now that I think of it, next extends will generate an _already-seen_ `loadedFrom` object. We need to make sure recursion exits at the correct moment

src/tests/drafter/drafter.e2e.test.ts
```diff
@@ -1422,6 +1422,7 @@ describe('drafter e2e', () => {
             * Add alien technology (#23) @jetersen
             * Add documentation (#5) @TimonVS
             * Update dependencies (#4) @TimonVS
+            * 👽 Add alien technology (#1) @TimonVS
             ",
                 "draft": true,
```
cchanche, line 1427: Oh, there it is ! 😅

mise.toml
cchanche, line 1: Feels redundant to the `.node-version` file, though I am not familiar to `mise`

package-lock.json
cchanche, line 1: I checked-out your branch and installed deps, lock file had significant changes

`npm: 11.8.0` & `node: 24.13.1`

package.json
```diff
@@ -32,17 +32,16 @@
...
     "lint": "npx eslint .",
     "codegen": "graphql-codegen -c src/scripts/graphql.codegen-config.ts -v",
-    "schemas": "node dist/scripts/json-schema.js && npx prettier --write **/schema.json schema.json",
+    "schemas": "npx tsx src/scripts/json-schema.ts && npx prettier --write **/schema.json schema.json",
     "all": "npm run format:write && npm run lint && npm run test && npm run coverage && npm run build"
   },
   "license": "MIT",
   "dependencies": {
-    "@actions/core": "^1.11.1",
+    "@actions/core": "^3.0.0",
```
cchanche, line 41: Nice catch !

package.json
```diff
@@ -32,17 +32,16 @@
     "format:check": "npx prettier --check .",
     "lint": "npx eslint .",
     "codegen": "graphql-codegen -c src/scripts/graphql.codegen-config.ts -v",
-    "schemas": "node dist/scripts/json-schema.js && npx prettier --write **/schema.json schema.json",
+    "schemas": "npx tsx src/scripts/json-schema.ts && npx prettier --write **/schema.json schema.json",
```
cchanche, line 36: Very nice ! `tsx` looks great and small enough

maybe one day we will get to using bun or some modern ts-native runtime altogether and spare the dependency

package.json
```diff
@@ -70,8 +68,9 @@
     "prettier": "^3.7.4",
     "prettier-eslint": "^16.4.2",
     "rollup-plugin-copy": "^3.5.0",
+    "tsx": "^4.21.0",
     "typescript": "^5.9.3",
-    "vite": "^7.3.1",
+    "vite": "npm:rolldown-vite@7.3.1",
```
cchanche: Love it ! Build times also went down on my machine, though build hangs from time to time I can't understand why

Image


(and then nothing forever)

I am guessing this may be something with my setup. If you had something similar maybe it's worth investigating

vite.config.ts
```diff
@@ -1,101 +1,57 @@
...
+    })
+  ],
+  build: {
+    target: 'node24',
+    rollupOptions: {
+      external: (id) => id.startsWith('node:') || builtinModules.includes(id),
+      input: {
+        'actions/drafter/run': 'src/actions/drafter/run.ts',
+        'actions/autolabeler/run': 'src/actions/autolabeler/run.ts'
       },
```
cchanche, line 34: Definitely simpler, nice one here too

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Acknowledged the test you mentioned below

cchanche:
> @cchanche what do you think about this approach? 😅
> 
> I took care of fixing the action.yml path for auto label action and also fixed it to ES2024.

Thank you greatly for taking care of so much optimization stuff...

Please see review comments above

src/common/config/get-config-files.ts
jetersen: I'll look at adding tests for the fix :sweat:

mise.toml
jetersen: I use mise cause it has it supports many different package managers, I did not want to install node on system level cause of conflicts :sweat_smile:

package.json
jetersen: it was pulling in two different versions of undici used by actions/core and actions/http-client :sweat_smile:

package-lock.json
jetersen: I was using `11.4.2` npm. Went ahead and bumped to latest npm and v24 nodejs

package.json
jetersen: we could switch to bun :smile:

package.json
jetersen: maybe we should jump go with vite@8.0.0 beta

src/common/config/get-config-files.ts
jetersen: Hopefully the added test and fixes cover the new local changes.
I think local file should always win over remote that potentially extends itself whether by ref or by path. Like anything else feels like someone did something wrong in my opinion.
Remember local files still have the context from github action run ie, we know which org and repo it is running in.

Code wise I prefer this fix over the omit approach simply comparing the files content.

jetersen: The key invariant: file: always wins over github: for the same filepath+repo, because the local checkout is the authoritative version regardless of what ref the remote chain references.

Obviously one can generate a local find but not put it at the same path you have remotely in the very same repo, which is fine.

package-lock.json
cchanche: I saw you added those npm and node versions to the mise config file. Maybe it's worth also changing package.json's `engine` property to keep everything generic then

package.json
cchanche: Never had the opportunity to use it, did you ? If it was up to me i'd prefer to scope this in a dedicated PR - but that could be fun to work on for sure

package.json
cchanche: Hm this feels a bit prematured tbh

But I'm not sure which is more stable : v7 with early rolldown support, or v8beta with native rolldown 😅 

If you had no problem building on your system I think that's fine. I'll do with having to retry my build commands that's no big deal ! I'll keep myself updated on vite's releases to eventually switch to v8 later on 😃

src/common/config/get-config-files.ts
```diff
@@ -58,18 +58,21 @@ export const getConfigFiles = async (
...
 
-    // Avoid loops
-    // Scheme is intentionally excluded: file:A and github:A (same filepath+ref+repo)
-    // are treated as the same source so that a local config loading a remote chain
-    // that circles back to the same file is correctly detected as recursion.
+    // Avoid loops.
+    // When a file: entry is already in the chain, any github: entry for the same
+    // filepath+repo is treated as a duplicate regardless of ref — the local file
+    // takes priority over any remote version (including at a different ref).
+    // For same-scheme comparisons the ref must also match.
```
cchanche: > is treated as a duplicate regardless of ref

Why ?

With : 

```
file:A@main -extends-> github:B@other -extends-> github:A@long-running-branch 
```

... i think contents of `file:A@main` will correctly override whatever would also be in `github:A@long-running-branch` right ? Maybe one has a dedicated branch for storing configs, and want it's actively maintained branches to extend those ?

What do you think about : 

`return sameFilepath && sameRepo && sameRef`

... ?

src/common/config/get-config-files.ts
cchanche: > Code wise I prefer this fix over the omit approach simply comparing the files content.

I agree, comparing file contents was an escape hatch of me banging my head too much on my desk on the matter 😅  glad we can think about this together now !

> I think local file should always win over remote that potentially extends itself whether by ref or by path. 

See [this comment](https://github.com/release-drafter/release-drafter/pull/1526#discussion_r2902416478). Files win over each other with the `extends` relation already. Extending a remote file that was pointed-at locally already with a different ref feels weird indeed, but should we ban this ?

> feels like someone did something wrong in my opinion.

agreed, we are touching edge edge edge cases here 😅

jetersen: Ya so with the current implementation in this PR it eliminates those cases and we even prevent the fetching of the files avoiding extra network operations.

package.json
jetersen: we can switch to bun later, indeed.

package.json
jetersen: Definitely premature https://github.com/vitejs/vite/issues/21784 :sweat:

cchanche: Ah you went all the way ! Thank you for this, I'll keep an eye on it

jetersen: Will we can force rolldown to rc 7 and it should fix the issue :thinking: at least for our use case.

package-lock.json
jetersen: well npm was the only one with blocking changes ie. different package lock output.
The nodejs version can be left open otherwise the npm i will fail if you have an older version of nodejs than the most recent release :sweat_smile:

Mise is just there to help other developers ensure they have a known good version.
https://mise.jdx.dev/

src/common/config/get-config-files.ts
jetersen: My reasoning is again, local file should win over any extends.
Simply because wanting to avoid edge cases.

If it is the other way around that github:A@long-running-branch should win over file:A@main and we exclude it from the local files from config files merge set that feels odd, it also feels odd to include both in the merge set.

package-lock.json
jetersen: 👍
```
[settings]
idiomatic_version_file_enable_tools = ['node']
```

src/common/config/get-config-files.ts
jetersen: Thinking about it some more perhaps sameRef for local file vs git repo recursion should definitely be prevented in recursion but perhaps there could be made a argument for long running branch or simply wanting to make a small change via CI with a simple replace that you still want to merge :thinking:

If you feel strongly about the edge case I can revert that change :thinking:

cchanche: Nah that's fine, I slept over it too 😅 

I came to think that if this ends up being an actual need for someone, that person would already be so deep into understanding how config loading works he would probably just open an issue and lay out actual / real-world use-cases to revert your change. We are just speculating here...

Also, I prefer we raise unnecessary errors instead of risking infinite loops. Please keep this, sorry for the fuss 😉

package.json
```diff
@@ -20,7 +20,8 @@
     "release-automation"
   ],
   "engines": {
-    "node": ">=24.0.0"
+    "node": ">=24.0.0",
+    "npm": ">=11.9.0"
```
cchanche, line 24: Ah thank you!

This will prevent me from altering lock file on my corporate laptop I can't install anything on... 😅 I'll give mise a try, let's hope it can be installed non-root in portable mode

package.json
```diff
@@ -47,6 +47,9 @@
     "yaml": "^2.8.2",
     "zod": "^4.3.5"
   },
+  "overrides": {
+    "rolldown": "1.0.0-rc.7"
```
cchanche: Nice catch on that one ! Glad we can use rolldown

cchanche:
I think we are good here ! Do you still have things you mean to work on ?

To recap what we identified as _do later stuff_ :
- switch to `bun` (and remove `tsx`)
- (draft) switch to vite `v8` stable - maybe dependabot can do this one
- [EDIT] 100% esm deps and get rid of side effects in the bundle (see [discussion](https://github.com/release-drafter/release-drafter/pull/1475#discussion_r2902620016))

package.json
cchanche: Ah, see [this comment](https://github.com/vitejs/vite/issues/21784#issuecomment-4021456895) on vite's issue you opened !

looks like `vite@8.0.0-beta.18` is already out with new rolldown version 🥳  these guys are fast

[changelog](https://github.com/vitejs/vite/blob/v8.0.0-beta.18/packages/vite/CHANGELOG.md#800-beta17-2026-03-09)

jetersen:
@cchanche all good now, bumped the version. So we are good to go, do you just want to merge it into your PR?

PR #1535, opened 2026-03-12
fix: JSON schema too strict with required fields

### LINKED ISSUE #1534: JSON Schema is too strict
## Description

I believe that the JSON schema added a lot of fields as required with the latest version, but these fields have defaults and don't need to be required.

I noticed this because I opened our release-drafter.yml file and my YAML lsp showed me a ton of errors i haven't seen before, which is what allowed me to investigate.

These required fields should probably be less strict if they have defaults and can be omitted.

Thanks!

jetersen:
## Summary

- Fixes #1534 — YAML LSPs flagged false errors because the JSON schema marked fields with defaults as `required`
- Root cause: Zod's `toJSONSchema` defaults to `io: "output"`, which treats `.default()` fields as required (since they always exist in the parsed output). The JSON schema is consumed by LSPs to validate **user input**, where defaulted fields are optional.
- Fix: Pass `{ io: 'input' }` to `toJSONSchema` so fields with `.optional().default()` are correctly not required
- Added regression tests to ensure defaulted fields are never marked as required in the generated schema

## Test plan

- [x] All 285 existing tests pass
- [x] New `json-schema.test.ts` with 3 test cases:
  - Top-level fields with defaults are not required
  - Nested fields (e.g. category items) with defaults are not required
  - `template` (no default) remains required

### DIFF
--- a/drafter/schema.json
+++ b/drafter/schema.json
@@ -152,7 +152,7 @@
       }
     },
     "category-template": { "default": "## $TITLE", "type": "string" },
-    "template": { "type": "string", "minLength": 1 },
+    "template": { "default": "", "type": "string" },
     "latest": { "anyOf": [{ "type": "string" }, { "type": "boolean" }] },
     "prerelease": { "anyOf": [{ "type": "string" }, { "type": "boolean" }] },
     "initial-commits-since": {
@@ -167,6 +167,5 @@
     "commitish": { "type": "string" },
     "header": { "type": "string" },
     "footer": { "type": "string" }
-  },
-  "required": ["template"]
+  }
 }
--- a/dist/actions/drafter/run.js
+++ b/dist/actions/drafter/run.js
@@ -1301,7 +1301,7 @@ var exclusiveConfigSchema = object({
 		default: "patch"
 	}),
 	"category-template": string().optional().default("## $TITLE"),
-	template: string().min(1)
+	template: string().optional().default("")
 }).meta({
 	title: "JSON schema for Release Drafter yaml files",
 	id: "https://github.com/release-drafter/release-drafter/blob/master/drafter/schema.json"
--- a/schema.json
+++ b/schema.json
@@ -152,7 +152,7 @@
       }
     },
     "category-template": { "default": "## $TITLE", "type": "string" },
-    "template": { "type": "string", "minLength": 1 },
+    "template": { "default": "", "type": "string" },
     "latest": { "anyOf": [{ "type": "string" }, { "type": "boolean" }] },
     "prerelease": { "anyOf": [{ "type": "string" }, { "type": "boolean" }] },
     "initial-commits-since": {
@@ -167,6 +167,5 @@
     "commitish": { "type": "string" },
     "header": { "type": "string" },
     "footer": { "type": "string" }
-  },
-  "required": ["template"]
+  }
 }
--- a/src/actions/drafter/config/schemas/config.schema.ts
+++ b/src/actions/drafter/config/schemas/config.schema.ts
@@ -144,8 +144,9 @@ export const exclusiveConfigSchema = object({
   'category-template': string().optional().default('## $TITLE'),
   /**
    * The template for the body of the draft release.
+   * Optional as it may be inherited via `_extends`.
    */
-  template: string().min(1)
+  template: string().optional().default('')
 }).meta({
   title: 'JSON schema for Release Drafter yaml files',
   id: 'https://github.com/release-drafter/release-drafter/blob/master/drafter/schema.json'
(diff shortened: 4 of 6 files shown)


jetersen:
@cchanche you been busy using semantics commits, do you want to continue that for PR titles? :thinking: or should we take the label approach. I am down for either we just have to agree on style

cchanche:
Yes I think that for us to support showing individual commits in the changelog (release body), we will have to provide some sort of categorizing capability for them like we do with PRs

The obvious mechanism is conventional semantics, since commits don't support labels like PRs do

Lucky for us, and as you suggest, supporting conventional semantics does not need to wait for us to support commits since PRs also have titles we could parse the semantics of

I would like to go down that path indeed, meaning adding the option to categorize PRs based on their title using the _conventional commits_ spec. I have some PRs waiting in my forklift I still need to rebase.

But that does not mean dropping the current label categorization, does it ?

Oh, and thx for this PR ! Glad we have some beta testers out there 

Edit : kept the forklift typo because it's hilarious 😆

jetersen:
@cchanche semantic commits do not feel like that far away, I already have an idea for creating an individual action to support checking PR title :bulb:
No we can support both in the codebase it just a matter of workflows and tests covering the coverage of both semantics prs/commits and pr label categorization

drafter/schema.json
```diff
@@ -96,105 +137,158 @@
...
-    "sort-by",
-    "sort-direction",
-    "filter-by-commitish",
-    "pull-request-limit",
-    "history-limit",
-    "replacers",
-    "categories",
-    "version-resolver",
-    "category-template",
     "template"
```
jetersen: I wonder if template is even required? Like the whole extends thing 🤔

cchanche:
I would simply like that we ship each individual functional aspect of it gradually (edit: well, there's only two, but still 2 concerns) : 

1. add support for categorizing PRs based on the [_Conventional Commits_](https://www.conventionalcommits.org/en/v1.0.0/) spec (using PR titles)
  - just find a lib that analyses titles and add some categorization customization in the config
  - i've found conventional-commits-parser convenient
2. Include individual commits from the commitish alongside PRs as changes
  - I'd like for use to represent this in some kind of class based abstraction, maybe `class Change { public type: 'pr' | 'commit' }`
  - I think we may add an option do enable (or disable this)
  - and also an option to show the 'commit'-typed changes to a dedicated section in the release. Here is how I like it : 
  
image

  - ...but people may want to have a unique "Changes" section (this can be a separate PR !)
  - and then also have separate `change-template`s for commits and PRs
  - and finally, categorize those `'commit'`-typed changes through the [_Conventional Commits_](https://www.conventionalcommits.org/en/v1.0.0/) spec (using PR titles) introduced in 1.
  
  
> I already have an idea for creating an individual action to support checking PR title 💡

Do you really wish for this to be part of release-drafter ? I feel like people should freely enforce their titles through whatever spec and/or tools they have (just like commit mesasges), and then we jsut require our internal parsing works. I feel like simply saying : "hey, do whatever you want with your titles, they jsut have to be parseable with [insert lib name]" is enough

drafter/schema.json
cchanche: Hmm good question !

Maybe the whole json schema need not have any required fields... I would be ok with that. Validation can wait runtime IMO

Are there any other required fileds except from `template` ? I don't think there are...

jetersen:
I know existing actions/scripts exists for enforcing pr title but I feel like there could be an configuration nuisance where it would make sense for us to offer first party support instead of having to ask someone else to support our configuration style and also segmentation of asking users to use different actions to cobble things together.
Maybe it is too much but it was at least my first thought thinking of the systematic approach.

Maybe worth moving some of this to discussions :thinking:

cchanche:
@jetersen so to be clear, by _enforcing_ you do mean sth that fails on your PR workflows (`event: pull_request`) when title is wrong, correct ? Or is it something like generating the title ?

One or the other, I feel like it should be another action, or at least another step running doing this _enforcing_ - or did you have a feeling this could be on the same runtime we also do release stuff (e.g. `- use: release-drafter/release-drafter@v7`) ?

Anyway, If you have something in mind that could make user's lives easier I d'ont see why not to go for it :) this also does not need to wait for the rest supporting any of what I have mentioned above  !

jetersen:
Indeed offering a README with a `event: pull_request` workflow where we would have a separate action we would support for the flagging PRs and no we would do the same thing as auto label ie creating a separate action:
`- use release-drafter/release-drafter/semantic-pr-title@v7`

And then the main `release-drafter/release-drafter` and config option or via some detection could see semantic prs exist.

cchanche:
LGTM !!

PR #1553, opened 2026-03-13
Chore/migrate to biome suggestions

cchanche:


jetersen:
pushed a fix for the check:clean :sweat_smile: making it cross platform ie. it would not work on windows.

cchanche:
> pushed a fix for the check:clean 😅 making it cross platform ie. it would not work on windows.

ah, good catch ! Isolating in a script is also nice

PR #1554, opened 2026-03-13
fix: fall back to org .github repo when config not found in current repo

jetersen:
## Summary

Restores the behavior from the [Probot config plugin](https://github.com/probot/octokit-plugin-config) that was lost in the v7 rewrite: when the config file is not found in the current repository, release-drafter automatically falls back to looking for the same file in the org/user `.github` repository.

This is a common pattern used by large organisations (e.g. jenkinsci) where a shared `release-drafter.yml` lives in `{org}/.github/.github/release-drafter.yml` and individual repos rely on the implicit fallback rather than specifying an explicit `config-name`.

## Changes

- `get-config-files.ts`: catch 404 on initial config fetch and retry against `{owner}/.github` with the same path (only when targeting the current repo via the `github:` scheme)
- `config.test.ts`: two new tests (successful fallback, both-404 error) and updated the existing "should error" test to account for the fallback attempt

## Test plan

- [ ] Config found in current repo — no change in behaviour
- [ ] Config missing in current repo, found in `{owner}/.github` — falls back silently with an info log
- [ ] Config missing in both — error references the `.github` fallback repo
- [ ] Explicit cross-repo `config-name` (e.g. `other-repo:release-drafter.yml`) — no fallback attempted
- [ ] `file:` scheme — no fallback attempted


src/common/config/get-config-files.ts
```diff
@@ -16,7 +16,37 @@ export const getConfigFiles = async (
...
+
+  let requestedRepoConfig: Awaited<ReturnType<typeof getConfigFile>>
+  try {
+    requestedRepoConfig = await getConfigFile(configTarget)
+  } catch (error) {
+    if (
+      canFallBackToOrgRepo &&
+      error instanceof Error &&
+      error.message.includes('Config file not found')
+    ) {
```
cchanche, line 37: Event if adding `configTarget.scheme === 'github'` probably does not fix anything, I'd like it if we were more explicit here. Matches your "Test plan" from the PR body, and is not explicitly tested either

```suggestion
    if (
      canFallBackToOrgRepo &&
      error instanceof Error &&
      error.message.includes('Config file not found') &&
      configTarget.scheme === 'github'
    ) {
```

src/tests/drafter/config.test.ts
```diff
@@ -76,6 +76,110 @@ describe('get config file', () => {
...
+      expect(scope.isDone()).toBe(true)
+
+      expect(res.contexts.length).toBe(1)
+      expect(res.config).toMatchInlineSnapshot(`
+        {
+          "template": "org-template-content",
+        }
+      `)
+      expect(res.contexts[0]).toMatchInlineSnapshot(`
+        {
```
cchanche, line 121: Maybe also `expect()` the core.info call. I think it's important

src/common/config/get-config-files.ts
```diff
@@ -16,7 +16,37 @@ export const getConfigFiles = async (
     `getConfigFiles: Parsed config target - scheme: ${configTarget.scheme}, filepath: ${configTarget.filepath}`,
   )
 
-  const requestedRepoConfig = await getConfigFile(configTarget)
+  const isCurrentRepoGithubScheme =
+    configTarget.scheme === 'github' &&
+    configTarget.repo.owner === currentContext.repo.owner &&
+    configTarget.repo.repo === currentContext.repo.repo
```
cchanche, line 22: There's a test-case missing for this, e.g. check we don't attempt fallback when in configTarget is on another repo

jetersen:
Addresses your review comments and added another test for non default config filename :sweat_smile:

docs/configuration-loading.md
```diff
@@ -10,6 +10,7 @@ a variety of syntax combinations to fetch your config file(s).
     - [Load your config from another repo](#load-your-config-from-another-repo)
     - [Load a file on the runner's file-system (dynamic config)](#load-a-file-on-the-runners-file-system-dynamic-config)
     - [Extend other config files using `_extends`](#extend-other-config-files-using-_extends)
+  - [Org-wide config via the `.github` repo](#org-wide-config-via-the-github-repo)
   - [Edge-cases](#edge-cases)
```
cchanche, line 14: A simple tab missing ! 

```suggestion
    - [Extend other config files using `_extends`](#extend-other-config-files-using-_extends)
    - [Org-wide config via the `.github` repo](#org-wide-config-via-the-github-repo)
  - [Edge-cases](#edge-cases)
```

PR #1558, opened 2026-03-14
feat: unify category config around change classification

cchanche:
## Summary

This reshapes `categories` from a changelog-only grouping feature into a general change-classification pipeline, which lays the groundwork for future conventional commit support.

## What changed

- add richer category parsing and schema support around `when` conditions, category `type`, `labels-mode` / `paths-mode`, `exclusive`, and `semver-increment`
- make `pre-include`, `pre-exclude`, and `version-resolver` part of the same category model used for changelog grouping
- preserve backward compatibility by translating deprecated top-level include/exclude and version-resolver settings into the new category model
- fix category/path matching edge cases, including prefilter behavior, matched-path preservation, and legacy `exclude-paths` handling
- update the README and generated schemas to document the new terminology and behavior
- add coverage for config parsing, category matching, schema validation, and path shorthand behavior

### 1. Include / exclude rules become categories

**Before**

```yml
include-labels:
  - app-foo
exclude-labels:
  - skip-changelog
exclude-paths:
  - docs/archive/**
include-paths:
  - packages/foo/**
```

**After**

```yml
categories:
  - type: "pre-include"
    when:
      labels:
        - app-foo
      paths:
        - packages/foo/**
  - type: "pre-exclude"
    when:
      label: "skip-changelog"
      paths:
        - docs/archive/**
```

### 2. Changelog category labels move into `when`

**Before**

```yml
categories:
  - title: "Bug Fixes"
    label: "type: bug"
  - title: "Maintenance"
    labels:
      - "type: maintenance"
      - "type: docs"
```

**After**

```yml
categories:
  - title: "Bug Fixes"
    when:
      label: "type: bug"
  - title: "Maintenance"
    when:
      labels:
        - "type: maintenance"
        - "type: docs"
```

### 3. Version resolution can be separated from changelog rendering

**Before**

```yml
categories:
  - title: "Features"
    label: "feature"
version-resolver:
  major:
    labels: ["major"]
  minor:
    labels: ["feature"]
  patch:
    labels: ["fix"]
  default: patch
```

**After**

```yml
categories:
  - title: "Features"
    semver-increment: "minor"
    when:
      label: "feature"
  - type: "version-resolver"
    semver-increment: "major"
    when:
      label: "major"
  - type: "version-resolver"
    semver-increment: "patch"
    when:
      label: "fix"
```

## Why

This aligns the config with how Release Drafter classifies changes internally and avoids another config reshuffle when conventional commit support is added.

Related to https://github.com/release-drafter/release-drafter/pull/1556#issuecomment-4060800582


cchanche:
@jetersen would love your opinion on this. Thank you ! 🙏 

I simply finalized how I'd like the config schema to look like, implementation is TODO

src/actions/drafter/config/schemas/config.schema.ts
```diff
@@ -97,21 +107,151 @@ export const exclusiveConfigSchema = object({
...
+       * @deprecated Use `matches.label` instead.
+       */
       label: string().min(1).optional(),
+
+      /**
+       * How different matching criterias combine.
+       *
+       * @default "labels-and-paths"
+       */
+      'matching-mode': zenum(['labels-and-paths', 'labels-or-paths'])
```
cchanche: Once we support the conventional commit spec, this will become :

```suggestion
      'matching-mode': zenum(['labels-and-paths-and-scopes-and-types-and-isBreaking', 'labels-or-paths-or-scopes-or-types-or-isBreaking', 'etc...'])
```

Now that I write about it, there may be a cleaner way to do this... 😅  open to suggestions !

src/actions/drafter/config/schemas/config.schema.ts
```diff
@@ -97,21 +107,151 @@ export const exclusiveConfigSchema = object({
...
+       *
+       * - `changelog`: Included in the generated changelog.
+       * - `pre-include`: Keep only matching changes for later changelog categorization.
+       * - `pre-exclude`: Exclude matching changes for later changelog categorization.
+       * - `version-resolver`: Used solely to determine `$RESOLVED_VERSION` based on matching changes. Use 'semver-increment' instead if this category should also be included in the changelog.
+       *
+       * `pre-include` always runs before `pre-exclude` in the pipeline.
+       *
+       * @default "changelog"
+       */
```
cchanche: My intention is to group configs under the same roof, but maybe that roof was only me defining a complex schema for matchers and not wanting to do the work multiple times...

Maybe doing this can be cleaner : 

```ts
const changeMatchersSchema = object({
  'blablabla-labels-paths-scopes-types-etc...': array()
})


const configSchema = object({
  // Changelog only
  categories: array(object({
    matches: changeMatchersSchema 
  })),

  // Prefiltering only
  "change-filters": array(object({
    type: 'include' | 'exclude'
    matches: changeMatchersSchema 
  })),
  
  // RESOLVED_VERSION only
  "version-resolve": object({
    major: changeMatchersSchema,
    minor: changeMatchersSchema,
    patch: changeMatchersSchema,
  })

})
```

... tho I quite like the idea of the `semver-increment` being mapped on the changelog categories. Everyone most probably want a patch if there are changes in the patch category (only), right ? So why is that a separate matching config and mechanism ?

src/actions/drafter/config/schemas/config.schema.ts
cchanche: Came up with a much scalable and intuitive solution

src/actions/drafter/config/schemas/config.schema.ts
cchanche: I'll just go for the all-in-one solution like it is done now

jetersen:
Seems like a breaking change?

Are we going to support the old version while giving people a chance to migrate or are we going for a major breaking?

cchanche:
Yep, fully backwards compatible with depreciation warnings.

You can look at the [smoke tests runs](https://github.com/release-drafter/release-drafter/actions/runs/26391894468/job/77683340056) from before the last commit where I migrate the config.

I also added a test suite that checks behavior is the same between an old and a new config

jetersen (approved):
LGTM

cchanche:
Thank you for looking at it !

PR #1561, opened 2026-03-17
docs: update README with pull_request_target example

jmeridth:
Now that https://github.com/release-drafter/release-drafter/pull/1560 has merged it is helpful to have this example in the README.

This was borrowed from [previous version of the README](https://github.com/release-drafter/release-drafter/blob/d849c828c3684bc16877d600d7c0a3c685ca564c/README.md) before v7.

`npm run all` was run with no changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,6 @@ on:
     branches:
       - main
       - master
-  # pull_request_target event is required for autolabeler to support PRs from forks
-  # pull_request_target:
-  #   types: [opened, reopened, synchronize]
 
 permissions:
   contents: write # allow GITHUB_TOKEN to update releases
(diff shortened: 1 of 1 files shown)


README.md
cchanche: This part of the doc is for the drafter action, please add to the autolabeler part below 🙏

jmeridth: Done https://github.com/release-drafter/release-drafter/pull/1561/changes/04d737e2fbda062cab65b144c855138d7eb41976

jmeridth:
@cchanche please let me know if my change is too much. Seems to be the minimal a user would need to know. wdyt?

cchanche:
> @cchanche please let me know if my change is too much. Seems to be the minimal a user would need to know. wdyt?

No it is quite good ! We have an issue pending somewhere to make a better doc section about permissions, this will do in the mean time

thank you :)

PR #1563, opened 2026-03-18
fix: implement disable-releaser and disable-autolabeler inputs

### LINKED ISSUE #1562: Bug: `disable-releaser` and `disable-autolabeler` inputs are defined in action.yml but unimplemented in v7.x
## Bug Description

The `disable-releaser` and `disable-autolabeler` inputs are defined in `action.yml` (and `drafter/action.yml`) but are **never read or acted upon** in the compiled action code. This affects both v7.0.0 (`3a7fb5c8`) and v7.1.0 (`44a942e4`).

### 1. Inputs defined in `action.yml`

Both inputs are present in [`action.yml`](https://github.com/release-drafter/release-drafter/blob/44a942e465867c7465b76aa808ddca6e0acae5da/action.yml#L82-L91):

```yaml
  disable-releaser:
    description: |
      A boolean indicating whether the releaser mode is disabled.
    required: false
    default: ''
  disable-autolabeler:
    description: |
      A boolean indicating whether the autolabeler mode is disabled.
    required: false
    default: ''
```

### 2. `getActionInput()` never reads them

In [`src/actions/drafter/config/get-action-inputs.ts`](https://github.com/release-drafter/release-drafter/blob/44a942e465867c7465b76aa808ddca6e0acae5da/src/actions/drafter/config/get-action-inputs.ts), the `getActionInput()` function does not include `disable-releaser` or `disable-autolabeler`:

```typescript
const actionInput: Record<keyof ActionInput, ReturnType<typeof getInput>> = {
    'config-name': getInput('config-name'),
    name: getInput('name'),
    tag: getInput('tag'),
    version: getInput('version'),
    publish: getInput('publish'),
    token: getInput('token'),
    latest: getInput('latest'),
    prerelease: getInput('prerelease'),
    'initial-commits-since': getInput('initial-commits-since'),
    'prerelease-identifier': getInput('prerelease-identifier'),
    'include-pre-releases': getInput('include-pre-releases'),
    commitish: getInput('commitish'),
    header: getInput('header'),
    footer: getInput('footer'),
    'dry-run': getInput('dry-run'),
    'filter-by-range': getInput('filter-by-range'),
}
```

Neither `disable-releaser` nor `disable-autolabeler` appear anywhere in the TypeScript source (`grep` across the entire `src/` tree returns zero hits).

### 3. `main()` unconditionally runs the full pipeline

In [`src/actions/drafter/main.ts`](https://github.com/release-drafter/release-drafter/blob/44a942e465867c7465b76aa808ddca6e0acae5da/src/actions/drafter/main.ts), the drafter always executes the full release-creation pipeline:

```typescript
const { draftRelease, lastRelease } = await findPreviousReleases(config)
const { commits, pullRequests } = await findPullRequests({ lastRelease, config })
const releasePayload = buildReleasePayload({ ... })
const upsertedRelease = await upsertRelease({ ... })
```

There is no conditional check to skip release creation when `disable-releaser: true`.

## Impact

This breaks the **least-privilege split-workflow pattern** that was recommended in v5/v6 and is documented in the README:

- **Autolabeler workflow** (triggered on `pull_request`) with `contents: read` + `pull-requests: write` uses `disable-releaser: true` — but the drafter still attempts to create a release and fails with:

  ```
  Error: Resource not accessible by integration
  ```

- **Release drafter workflow** (triggered on `push`) with `contents: write` uses `disable-autolabeler: true` — but this toggle is also ignored.

Users who migrated from v5/v6 with `disable-releaser: true` / `disable-autolabeler: true` are silently broken. The action accepts the inputs without warning but does not honour them.

## Reproduction

1. Create a workflow with `disable-releaser: true` and only `contents: read` permission
2. Use `release-drafter/release-drafter@v7` (or `@v7.1.0`)
3. Open a pull request
4. Observe the action attempts to create a release and fails with `Resource not accessible by integration`

Example failure: https://github.com/lfreleng-actions/github2gerrit-action/actions/runs/23248074942

## Expected Behaviour

When `disable-releaser: true` is set, the drafter's `runner.ts` should skip calling `main()` entirely (or `main()` should exit early), and only run the autolabeler logic.

When `disable-autolabeler: true` is set, the autolabeler should be skipped.

## Versions Affected

- v7.0.0 (`3a7fb5c8`)
- v7.1.0 (`44a942e4`)

## Related

- #1559 — autolabeler `pull_request_target` support (different bug, same area; the workaround in that issue suggests pinning to v6.4.0 with `disable-releaser: true`, which also won't work in v7 due to this bug)
- The v7 rewrite split the action into separate `drafter/` and `autolabeler/` sub-actions, but the root `action.yml` still advertises `disable-releaser`/`disable-autolabeler` as inputs despite them being dead code

ModeSevenIndustrialSolutions:
## Summary

The `disable-releaser` and `disable-autolabeler` inputs are defined in `action.yml` (and `drafter/action.yml`) but were never wired into the TypeScript source or compiled JavaScript. They were silently accepted but completely ignored.

This broke the **least-privilege split-workflow pattern** where:
- The autolabeler workflow runs with `contents: read` and `disable-releaser: true`
- The release drafter workflow runs with `contents: write` and `disable-autolabeler: true`

When `disable-releaser: true` was set but ignored, the drafter unconditionally attempted to create a release and failed with:
```
Error: Resource not accessible by integration
```

### Drafter (`src/actions/drafter/`)
- **`config/schemas/action-input.schema.ts`** — Added `disable-releaser` boolean field to `exclusiveInputSchema`
- **`config/get-action-inputs.ts`** — Added `disable-releaser` to the `getActionInput()` function
- **`runner.ts`** — Added early-exit check: when `disable-releaser` is truthy, logs an info message and returns before any release operations

### Autolabeler (`src/actions/autolabeler/`)
- **`config/action-input.schema.ts`** — Added `disable-autolabeler` boolean field to `actionInputSchema`
- **`config/get-action-inputs.ts`** — Added `disable-autolabeler` to the `getActionInput()` function
- **`runner.ts`** — Added early-exit check: when `disable-autolabeler` is truthy, logs an info message and returns before any labeling operations

### Compiled output
- Rebuilt `dist/actions/drafter/run.js` and `dist/actions/autolabeler/run.js` to include the fix

## Testing

- All 304 existing tests pass (`npx vitest run`)
- Verified the compiled dist files contain the new input parsing and early-exit logic via `grep`

## Affected versions

- v7.0.0 (`3a7fb5c8`)
- v7.1.0 (`44a942e4`)

Fixes #1562

### DIFF
--- a/dist/actions/autolabeler/run.js
+++ b/dist/actions/autolabeler/run.js
@@ -1,6 +1,9 @@
-import { C as warning, S as setOutput, T as __toESM, b as info, g as context, h as getOctokit, l as object, m as composeConfigGet, o as array, r as sharedInputSchema, t as stringToRegex, u as string, w as __commonJSMin, x as setFailed, y as getInput } from "../../chunks/common.js";
+import { C as warning, S as setOutput, T as __toESM, b as info, d as stringbool, g as context, h as getOctokit, l as object, m as composeConfigGet, o as array, r as sharedInputSchema, s as boolean, t as stringToRegex, u as string, w as __commonJSMin, x as setFailed, y as getInput } from "../../chunks/common.js";
 //#region src/actions/autolabeler/config/action-input.schema.ts
-var actionInputSchema = object({ "config-name": string().optional().default("release-drafter.yml") }).and(sharedInputSchema);
+var actionInputSchema = object({
+	"config-name": string().optional().default("release-drafter.yml"),
+	"disable-autolabeler": stringbool().or(boolean()).optional()
+}).and(sharedInputSchema);
 //#endregion
 //#region src/actions/autolabeler/config/config.schema.ts
 var configSchema = object({ autolabeler: array(object({
@@ -20,7 +23,8 @@ var getActionInput = () => {
 	return actionInputSchema.parse({
 		"config-name": getInput$1("config-name"),
 		token: getInput$1("token"),
-		"dry-run": getInput$1("dry-run")
+		"dry-run": getInput$1("dry-run"),
+		"disable-autolabeler": getInput$1("disable-autolabeler")
 	});
 };
 //#endregion
@@ -372,6 +376,10 @@ var main = async (params) => {
 async function run() {
 	try {
 		const input = getActionInput();
+		if (input["disable-autolabeler"]) {
+			info("Autolabeler is disabled via disable-autolabeler input. Skipping labeling.");
+			return;
+		}
 		const { labels, pr_number } = await main({
 			config: parseConfig({ config: await getConfig(input["config-name"]) }),
 			dryRun: input["dry-run"]
--- a/dist/actions/drafter/run.js
+++ b/dist/actions/drafter/run.js
@@ -25,7 +25,8 @@ var actionInputSchema = object({
 	name: string().optional(),
 	tag: string().optional(),
 	version: string().optional(),
-	publish: stringbool().optional().default(false)
+	publish: stringbool().optional().default(false),
+	"disable-releaser": stringbool().or(boolean()).optional()
 }).and(sharedInputSchema).and(commonConfigSchema);
 //#endregion
 //#region src/actions/drafter/config/get-action-inputs.ts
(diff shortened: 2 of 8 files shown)


cchanche:
Hey @ModeSevenIndustrialSolutions thank you for contributing.

`disable-releaser` and `disable-autolabeler` were removed in `v7.0.0` (see https://github.com/release-drafter/release-drafter/pull/1475)

We simply forgot to remove them from the action.yml file.

ModeSevenIndustrialSolutions:
Ah great, I've reviewed https://github.com/release-drafter/release-drafter/pull/1475 and can see what I think I need to get done. Let me attempt the fix and I'll happily cleanup the Issue and PRs here (close them!).

cchanche:
superseded by #1564

Issue #1418, opened 2024-11-22
Provide instructions on how to setup an access token


jedschneider:
Specifically now with fine grained tokens, its pretty confusing to know exactly what permissions the token requires in order for the release update to run smoothly. A table or some kind of documentation in the README would be a resource i would probably return to on the regular.

Thanks!

jetersen:
It's right in the top of the readme:

```yaml
permissions:
  contents: read
```

jedschneider:
Hmm. maybe I should have been more specific; the example workflow has the permissions required for the ci action, but I think that is different than the scopes that are required of the github token that is attached to the release-drafter workflow. The permissions as declared in the workflow are not a 1:1 match to the backing token, and with the introduction of [fine grained access tokens](https://docs.github.com/en/rest/authentication/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2022-11-28) there are a lot of options to choose from. In the 4 or 5 times over the last 5 months I've introduced release-drafter to a repo the tuning of this access is the most difficult part of the process. I have not successfully gotten a fine grain access token to work, and just fall back to the legacy one.

permissions; contents: read. yes, but read _what_? Issues, releases, tags, etc?

I think it would be useful to have some documentation that provides folks with a set of actionable steps and a review least-access scopes that allows the drafter to run without over-permissioning the token.  I'm not familiar enough with the project to know the ins and outs of this but I'm happy to help review, test, or support in any other way I can.

cchanche:
That is a good idea !

I have mapped-out the api calls already, but graphql calls are a different breed... I'll have to continue another time. For reference, here is what the docs say about GraphQL permissions : https://docs.github.com/en/graphql/guides/forming-calls-with-graphql?apiVersion=2022-11-28&versionId=free-pro-team%40latest#authenticating-with-graphql

- `get /repos/{owner}/{repo}/releases`
  - "Contents" repository permissions (read)
  - see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release
  - required by the drafter action
- `post /repos/{owner}/{repo}/releases`
  - "Contents" repository permissions (write) and "Contents" repository permissions (write) and "Workflows" repository permissions (write)
  - see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release
  - required by the drafter action
- `patch /repos/{owner}/{repo}/releases/assets/{asset_id}`
  - "Contents" repository permissions (write)
[...]

PR #1564, opened 2026-03-18
fix: remove disable-releaser and disable-autolabeler from action.yaml

### LINKED ISSUE #1562: Bug: `disable-releaser` and `disable-autolabeler` inputs are defined in action.yml but unimplemented in v7.x

cchanche:
close #1562
superseds #1563

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -78,16 +78,6 @@ inputs:
       A string that would be added after the template body.
     required: false
     default: ''
-  disable-releaser:
-    description: |
-      A boolean indicating whether the releaser mode is disabled.
-    required: false
-    default: ''
-  disable-autolabeler:
-    description: |
-      A boolean indicating whether the autolabeler mode is disabled.
-    required: false
-    default: ''
   dry-run:
     description: |
       A boolean indicating whether to run without performing any write operations.
--- a/drafter/action.yml
+++ b/drafter/action.yml
@@ -78,16 +78,6 @@ inputs:
       A string that would be added after the template body.
     required: false
     default: ''
-  disable-releaser:
-    description: |
-      A boolean indicating whether the releaser mode is disabled.
-    required: false
-    default: ''
-  disable-autolabeler:
-    description: |
-      A boolean indicating whether the autolabeler mode is disabled.
-    required: false
-    default: ''
   dry-run:
     description: |
       A boolean indicating whether to run without performing any write operations.
--- a/src/types/github.graphql.generated.ts
+++ b/src/types/github.graphql.generated.ts
@@ -10599,6 +10599,8 @@ export type IssueFieldSingleSelectValue = IssueFieldValueCommon & Node & {
   id: Scalars['ID']['output'];
   /** The option's name. */
   name: Scalars['String']['output'];
+  /** The selected option's global relay ID. */
+  optionId?: Maybe<Scalars['String']['output']>;
 };
 
 /** Represents a text issue field. */


cilindrox:
Just a heads up that this did not address the `Resource not accessible by integration` error - seems the conditionals are missing/the code blocks are still being called @cchanche

cchanche:
> Just a heads up that this did not address the `Resource not accessible by integration` error - seems the conditionals are missing/the code blocks are still being called @cchanche

Hm, weird... can you please share your workflow yaml @cilindrox  ?

avdv:
> Just a heads up that this did not address the `Resource not accessible by integration` error - seems the conditionals are missing/the code blocks are still being called @cchanche

But that's the point: the autolabeler functionality has been split out into the autolabeler action. If you use the release-drafter action, it will try to draft a release... :smile: 

Just use the [autolabeler action](https://github.com/release-drafter/release-drafter#autolabeler) instead of release-drafter with `disable-releaser: true`.

cilindrox:
I get that, no neeed for the condescending tone @avdv

Here's the workflows @cchanche 

```yaml
# .github/workflows/auto-label.yml
name: Auto Label

# yamllint disable-line rule:truthy
on:
  pull_request:
    types:
    - opened
    - reopened
    - synchronize

permissions:
  contents: read

jobs:
  auto_label:
    permissions:
      pull-requests: write
    runs-on: ubuntu-24.04
    steps:
    - uses: release-drafter/release-drafter/autolabeler@139054aeaa9adc52ab36ddf67437541f039b88e2  # v7.1.1
```

```yaml
# .github/workflows/release-drafter.yml
name: Release Drafter

# yamllint disable-line rule:truthy
on:
  push:
    branches:
    - main

  # we call this workflow directly to compose releases after a tag has been cut - thus the inputs
  workflow_call:
    inputs:
      tag:
        required: true
        type: string
      version:
        required: true
        type: string
      publish:
        required: false
        type: boolean

permissions:
  contents: read

jobs:
  update_release_draft:
    permissions:
      contents: write
    runs-on: ubuntu-24.04
    steps:
    - uses: release-drafter/release-drafter@139054aeaa9adc52ab36ddf67437541f039b88e2  # v7.1.1
      with:
        config-name: release-drafter.yml
        tag: ${{ inputs.tag }}
        version: ${{ inputs.version }}
        publish: ${{ inputs.publish }}
        token: ${{ secrets.GITHUB_TOKEN }}
```

And here's the config:

```yaml
name-template: v$RESOLVED_VERSION
tag-template: $RESOLVED_VERSION
exclude-contributors:
- github-actions
categories:
- title: New features
  labels:
  - enhancement
- title: Bug fixes
  labels:
  - bug
- title: Documentation
  labels:
  - docs
- title: Other changes
change-title-escapes: \<*_&
version-resolver:
  major:
    labels:
    - breaking-change
  minor:
    labels:
    - enhancement
  patch:
    labels:
    - bug
    - dependencies
  default: patch
template: |
  ## Changes since v$PREVIOUS_TAG

  $CHANGES

replacers:
- search: /@(renovate|dependabot(?:\[bot\])?)(?:,| |$)/gm
  replace: ''

autolabeler:
- label: docs
  files:
  - '*.md'
  title:
  - /docs/i
  branch:
  - /docs{0,1}\/.+/
- label: bug
  branch:
  - /fix\/.+/
  title:
  - /fix/i
- label: enhancement
  title:
  - /feat(\(.*\))?:/i
  branch:
  - /feat\/.+/
- label: breaking-change
  title:
  - /feat(\(.*\))?\!/i
  body:
  - /BREAKING-CHANGE/i
```

avdv:
> I get that, no neeed for the condescending tone @avdv

@cilindrox  sorry if that came off that way, I just wanted to be helpful.

cilindrox:
No worries @avdv - just trying to get to the bottom of the issue - as you can see from the config above, it's pretty vanilla, that's why I'm wondering if the original PR mentioning the logic skip conditional missing has a point

avdv:
And which workflow is failing for you? Both?

cilindrox:
`release-drafter` is the one failing - on merges to `main` it fails with the following trace:

```
Run release-drafter/release-drafter@139054aeaa9adc52ab36ddf67437541f039b88e2
Parsing inputs and configuration...
Config fetched on remote "example/foo@refs/heads/main".
Fetching releases from GitHub...
Found 442 releases
Draft release:
  tag_name:  1.1.0
  name:      v1.1.0
Last release:
  tag_name:  1.0.1
  name:      v1.0.1
Fetching parent commits of refs/heads/main since 2026-03-17T14:45:38Z...
Error: Request failed due to following response errors:
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
```

cchanche:
@cilindrox  For your release-drafter (drafter) workflow, would you be willing to :
- try removing `token: ${{ secrets.GITHUB_TOKEN }}`
- try top-level `premissions: contents: write` and remove override in job definition

... ? Thank you ! 🙏

cilindrox:
@cchanche tried both, first the token, then the token and global perms, and still get the same error - here's the workflow def for reference:

```yaml
name: Release Drafter

# yamllint disable-line rule:truthy
on:
  push:
    branches:
    - main

  workflow_call:
    inputs:
      tag:
        required: true
        type: string
      version:
        required: true
        type: string
      publish:
        required: false
        type: boolean

permissions:
  contents: write

jobs:
  update_release_draft:
    runs-on: ubuntu-24.04
    steps:
    - uses: release-drafter/release-drafter@139054aeaa9adc52ab36ddf67437541f039b88e2  # v7.1.1
      with:
        config-name: release-drafter.yml
        tag: ${{ inputs.tag }}
        version: ${{ inputs.version }}
        publish: ${{ inputs.publish }}
```

cchanche:
Okay yes so this is defenitely weird. You obfuscated `example/foo@refs/heads/main`, which is ok, but I'm just wondering if this is also the repo (`example/foo`) the action is running on ? (e.g. not running from a fork or something)

cilindrox:
not a fork, the repo it's running is indeed `example/foo` (deobfuscated ofc).

cchanche:
Do you overwrite `GITHUB_TOKEN` in your CICD variables secrets somehow ? In Settings > Secrets and Variables menu

cchanche:
Do you use Github enterprise runners ?

cilindrox:
No GHE runners, running on stock GH ones - no `GITHUB_TOKEN` overrides afaict, it should be using the temp token generated for the run.

cchanche:
I may need to ope a dedicated bug for this then, I'll try to replicate this afternoon whe I get home. Sorry for the fuss 

As a last attempt, could you try leaving hte permission stuff fully empty ? 

see https://github.com/release-drafter/release-drafter/blob/master/.github/workflows/draft.yml

cchanche:
I wonder if the `workflow_call` tokens are different

cilindrox:
Went ahead and removed all the workflow dispatch blocks, basically running as barebones as it gets, still the same error:

```yaml
name: Release Drafter

on:
  push:
    branches:
    - main

jobs:
  update_release_draft:
    runs-on: ubuntu-24.04
    steps:
    - uses: release-drafter/release-drafter@139054aeaa9adc52ab36ddf67437541f039b88e2  # v7.1.1
```

cilindrox:
Also switched the runner image to `ubunut-latest`, and still no dice

cchanche:
I'm sorry @cilindrox , could'nt reproduce... 

See : https://github.com/cchanche/release-drafter/actions/runs/23310514294
See : https://github.com/cchanche/release-drafter/actions/runs/23310536032

I've added you to my fork project where these exact workflows ran, so you can look around how the repo is configured. The only thing I can imagine at this point is that your repo on github has changed your token's permissions somehow

cilindrox:
thanks @cchanche - I'll take a look and report back any findings

ViacheslavKudinov:
Hi,
i had to add  `pull-requests: read` on push event to make it working.
But it is most likely Action' bug if it is not expected to do anything with PRs on push events, but can't say for sure, sorry.

My working wf:
```
name: Release draft for repository

on:
  push:
    branches:
      - master

permissions:
  contents: write
  pull-requests: read # was failing without it

jobs:
  release:
    name: Draft a release
    runs-on: ubuntu-24.04

    steps:
      - name: Prepare a draft of release
        uses: release-drafter/release-drafter@v7
        with:
          config-name: release-drafter.yml
```

avdv:
Oh yes, good find @ViacheslavKudinov.

@cilindrox  maybe you have the workflow permissions in your repo (at project/settings/actions) set to "Read repository contents and packages permissions" only which means "Workflows have read permissions in the repository for the contents and packages scopes only."?

cchanche:
Aaah yes that's probably it ! Thank you @avdv 

Here is the setting I use in cchanche/release-drafter :

{3407C28C-BB57-4F06-A3ED-B184BCC3073F}

related to https://github.com/release-drafter/release-drafter/issues/1418#issuecomment-4096956629

cilindrox:
Thanks for the suggestions all - unfortunately, I do have the `permissions.contents: write` (tried globally and in the job spec). I also tried adding the `pull-requests: read` and I still get the same error.

This is definitely a scope/perms issue, but I can't seem to figure out which one(s) I'm missing.

ViacheslavKudinov:
my repo workflow permissions as on this screen https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4097930990

i could suggest only to try on workflow level to set
`permissions: write-all`, it is max what can be set via workflow syntax (just to see if it helps).

But it is even more odd than how it was for me.

cchanche:
@cilindrox since you are in a private repo, do you also happen to be in an organisation ? Maybe there's token settings over there too (org-wide =/= repo-wide). I'll look at the orgs I have at hand and see what's what.

EDIT: at `https://github.com/organizations/[org]/settings/actions`, ther's the exact same "Workflow permissions" thing as in https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4097930990. have you had the opportunity to check this one out ?

EDIT: also check-out `https://github.com/organizations/[org]/settings/secrets/actions` if you are not overriding `GITHUB_TOKEN`

ViacheslavKudinov:
> EDIT: also check-out https://github.com/organizations/[org]/settings/secrets/actions if you are not overriding GITHUB_TOKEN

It is not possible to do as [GH prevents](https://docs.github.com/en/actions/reference/security/secrets#naming-your-secrets) to set any `GITHUB_` secrets 

image

ViacheslavKudinov:
For testing I set on org level only "read"
and then on repo level no option to set "write" (its disabled)
image

so, i think, if @cilindrox sees on repo level permissions as there https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4097930990, then with repo settings probably all is fine

cchanche:
> > EDIT: also check-out [https://github.com/organizations/[org]/settings/secrets/actions](https://github.com/organizations/%5Borg%5D/settings/secrets/actions) if you are not overriding GITHUB_TOKEN
> 
> It is not possible to do as [GH prevents](https://docs.github.com/en/actions/reference/security/secrets#naming-your-secrets) to set any `GITHUB_` secrets

Yeah ok my bad, good to know tho ! 😅 Thank you @ViacheslavKudinov

ViacheslavKudinov:
> > > EDIT: also check-out [https://github.com/organizations/[org]/settings/secrets/actions](https://github.com/organizations/%5Borg%5D/settings/secrets/actions) if you are not overriding GITHUB_TOKEN
> > 
> > 
> > It is not possible to do as [GH prevents](https://docs.github.com/en/actions/reference/security/secrets#naming-your-secrets) to set any `GITHUB_` secrets
> 
> Yeah ok my bad, good to know tho ! 😅 Thank you @ViacheslavKudinov

no worries. It is many moving pieces. Not possible to remember everything.

Im thinking if it could be possible to add on Action "core.debug()" when "Enable debug logging" checkbox selected to get some more output.

cilindrox:
> For testing I set on org level only "read" and then on repo level no option to set "write" (its disabled) image
> 
> so, i think, if @cilindrox sees on repo level permissions as there [#1564 (comment)](https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4097930990), then with repo settings probably all is fine

Can confirm, this ^ is the current setup. Previously, with v6.x, `contents: write` was enough for the action to run, so there might be some more privileges required for running on private org repos after the TS rewrite.

`write-all` still results in the same error

cchanche:
Are you still able to run under `v6` ? If you are not too much locked-in with `v7`, maybe that could help with your current needs, and also help us determine this is in fact something with the rewrite

cchanche:
> add on Action "core.debug()" when "Enable debug logging" checkbox selected to get some more output.

What are you suggesting we debug @ViacheslavKudinov ?

The following strongly hints at the graphql request not going out :

```log
Fetching parent commits of refs/heads/main since 2026-03-17T14:45:38Z...
Error: Request failed due to following response errors:
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
 - Resource not accessible by integration
```

... the many repeated errors being pagination I guess.

You suggest we print some more stuff in between ?

ViacheslavKudinov:
Its not easy to say what "debug info" could help to understand what is wrong there.
Just thinking loudly what else can help to get better understanding.
Maybe graphql query to see its correct or not. At least it will be the chance to run on something like Altair or other graphql client.

cchanche:
Nah it's always a good thing to have more logs anyway, you are correct. I'll just push that on my fork and @cilindrox if you are still around you'll be able to just use it and give us your output :)

cchanche:
There you go : https://github.com/cchanche/release-drafter/actions/runs/23357047090/job/67950533606

`uses: cchanche/release-drafter@master` is available with more debug logs @cilindrox 🙏

ViacheslavKudinov:
~~when i run on my repo i have not all these queries~~ it is actually one query


PS when i run on fresh repo without tags/releases i got 
```
Error: Request failed due to following response errors:
 - Resource not accessible by integration
```
 but after creating the first release it started to work without `pull_requests: read`  🤷‍♂️
 
 but it is a private repo on non Enterprise org, on Enterprise i had to add  `pull_requests: read` 

PSS With a single release now i see more debug queries, but without one seems it was failing  somewhere earlier.

ViacheslavKudinov:
screenshot when it failed without any tags/releases
Screenshot 2026-03-20 at 20 11 20

ViacheslavKudinov:
Hm, hm
i removed tag and release, added `pull-requests: read` and now Action doesn't fail as was failing without this permission, even when no tags/releases.  

It found
```
Found 24 commits.
Found 7 pull requests associated with those commits. 7 of those are merged and come from vk-or/test-repo-tf : #8, #7, #4, #5, #3, #2, #1
```

im confused

ViacheslavKudinov:
i can't use this fork on repo where it was failing without `pull-requests: read` which has tags/releases as it is allow list from admins

we need  @cilindrox if he could use this fork 🙏🏻

cchanche:
My guess is : ~~what you identified as being fixed by "adding a release" may have actually been fixed by adding commits (which led to you being able to create a release and tag). Would that be correct ?~~

EDIT : there must have been at least 1 commit prior to the workflow running, or else no workflow file could exist for the workflow to run 😅 so I don't know what happened there. I'll try to replicate : empty repo, single commit adding workflow file, am I correct ?

When log (regular log) "Fetching parent commits of refs/heads/main" appear, we are about to lookup commits and their associated pull requests on the repo. If no preious releases were found, we just scan a whole lot of them (there's a limit in the settings), and the log above shows no date

When we do find a previous release, the log is completed with " since 2026-03-17T14:45:38Z..."

But in any case, the same query triggers right after :

```gql
query findCommitsWithAssociatedPullRequests(
  $name: String!
  $owner: String!
  $targetCommitish: String!
  $withPullRequestBody: Boolean!
  $withPullRequestURL: Boolean!
  $since: GitTimestamp
  $after: String
  $withBaseRefName: Boolean!
  $withHeadRefName: Boolean!
  $pullRequestLimit: Int!
  $historyLimit: Int!
) {
  repository(name: $name, owner: $owner) {
    object(expression: $targetCommitish) {
      ... on Commit {
        __typename
        history(first: $historyLimit, since: $since, after: $after) {
          __typename
          totalCount
          pageInfo {
            __typename
            hasNextPage
            endCursor
          }
          nodes {
            __typename
            id
            committedDate
            message
            author {
              __typename
              name
              user {
                __typename
                login
              }
            }
            associatedPullRequests(first: $pullRequestLimit) {
              __typename
              nodes {
                __typename
                title
                number
                url @include(if: $withPullRequestURL)
                body @include(if: $withPullRequestBody)
                author {
                  __typename
                  login
                  url
                }
                baseRepository {
                  __typename
                  nameWithOwner
                }
                mergedAt
                isCrossRepository
                labels(first: 100) {
                  __typename
                  nodes {
                    __typename
                    name
                  }
                }
                merged
                baseRefName @include(if: $withBaseRefName)
                headRefName @include(if: $withHeadRefName)
              }
            }
          }
        }
      }
    }
  }
}

```

> i can't use this fork on repo where it was failing without pull-requests: read which has tags/releases as it is allow list from admins

I did not get that sorry 😢  Do you need to use my `cchanche/release-drafter` fork somehow ? I'll just provide access anyway, so you can make changes if you want.

cchanche:
> EDIT : there must have been at least 1 commit prior to the workflow running, or else no workflow file could exist for the workflow to run 😅 so I don't know what happened there. I'll try to replicate : empty repo, single commit adding workflow file, am I correct ?

image

Cannot create release ! That's another request... namely :

- `post /repos/{owner}/{repo}/releases`
  - "Contents" repository permissions (write) and "Contents" repository permissions (write) and "Workflows" repository permissions (write)
  - see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release
  - required by the drafter action
  
  
 For reproduction, this is the workflow : 
 
 ```yaml
name: Release Drafter

on:
  push:
    branches:
      - main

jobs:
  update_release_draft:
    runs-on: ubuntu-latest
    steps:
      - uses: cchanche/release-drafter@master
 ```
 
 and here is how token is configured : 
 
 
image


So this kind of makes sence. I can read stuff (contents), and fetching commits was no problems. 

Now about pull_requests, I was under the impression the token needed extra stuff to read them, PRs not falling into the `contents` category, but maybe graphql is smart enough not to fail the `associatedPullRequests` is empty - because I don't have any !

And then it failed when creating a release, which makes sense because my token is in read only

ViacheslavKudinov:
Sorry, it was confusing about usage on Enterprise organization.
There is allow list of Actions and i can't use the fork as its not allowed. 

Only could do on my own org, which is not Enterprise. 
But there is behavior that I shared before. Its different to what I got on Enterprise one earlier today.

Its probably not Enterprise itself the issue, but some "conditions" to figure out.

I hope @cilindrox could use fork with more debug on their repo.

cchanche:
> There is allow list of Actions and i can't use the fork as its not allowed.

Ah, got it ! If you are willing (still), you could create an internal fork of release-drafter. Would that be on the _allow-list_ ? But don't feel obliged, @cilindrox reported not being in Enterprise anyway (if I recall correctly)

I just edited my previous comment with my thoughts on what I was able to reproduce. I will go ahead an keep playing with it to see if I can get to this : https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4100442303

> Only could do on my own org

Is this where you were able to get this : https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4100442303 ?

cchanche:
Yep, so iterating on my reproduction repo, here is the current graph : 

image


Workflow that ran on `chore: two commits now` had the exact same output : 

image

And workflows that ran on `Merge pull request #1 from cchanche/chore/first-pr` **AND** `chore: another commit after pr merge` both had the following : 

image


My guess is that the `associatedPullRequest` relation in the graphql query here :

```gql
query findCommitsWithAssociatedPullRequests() {
  repository() {
    object() {
      ... on Commit {
        history() {
          totalCount
          nodes {
            id
            committedDate
            associatedPullRequests() { # <-- relation is here
              nodes {
                title
                number
                merged
              }
            }
          }
        }
      }
    }
  }
}
```

...only asks permissions to _read pull_requests_ out of the token when there actually are associated pull requests

cchanche:
So that settles how token permissions work for grpahql calls, and that `pull-requests: read` indeed is required by our graphql queries (on top of `contents: write`)


Now the mystery is what the *** is happening with our friend's @cilindrox setup 😅  My guess is that his token is missing `pull-requests: read` because it's the one thet triggered the same error for me, but it may also be missing either `contents: write` or `contents: read` for the matter... (because the error he has may be because he cannot read commits all together, not just PRs)

cchanche:
Just learned about this : 

image

The above is from my reproduction private repo that has the config from this comment : https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4100747335

And the following : 

image

Is from this workflow : https://github.com/release-drafter/release-drafter/actions/runs/23356272883/job/67947770115 . This is release-drafter's public repo. I don't have permissions to see the repo's settings there, but the workflow file does not specify any special permission : 

https://github.com/release-drafter/release-drafter/blob/0c6d0637cb68460a43d1bc087b2eabc5e1623465/.github/workflows/draft.yml

So I guess it was set to read & write as shown in https://github.com/release-drafter/release-drafter/pull/1564#issuecomment-4097930990

When I change to read & write at repo level in my reproduction repo (private), here is what happens : 

image


Same `GITHUB_TOKEN Permissions` output as in https://github.com/release-drafter/release-drafter/actions/runs/23356272883/job/67947770115 ! So release-drafter's public repo probably has read & write permissions at repo level too. Also, workflow is successfull, confirming that this read & write permissions at repo level are enough for graphql to fetch those assocaitedPullRequests AND create a release.

I'd like to see the output of this `Set up job > GITHUB_TOKEN Permissions` in your workflows @cilindrox  ! 🙏

ViacheslavKudinov:
Not sure I'm getting everything right, but there is default what token has if nothing is set on workflow.
If something is set everything else that is not mentioned becomes  `none`, means no permissions.

PS its about standard GITHUB_TOKEN, not a custom token which can have what ever scope that was granted.

jetersen:
@cchanche sorry very long thread :sweat_smile: 

If you need help debugging with GitHub Cloud enterprise. I can help.
What are steps for the repro?

Also maybe the answer is we should just switch away from the GraphQL? If there is some git steps that could give us the same state :thinking: I know it a topic that's come up at least and now it comes up as a permission issue due to the usage of associatedPullRequests graphql query.

ViacheslavKudinov:
I think its important to get permissions and way how workflow was used.
If `permissions` set on workflow level and job level, job level will override. 
Also as it is section with `workflow_call` then permissions on caller workflow also matters.
Too many aspects which affect behavior.

jetersen:
https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow

> If jobs.<job_id>.permissions is not specified in the calling job, the called workflow will have the default permissions for the GITHUB_TOKEN. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).
The GITHUB_TOKEN permissions passed from the caller workflow can be only downgraded (not elevated) by the called workflow.

https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#access-and-permissions-for-nested-workflows

> A workflow that contains nested reusable workflows will fail if any of the nested workflows is inaccessible to the initial caller workflow. For more information, see [Access to reusable workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#access-to-reusable-workflows).
> GITHUB_TOKEN permissions can only be the same or more restrictive in nested workflows. For example, in the workflow chain A > B > C, if workflow A has package: read token permission, then B and C cannot have package: write permission. For more information, see [Use GITHUB_TOKEN for authentication in workflows](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).
> For information on how to use the API to determine which workflow files were involved in a particular workflow run, see [Reuse workflows](https://docs.github.com/en/actions/how-tos/sharing-automations/reuse-workflows#monitoring-which-workflows-are-being-used).

Honestly pretty reasonable it all has to do with permissions. Just set the right permission the caller workflow. Annoying but needed.
Also lets focus on release-drafter issue and not how to use reusable workflows :sweat_smile:

cchanche:
> we should just switch away from the GraphQL

Y E S 👍 👍  this is a major issue as you say, and doing the same with the git cli should work. see https://github.com/release-drafter/release-drafter/pull/1408#issuecomment-4082227574. Since it kind of comes up more and more lately (see recent https://github.com/release-drafter/release-drafter/issues/1566#issuecomment-4099689290), I think i'll switch working on this first (before the conventional commit semantics). In https://github.com/release-drafter/release-drafter/issues/871#issuecomment-3944566391 I even stated this should be worked on before the commit support anyway. Actually still accurate IMO.

Now for the moment, I'd like to get to the bottom of @cilindrox 's issue. Release-drafter will keep using API calls no matter what, and knowing all about permissions on the GITHUB_TOKEN always helps.

Current thread helped (me at least) understand : 

- default permissions can be defined at repo and/or org level in the settings. If set to read & write, no need to specify anything in your workflows
- in gql, `pull_request: read` is required by the `associatedPullRequests` relation WHEN IT EXISTS
- Every GHA workflow outputs the tokens permissions scopes in the `Set up job` step for debugging
- `workflow_call`ed workflows cannot elevate inherited token permissions (thank you for clarifications @jetersen !)

ViacheslavKudinov:
I would like to believe that as snippet was updated with correct permissions we are good.
Its still unclear what is wrong with @cilindrox setup.

PR #1456, opened 2025-08-12
feat: support filter by base ref name

will4j:
this pr add two parameters (`exclude-base-ref-names` and `include-base-ref-names`) to support filter by base ref names, which can be used in case like only take account PR to specific branch like `main`. examples:

```
# file .github/release-drafter.yml
exclude-labels:
  - 'skip changelog'
  - 'reverted'
include-base-ref-names:
  - 'main'

```

Issue #1354, opened 2023-08-24
Edge-case: some PRs being omitted from release summary


ssolbeck:
I've discovered an oddity / edge-case of this action when determining the PRs included in a release, which results in some PRs being omitted from the release summary in certain situations.

The edge-case is due to the behavior of the `associatedPullRequests` request (doc-link is for the REST API but matches GraphQL API): https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-pull-requests-associated-with-a-commit

> Lists the merged pull request that introduced the commit to the repository. If the commit is not present in the default branch, will only return open pull requests associated with the commit.

release-drafter queries for the first 5 associated pull requests on each commit. In cases where release-drafter is used on a non-default branch, and PRs are long-lived and often updated via merge, the first 5 associated pull requests may include only other pending PRs, and not the merged PR which introduced the commit.

## Proposal 

Increasing the number of associated pull requests queried should avoid this issue. If there's any concern that this increase would impact the action's performance, we can implement it as a new optional config option, and keep the default limit at 5.

I'm filing this issue to gauge receptiveness before I write a PR here for the behavior. Please let me know -- I'd much rather contribute to this upstream project than keep a fork.

---

## Context

1. We are triggering the release-drafter action on updates to `staging`, a non-default branch.
2. Feature PRs are merged to `staging`, and may wait there for some period of time before being merged to `main`, the default branch.
3. While waiting for release, other open feature PRs may merge `staging` into their feature branches in order to resolve conflicts.
4. This means that the commits of a feature PR may be present in several other PRs before reaching `main`.

## Reproduction

1. Configure release-drafter to run on pushes to a non-default branch `staging`.
2. Create a feature branch `feature1` off of `staging`, make a commit on the branch, and open a PR back to `staging`.
  a. Repeat this four more times, for a total of 5 feature branches and PRs.
3. Create another feature branch `feature-in-release` off of `staging`, make a commit on the branch, and open a PR back to `staging`.
4. Merge this `feature-in-release` PR into `staging`.
5. Observe that the release-drafter action is triggered, and that the output release summary includes an entry for `feature-in-release` (as expected).
6. Update `feature1` by merging in `staging` (keeping this feature branch up to date)
  a. Repeat this four more times for the remaining feature branches.
  b. All five open feature branch PRs will now include the commits from `feature-in-release` (in addition to other commits).
[...]

Issue #637, opened 2020-09-03
Drafting a release from previous PRs


matewilk:
Hey, quick question as I haven't seen it in the description/docs.

Is there an automatic way to create release notes from my previous PRs when I didn't have `release-drafter` installed?

I'm looking at various options right now to use with github actions and this project looks promising.

The only issue is, I don't want to manually create the release changelog :)

lucacome:
For anybody stumbling upon this, release-drafter automatically adds old merged PRs (I guess since last tagged version?) when enabled.

PR #1570, opened 2026-03-22
feat: switch release discovery to ref comparison and explicit missing-baseline warnings

### LINKED ISSUE #1596: Allow truncation of release bodies
When adopting a release-based flow with release-drafter, the initial release may contain **very** many PR's (even when using `initial-commits-since`), too many to fit into a GitHub release.

This can make GitHub respond with `{"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"}`, and the release will fail.

For our use-case, it is perfectly fine to just truncate the message at 125 000 characters for the initial release.

jetersen:
## Summary

This PR changes how Release Drafter discovers the commits and pull requests that belong in a draft release.
 
Instead of walking commit history backward from a date (`since: lastRelease.created_at`), it now uses a git ref comparison between the previous published release and the current commitish. This makes changelog discovery follow the commit graph (ref to ref) instead of commit authored dates (similar to all-ancestors).
 
This PR also folds in the work from #1573: when Release Drafter cannot find a previous **published** release to use as a baseline, it no longer tries to infer an open-ended history window. It still creates a draft, but appends a clear warning to the release body and points users to a dedicated issue template if that missing baseline is unexpected.
 
## Root cause
 
`since: lastRelease.created_at` filters by commit authored date.

That breaks down when commits were authored long before they were merged, especially with rebase merges. In that case, a recently merged PR can be silently skipped because its commits predate the release timestamp.

Walking `Commit.history(...)` also makes the changelog boundary harder to reason about on long-lived branches, because the effective range is based on history traversal + dates rather than a concrete git comparison between two refs.

Using `Ref.compare(...)` makes the normal case much easier to reason about: the changelog is based on the commit graph between two SHAs, not on authored timestamps.

## What changed

- replaces date-based PR discovery with SHA/ref comparison
- switches path-filtered commit lookup to the same comparison-based boundary
- removes the `initial-commits-since` input/config/schema/docs
- appends a warning to the release body when no previous published release is found
- adds a dedicated issue template for unexpected “previous published release not found” cases
- introduces `executeGraphql` to simplify GraphQL call sites
- switches queries to typed-document-node generated documents for stronger type inference
 
## User-facing impact

- changelog generation is now based on the commit graph between refs, not commit authored dates
- rebased PRs are less likely to be silently skipped
- long-running / alternate branch history is less likely to leak unrelated PRs into the changelog
- when no previous published release exists, Release Drafter produces a draft with a warning instead of trying to backfill an open-ended baseline
- users can report unexpected missing-baseline cases through a dedicated issue template
 
### Fixes
 
- Fixes #1596 
 
### Closely related context
 
- supersedes #1456 — proposes base-ref-name filtering as a way to avoid unrelated PRs from long-lived branches; this PR addresses that problem earlier by tightening the commit range itself
- #1354 — associatedPullRequests edge case on non-default branches
- reopens #637 — drafting a first release from historical PRs; this PR intentionally moves the no-baseline case toward an explicit manual starting point instead of auto-including open-ended history

## Test plan
 
- new comparison fixtures cover merge commit, squash merge, rebase merge, forking, overlapping labels, empty, paginated, and no-baseline cases
- path-filtered commit discovery is tested against the new comparison-based stopping behavior
- existing schema / e2e coverage was updated to reflect the new comparison-only behavior


cchanche (commented):
What would be the equivalent `git log` or `git rev-parse` command to the new compare ? I'd like to fully understand the commits that are returned by this call.

Previously, we would simply return every reachable ancestor of `refs/heads/[commitish]`. This meant long-lived branches that were recently merged may also be traversed. This proved to be unwanted behavior for some (see #1456) but if this is to be non-breanking, it should be preserved.

From the docs, it feels like compare simply returns the (only) existing chain of ancestors from one commit to the other. This would exclude commits from alternate ancestors as stated above.

Let me know what you think ! @jetersen 

I have yet to look at your changes, maybe you covered this already.

jetersen:
As long as you compare against the base and head ref it would only find commits that are between those two SHA and that should include any PRs that.
This would supersede #1456 as the correct solution for this problem.

cchanche:
> As long as you compare against the base and head ref it would only find commits that are between those two SHA and that should include any PRs that. This would supersede #1456 as the correct solution for this problem.

Ah ok so that's what I thought ! I think that's a good thing to have as a default behavior, but it breaks the current one...

jetersen:
You could argue the old approach was always flawed

cchanche:
That's fair ! Would be a bit heavy to do `v8` just for that anyway... I'm ok with _type: feature_ then, but let's keep an eye on #1456's suggestion in case someone needs the old ways.

EDIT: #1456 does not help once this is merged, because we won't see those branches anymore (at all). But that's ok, we can find sth else when (if) needed

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
```diff
@@ -1,20 +1,111 @@
...
+        ...sharedComparisonParams,
+      })
+    } else {
+      // Case 3: no tags at all → cap-and-bail
+      core.info(
+        `No previous release or tag found. Fetching first page of commits (cap-and-bail)...`,
+      )
+      const capResult = await executeGraphql(
+        octokit.graphql,
+        FindCommitsWithAssociatedPullRequestsDocument,
```
cchanche: I am unsure how extracting this out of pagnination helps

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
```diff
@@ -1,20 +1,111 @@
...
+          after: null,
+          withPullRequestBody:
+            params.config['change-template'].includes('$BODY'),
+          withPullRequestURL: params.config['change-template'].includes('$URL'),
+          withBaseRefName:
+            params.config['change-template'].includes('$BASE_REF_NAME'),
+          withHeadRefName:
+            params.config['change-template'].includes('$HEAD_REF_NAME'),
+          pullRequestLimit: params.config['pull-request-limit'],
+          historyLimit: params.config['history-limit'],
```
cchanche: You only fetch first 15 commits no matter what `'initial-commits-since'` is here
```suggestion
```

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
```diff
@@ -1,20 +1,111 @@
...
+        `No previous release or tag found. Fetching first page of commits (cap-and-bail)...`,
+      )
+      const capResult = await executeGraphql(
+        octokit.graphql,
+        FindCommitsWithAssociatedPullRequestsDocument,
+        {
+          name: context.repo.repo,
+          owner: context.repo.owner,
+          targetCommitish: params.config.commitish,
+          since: params.config['initial-commits-since'],
```
cchanche: If :

> `initial-commits-since` remains as the only date boundary

and we keep a :

> single non-paginated fetch

Wouldn't it be better if `initial-commits-since` had a default value ?

I like the idea that you can draft your first release with all of your previous content since forever - but that requires pagination.

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
```diff
@@ -1,20 +1,111 @@
...
+      baseRef: `refs/tags/${params.lastRelease.tag_name}`,
+      ...sharedComparisonParams,
+    })
+  } else {
+    // Case 2: no previous release, look for most recent tag
+    const mostRecentTag = await findMostRecentTag({
+      name: context.repo.repo,
+      owner: context.repo.owner,
+      tagPrefix: params.config['tag-prefix'] || undefined,
+    })
```
cchanche: I think this should take in `commitish: string` and `filter-by-commitish: boolean` the same way `find-previous-release.ts` does

src/tests/drafter/drafter.e2e.test.ts
```diff
@@ -2910,7 +2931,7 @@ describe('drafter e2e', () => {
 
           ## Contributors
 
-          No contributors
+          @TimonVS
 
           ## Previous release
```
cchanche, line 2961: What's happened here ? Looks like a something got fixed ?

src/tests/drafter/drafter.e2e.test.ts
```diff
@@ -3277,11 +3301,10 @@ describe('drafter e2e', () => {
...
-            body.query.includes(
-              'query findCommitsWithAssociatedPullRequests',
-            ) && body.variables.since === '2018-06-29T05:45:15Z',
+            body.query.includes('query findCommitsInComparison') &&
+            body.variables.baseRef === 'refs/tags/v2.0.0',
         )
-        .reply(200, getGqlPayload('graphql-commits-no-prs'))
+        .reply(200, getGqlPayload('graphql-comparison-no-prs'))
 
       await runDrafter()
```
cchanche: This eliminates the original purpose of the suite. Maybe try to return no previous tags ?

cchanche (changes requested):

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Limit to the size of the pagination helped some user swith microsoft's unstable servers (see https://github.com/release-drafter/release-drafter/issues/1450#issuecomment-3436676585 and related)

I do think we should keep paginating everything, and with this `history-limit` input if possible

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Don't want to fetch a (more recent) tag from another branch I am not drafting a release for

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
jetersen: It becomes strongly typed by typescript we could problem drop a bunch of things and infer the type instead.

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
jetersen: it's meant as a fallback and history-limit is now input parameter so it could change? :thinking:

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
jetersen: this is only meant as a fallback if we can't find a release.

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
jetersen: properly but again only meant as a fallback if no release is found.

src/tests/drafter/drafter.e2e.test.ts
jetersen: We are actually correctly fetching data that was hidden by the old approach

jetersen: Digged a little more into a found the issue to be related to the change from date to SHA based, this is fixed in e4e46c7 but will most likely have to be detected differently for the conventional commits.

jetersen:
@cchanche maybe to explain it a bit I added the most recent tag as a way to fallback when there is no releases. This is properly incorrect. Even the cap and bail feels wrong.
Honestly I think we should just default to a release draft saying "first release :tada:" or something basic there are too many foot guns when doing the first release. One of them is you can have a very old repo with 1000 of commits or you can have a new repo with only 10 commits. Maybe we can handle that case but what's the right choice.

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
jetersen: is it the executeGraphql or is the AST generated graphql you did not like?

cchanche: No yeah so the new type inference is great ! My point is much simpler : this call used to be paginated and now it is not. Since you kept the `paginateGraphql` helper for `findCommitsInComparison`, I thought you may do the same here

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Yes, `Case 3: no tags at all → cap-and-bail` is quite clear - this is a fallback

Or do you mean `historyLimit` (the parameter) is a fallback ? What fallback would it be to ?

`history-limit` was introduced for people hitting 502 errors during pagination. This allowed having a smaller pagination window. (default `15`, not `undefined` !)

My point with this (these) comment(s) are two-fold : 

- I think we should paginate all of our big graphql queries (`FindCommitsInComparisonQueryVariables` & `FindCommitsWithAssociatedPullRequests`)
  - by _big_ I mean that they return a lot of stuff (relations, fields etc...)
  - I know and understand `FindCommitsWithAssociatedPullRequests` would only be called once on rare instances (but we have the code it takes at hand to make sure there's not issue)
- `history-limit` is related to pagination, and should be treated as such when we do pagination (and left-out if we don't)

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Yes it is

src/actions/drafter/lib/find-pull-requests/find-pull-requests.ts
cchanche: Yes I understand how this is a fallback, `Case 2: no previous release, look for most recent tag` says it all

But this looks like a straight bug to me (no offense). I use multiple release channels in some of my projects, some of which are isolated by a different comitish. If I want to add a new one for some reason (and start drafting releases), I'd like the current release-drafter code to fallback down to `Case 3: no tags at all → cap-and-bail` instead of returning a tag from another branch in `Case 2`

src/tests/drafter/drafter.e2e.test.ts
cchanche: Ah nice catch 👏

src/tests/drafter/drafter.e2e.test.ts
cchanche: > This eliminates the original purpose of the suite

...meaning the removal of `body.variables.since === '2018-06-29T05:45:15Z'` in favor of `body.variables.baseRef === 'refs/tags/v2.0.0'`

cchanche:
> I added the most recent tag as a way to fallback when there is no releases

Yes that part is pretty clear to me, with additional fallback to the original `all-ancestors` strategy as a last resort

> This is properly incorrect. Even the cap and bail feels wrong.

Am I understanding you feel this fallback strategy is wrong ? Maybe there's a few things in what you implemented here that could be implemented better, but the mainframe of falling back to tags and then to nothing or a date would be ok with me !

However, I also like to think we could indeed do much simpler and try to be less inclusive of every single scenario. What do you think of keeping only `Case 1: previous release with tag → comparison`, and then issuing an empty release when `params.lastRelease?.tag_name` is non-existant ? We could seize the opportunity to explain in that release's body what's happening, and guide users into manually crafting what they want - instead of trying to accommodate everyone and adding complexity (and potential bugs !) to the codebase.

As an example :

image

jetersen:
I love that example

cchanche:
> I love that example

see #1573

jetersen:
@cchanche I think something is just wrong in this picture :sweat_smile:
image

cchanche:
Yeah coverage badge conflicts are annoying... I really don't know what to do to prevent those except manually resolving the thing.

But anyway, now I am confused - you mean to merge this now ? I was under the impression something like #1573 (PR targeting this branch, not master) was more in line with what you implied [here](https://github.com/release-drafter/release-drafter/pull/1570#issuecomment-4112776046).

But since #1573 is still WIP - my time rarified lately - I'll be ok if you just want to merge for this (#1570) now and get things going.

jetersen:
Nah nah I just wanted to flag I did not like the badge :+1:
I feel like the action summary is enough. We could perhaps just have the percentage shown in github status check? Getting rid of the badge?

PR #1573, opened 2026-04-02
feat: append warnings to the release body when no previous published release exists

cchanche:
Prevents fetching endless commits and associated pull requests when no previous published release can be found, and used as a baseline to compare with current commitish's HEAD. This warning redirects to a newly introduced issue template that both guides user towards stating their issue, aswel helping them understand what has happened.


cchanche:
Still need to test & write/rewrite tests

src/common/parse-commitish.ts
```diff
@@ -0,0 +1,24 @@
...
+ * GITHUB_REF or the ref from webhook start with `refs/tags/`, so we handle
+ * those here. If it doesn't but is still a tag - it must have been set
+ * explicitly by the user, so it's fair to just let the API respond with an error.
+ */
+export const parseCommitishForRelease = (
+  commitish: ParsedConfig['commitish'],
+) => {
+  let mutableCommitish = structuredClone(commitish)
+
+  if (mutableCommitish.startsWith('refs/tags/')) {
```
jetersen, line 20: We have access to the commit sha from the tag so we could resolve the commitish to a sha instead of tag or branch?

cchanche: Yeah why not ! good idea

cchanche: I've mentioned how this commitish parsing thing could be done better in the code directly (with TODOs)

I'd like to open a dedicated issue&PR later-on if that's ok

PR #1590, opened 2026-04-10
fix(initial-commits-since): override via input not working

JGAntunes:
First of all thanks for your work 👍 

I was testing out the release drafter action and while doing a initial release I was trying to leverage the `initial-commits-since` property via the action inputs however I noticed it wasn't producing any effect (the repo is large so it was basically trying to pickup everything since the beginning of time I guess).

I believe this a bug, consequence of `initial-commits-since` not being set in - https://github.com/release-drafter/release-drafter/blob/8a1e7e7867896b504dede2455891b90cff66cc2b/src/actions/drafter/config/merge-input-and-config.ts#L15-L45

So it will never be hand over to the main execution function.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
       - master
 
 permissions:
-  contents: read
+  contents: write # Required for "Set coverage status"
   statuses: write
 
 jobs:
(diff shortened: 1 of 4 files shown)


cchanche:
Hey @JGAntunes , thank you for your suggestion ! 

Has your fork enabled what you were looking to accomplish ?

This is indeed an overlook/bug, and would need to be fixed, but we are in an active discussion to disable this behavior of fetching everything alltogether : see https://github.com/release-drafter/release-drafter/pull/1573

If you really need this in release-drafter's master branch, please address CI issues. If not, I suggest you draft your first release using your fork and switch back the official release channels. Feel free to share your thoughts on #1573 😉

JGAntunes:
Hey @cchanche 👋 thanks for the reply.

I was able to circumvent the issue altogether by adding the `initial-commits-since` to the config instead of the action input. This was just more about fixing a bug I knew about and giving back to the project. If y'all already have a different path you want to go through then I think we can close it 👍

PR #1591, opened 2026-04-13
Fixed initial-commits-since in config not being overwritten by input

sroebert:
Currently the `initial-commits-since` input variable is not overwriting the config value, resulting in it not being used. This PR fixes that by adding it to the merge function.

### DIFF
--- a/dist/actions/drafter/run.js
+++ b/dist/actions/drafter/run.js
@@ -957,6 +957,10 @@ var mergeInputAndConfig = (params) => {
 		if (config["filter-by-range"] && config["filter-by-range"] !== input["filter-by-range"]) info(`Input's filter-by-range "${input["filter-by-range"]}" overrides config's filter-by-range "${config["filter-by-range"]}"`);
 		config["filter-by-range"] = input["filter-by-range"];
 	}
+	if (input["initial-commits-since"]) {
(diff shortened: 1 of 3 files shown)


sroebert:
I realized it was already mentioned in this PR https://github.com/release-drafter/release-drafter/pull/1590, but I do need this feature to work.

I have a different way this can be solved as well, but I need release drafter to only look at commits since the publish date of the latest release. This is currently not possible, but I can work around it by using `initial-commits-since` and a version filter.

cchanche:
Hey @sroebert ! Are you sure `initial-commits-since` even works when previous releases exist (at all) ? I would need to read that code path again... but from what I recall, `initial-commits-since` is really only useful when there's absolutely no previous releases in the repo (as old as they may be)

sroebert:
It does work, because I add a filter that results in no previous releases, it is a bit of a hack. But I fetch my own previous release and take the publish date. I then set a filter that results in no old releases and set the `initial-commits-since` to the publish date to get the commits I want for the release notes.

sroebert:
I created a new PR with the correct branch naming, I guess that should also fix the missing label for the PR: https://github.com/release-drafter/release-drafter/pull/1593

PR #1593, opened 2026-04-22
fix: initial-commits-since in config not overwritten by input

sroebert:
Currently the initial-commits-since input variable is not overwriting the config value, resulting in it not being used. This PR fixes that by adding it to the merge function.

I created a new PR with correct branch naming, original was https://github.com/release-drafter/release-drafter/pull/1591

### DIFF
--- a/dist/actions/drafter/run.js
+++ b/dist/actions/drafter/run.js
@@ -941,6 +941,7 @@ var applyOverrides = (config, input) => {
 	applyBooleanOverride(config, input, "include-pre-releases");
 	applyBooleanOverride(config, input, "latest");
 	applyStringOverride(config, input, "filter-by-range");
+	applyStringOverride(config, input, "initial-commits-since");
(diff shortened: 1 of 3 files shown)


sroebert:
I updated to match the latest changes in the config overwrite code

cchanche (approved):
LGTM ! Please fix dist

sroebert:
Fixed it!

pnpm-lock.yaml
cchanche, line 1: Please don't use `pnpm` to install dependencies.

cchanche (changes requested):

pnpm-lock.yaml
sroebert: Ah damn, sorry missed that, this was my brain on autopilot as I use it in my work projects, will fix it soon

sroebert: done, hopefully ok now :)

cchanche (approved):

PR #1594, opened 2026-04-22
fix: clarify prerelease-identifier behavior and precedence in configuration

### LINKED ISSUE #1592: Bug: Explicit `prerelease: false` should not be overridden by config-level `prerelease-identifier`
I am using Release Drafter with a single repository-level configuration file that defines the release journey for the project, including labels, release notes, prerelease handling, and the prerelease identifier.

In my GitHub Actions workflow, the user can choose whether the run should create/publish a prerelease or a stable release via a `workflow_dispatch` input.

When I run the workflow for a stable release and pass:

```yaml
prerelease: false
```

Release Drafter still switches the release to a prerelease if `prerelease-identifier` is present in the config file.

The workflow logs show:

```text
Warning: You specified a 'prerelease-identifier' (rc), but 'prerelease' is set to false. Switching to true.
```

As a result, the action creates a prerelease even though the workflow explicitly requested a stable release.

## Current behavior

Given this config:

```yaml
# .github/release-drafter.yml
prerelease-identifier: rc

# other release-drafter config:
# categories
# version-resolver
# templates
# autolabeler
# etc.
```

And this workflow step:

```yaml
  with:
    publish: true
    prerelease: false
  env:
```

Release Drafter logs:

```text
Warning: You specified a 'prerelease-identifier' (rc), but 'prerelease' is set to false. Switching to true.
```

Then it creates a prerelease.

## Expected behavior

When `prerelease: false` is explicitly provided as an action input, Release Drafter should respect that value and create a stable release.

`prerelease-identifier` should only be used when `prerelease: true`, or when `prerelease` is not explicitly set.

In other words:

| `prerelease` input | `prerelease-identifier` config | Expected result |
|---|---|---|
| `true` | `rc` | prerelease |
| `false` | `rc` | stable release |
| unset | `rc` | current behavior could remain prerelease, if desired for backward compatibility |

## Why this matters

The current behavior makes it difficult to keep release policy in repository config while keeping release intent in CI/workflow inputs.

In my case:

- `prerelease-identifier: rc` is repository configuration.
- Whether the current run is a prerelease or a stable release is workflow/runtime input.
- I do not want to duplicate or dynamically manage `prerelease-identifier` in CI.
- I expect an explicit `prerelease: false` action input to take precedence over a config-level prerelease identifier.

This feels especially surprising because the warning says `prerelease` is set to false, but the action still changes it to true.

## Possible solutions

A few possible approaches:

1. Respect explicit `prerelease: false` when provided through action inputs.
2. Only auto-enable `prerelease` from `prerelease-identifier` when `prerelease` is undefined.
3. Add a new option such as:

```yaml
auto-enable-prerelease-from-identifier: true
```

with the current behavior as the default for backward compatibility.

## Workaround

The current workaround is to remove `prerelease-identifier` from the config file and pass it conditionally from the workflow only when creating a prerelease.

However, this moves release configuration into CI logic, which is not ideal for repository-level release policy and makes the workflow harder to maintain.

## Additional context

The current documentation says that `prerelease-identifier` automatically enables `prerelease` when not already set to true.

This issue is about the precedence behavior when `prerelease: false` is explicitly provided by the workflow.

neilime:
This PR fixes the prerelease override bug from #1592 and refactors the config merge logic for readability.

An explicit action input of `prerelease: false` now takes precedence over a config-level `prerelease-identifier`, so stable releases are no longer incorrectly turned into prereleases. The related docs were updated to reflect that `prerelease-identifier` only auto-enables prerelease mode when both values come from the same config location.

It also splits `mergeInputAndConfig` into smaller scoped helpers and generalizes the override helper typing, making the merge behavior easier to read and maintain.

Validation:
- focused unit tests for `mergeInputAndConfig`
- targeted e2e regression coverage for the prerelease scenario
- full project pipeline run

**Fix and refactoring are in distinct commits to facilitate code reviews**


neilime:
Hi @cchanche, can you please have a look at this PR? To be honest I'm waiting for this fix as it disrupts many of my workflows.

Thanks

src/actions/drafter/config/merge-input-and-config.ts
```diff
@@ -17,110 +38,127 @@ export const mergeInputAndConfig = (params: {
...
+  applyBooleanOverride(config, input, 'include-pre-releases')
+  applyBooleanOverride(config, input, 'latest')
+  applyStringOverride(config, input, 'filter-by-range')
+
+  return {
+    hasInputPrerelease: typeof input.prerelease === 'boolean',
+    hasInputPrereleaseIdentifier,
+  }
+}
+
```
cchanche: I understand how this function relies on `apply*` helpers mutating the config, but then it feels weird to also return the few fields needed for the release mode thing.

I'd like better if `applyReleaseModeValidation` is renamed to represent an override instead of a validation (because it does not throw), and is also incorporated into the `applyOverrides` directly, such and the latter can remove it's return statement.

Let me know what you think :)

cchanche (changes requested):
Thank you @neilime for the quick follow up with this

Your refactor in `merge-input-and-config.ts` is very welcome, and also quite well done IMO !

A very small change requests nonetheless, and we'll be ready to go

src/actions/drafter/config/merge-input-and-config.ts
neilime: I've made some changes, let me know if its better.

cchanche (approved):

src/actions/drafter/config/merge-input-and-config.ts
cchanche: Thank you !

PR #1604, opened 2026-05-02
feat: recover recently merged PRs missed by associated PRs lag

### LINKED ISSUE #871: Drafts with No Changes

jetersen:
## Summary

fixes #871

Adds a safety-net query that runs after the main GraphQL comparison to catch PRs whose merge commit is in the comparison range but whose `associatedPullRequests` index has not yet been updated by GitHub (a known lag for very recently merged PRs).

The fix adds `oid` to the comparison query to get real git SHAs, then runs a second GraphQL query — `repository.pullRequests(states: MERGED)` — which hits the PR table directly and is immediately consistent. Any PR whose `mergeCommit.oid` matches a comparison commit but wasn't found via `associatedPullRequests` is merged into the result.

### Scoping

- **`baseRefName` filter** — when `commitish` is a confirmed branch ref (`refs/heads/...`), the query is filtered to PRs merged into that branch so a busy non-release branch can't crowd out recent merges in the `first: 5` window. For bare values (e.g. `main`, `v1.2.3`) the filter is omitted and correctness relies on OID intersection. For `refs/tags/...` and `refs/pull/...` the safety net is skipped entirely.
- **Path-filter aware** — `include-paths` / `exclude-paths` reduce the comparison commit set; the OID set used by the safety net is built **after** that filtering so the recovery can't reintroduce PRs the user explicitly excluded.
- **Shared GraphQL fragment** — the PR field selection is now a `PullRequestFields` fragment used by both the comparison query and the recovery query, so codegen emits a single named type for the PR shape.

### Live reproduction

Set up a fresh test repo (`jetersen/release-drafter-lag-test`) and ran a tight loop: open PR → squash-merge → immediately invoke release-drafter (dry-run) from this branch's bundled `dist/`. Each iteration's merge→drafter delay was ~1.5s (limited by `gh pr merge` round-trip).

| Iteration | PR | Safety net triggered? |
|---|---|---|
| 1 | #1 | ✅ recovered #1 |
| 2 | #2 | ❌ index already current |
| 3 | #3 | ✅ recovered #3 |
| 4 | #4 | ✅ recovered #4 |
| 5 | #5 | ✅ recovered #5 |
| 6 | #6 | ✅ recovered #6 |

In 5 of 6 runs, GitHub's `associatedPullRequests` returned the merge commit with `nodes: []`, and the safety-net query recovered the PR via `mergeCommit.oid` matching the comparison's commit OIDs. Without this fix, those 5 drafter runs would have produced a release draft silently missing the just-merged PR.

### Tests

- Unit tests for `findRecentMergedPullRequests` (commit-OID matching, dedup against existing PRs, null `mergeCommit`, multiple recoveries, both filtered and unfiltered `baseRefName`).
- e2e test recovering a PR missing from `associatedPullRequests` via the direct PR query.
- e2e regression test verifying `include-paths` is honored during recovery.

## Test plan

- [x] Existing test suite: 339/339 passing
- [x] Live GraphQL API verified: `baseRefName: null` returns recent merged PRs across all branches; `baseRefName: "master"` filters to that branch; tag values and `refs/heads/<branch>` (qualified) silently match nothing — confirms the bare-name strip is required
- [x] Live reproduction in `jetersen/release-drafter-lag-test`: safety net triggered in 5/6 quick-merge iterations
- [x] Regression test confirmed to fail when the OID set is moved back to pre-path-filter position


src/actions/drafter/lib/find-pull-requests/graphql/pull-request-fields.fragment.gql
cchanche, line 1: That's a nice gql feature !

src/tests/drafter/drafter.e2e.test.ts
```diff
@@ -3400,4 +3400,91 @@ describe('drafter e2e', () => {
...
+          },
+        ]
+      `)
+
+      expect(scope.isDone()).toBe(true)
+      expect(gqlScope.pendingMocks().length).toBe(0)
+      expect(mocks.core.setFailed).not.toHaveBeenCalled()
+    })
+
+    // Regression: the OID set used to gate recovery must be built AFTER path
```
cchanche:
```suggestion
    // the OID set used to gate recovery must be built AFTER path
```

This comment looks to me like an artifact of your debugging process... ? You fixed that regression already, right ?

cchanche (approved):
Damn ! That's super great !

So as I understand this, `Commit() { associatedPullRequests() { ... } }` lags, but `pullRequest() { mergeCommit() { ... } }` does not ? 

Thank you for working on this

src/tests/drafter/drafter.e2e.test.ts
jetersen: Ya, just working through the code shown it had to be after the filter.

jetersen:
> So as I understand this, Commit() { associatedPullRequests() { ... } } lags, but pullRequest() { mergeCommit() { ... } } does not ?

At least my testing was based on calling release-drafter which confirmed the lag and that the associatedPullRequests had. Where pullRequest mergeCommmit had no lag.
We still want to use associatedPullRequests as it provides a cleaner association and does multiple commits when it comes to PR especially for rebase or merge commits and provides the commits history for the filtering.

However OID will point the merge commit sha or the rebase top most rebased sha or squash and merge sha result.

PR #1628, opened 2026-05-30
fix(version): bump initial version to 1.0.0 for premajor on first run

### LINKED ISSUE #1603: `premajor` on first run (no prior releases) produces `0.1.0-rc.0` instead of `1.0.0-rc.0`
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Add a description

When `premajor` is configured as the version increment and there are no prior releases, `$RESOLVED_VERSION` resolves to `0.1.0-rc.0` rather than the semantically correct `1.0.0-rc.0`.

This is a pre-existing behaviour from v6 (not a v7 regression) — `0.1.0-rc.0` is used as a safe default for all `pre*` types on first run.

Discovered while working on #1601 and #1602

spokodev:
Closes #1603.

`getVersionInfo` hardcoded the first-run reference version to `0.1.0-<id>.0` for every `pre*` increment whenever `prerelease-identifier` was set. For `premajor` that contradicts semver intent: a major bump from `0.0.0` should reach `1.0.0`, not stay at `0.1.0`. As @jrbeilke noted in #1603 this has been the behaviour since v6 and is independent of #1602.

The fix anchors the initial version to the requested bump axis:

- `premajor` → `1.0.0-<id>.0`
- `preminor` → `0.1.0-<id>.0` (unchanged)
- `prepatch` → `0.1.0-<id>.0` (unchanged)

`preminor` and `prepatch` keep their existing `0.1.0` anchor so the `$RESOLVED_VERSION` output for those increments is identical to today.

### Tests

`src/tests/drafter/get-version-info.test.ts`:

- Updated `uses premajor prerelease-identifier on first run` to assert `1.0.0-rc.0` plus the matching `$RESOLVED_VERSION_MAJOR/MINOR/PATCH/PRERELEASE` parts.
- Added `uses premajor prerelease-identifier on first run with custom prerelease identifier` covering a non-`rc` identifier on the `premajor` path.
- Added `keeps preminor on first run anchored to 0.1.0` and `keeps prepatch on first run anchored to 0.1.0` to lock in the no-change boundary for the other two increments.

RED on `master`: both `premajor` cases fail (`expected '1.0.0-rc.0' to deeply equal '0.1.0-rc.0'`).
GREEN on this branch: 25/25 in `get-version-info.test.ts`, full suite 394/394.

`npm run tsc:check`, `npm run lint`, and `npm run format:check` all clean.

### DIFF
--- a/src/actions/drafter/lib/build-release-payload/get-version-info.ts
+++ b/src/actions/drafter/lib/build-release-payload/get-version-info.ts
@@ -69,33 +69,25 @@ export const getVersionInfo = (params: {
         : _localIncrement
     referenceVersion = versionFromLastRelease
   } else {
-    // No prior release and no input version: start from 0.1.0.
-    // For prerelease-based increments (prepatch/preminor/premajor) with a
-    // prerelease-identifier, build the reference as "0.1.0-<identifier>.0" and
-    // use no_increment, so $RESOLVED_VERSION = "0.1.0-rc.0" — a prerelease *of*
-    // the initial version rather than a prerelease of the next incremented
-    // version (which semver's prepatch/preminor/premajor would otherwise
-    // produce, e.g. "0.1.1-rc.0").  This restores the v6 behaviour introduced
-    // in PR #1303.
-    if (
-      _versionKeyIncrement?.startsWith('pre') &&
-      config['prerelease-identifier']
-    ) {
-      _localIncrement = 'no_increment'
-      referenceVersion = new VersionDescriptor(
-        `0.1.0-${config['prerelease-identifier']}.0`,
-        {
-          preReleaseIdentifier: config['prerelease-identifier'],
-          tagPrefix: config['tag-prefix'],
-        },
-      )
-    } else {
-      _localIncrement = 'no_increment'
-      referenceVersion = new VersionDescriptor('0.1.0', {
-        preReleaseIdentifier: config['prerelease-identifier'],
-        tagPrefix: config['tag-prefix'],
-      })
+    // No prior release and no input version: anchor to a 0.0.0 baseline and let
+    // semver apply the resolved increment, so the first draft is canonical and
+    // matches semver's own behaviour from zero:
+    //   premajor → 1.0.0, preminor → 0.1.0, prepatch → 0.0.1
+    // (each carrying the prerelease suffix when an identifier is set), and the
+    // stable axes line up too (major → 1.0.0, minor → 0.1.0). Fixes #1603,
+    // where premajor on first run incorrectly produced 0.1.0-rc.0.
+    //
+    // A default stable bump (`patch`, the resolver's fallback) is promoted to
+    // `minor` so a first stable draft is 0.1.0 rather than 0.0.1, matching the
+    // long-standing "first release is 0.1.0" convention. Prerelease increments
+    // keep their axis, so an explicit prepatch yields 0.0.1-<id>.0.
+    if (_localIncrement === 'patch') {
+      _localIncrement = 'minor'
     }
+    referenceVersion = new VersionDescriptor('0.0.0', {
+      preReleaseIdentifier: config['prerelease-identifier'],
+      tagPrefix: config['tag-prefix'],
+    })
   }
 
   return {
--- a/src/tests/drafter/get-version-info.test.ts
+++ b/src/tests/drafter/get-version-info.test.ts
@@ -200,15 +200,15 @@ describe('versions', () => {
         "$NEXT_MAJOR_VERSION_MAJOR": "1",
         "$NEXT_MAJOR_VERSION_MINOR": "0",
         "$NEXT_MAJOR_VERSION_PATCH": "0",
-        "$NEXT_MINOR_VERSION": "0.2.0",
+        "$NEXT_MINOR_VERSION": "0.1.0",
         "$NEXT_MINOR_VERSION_MAJOR": "0",
-        "$NEXT_MINOR_VERSION_MINOR": "2",
+        "$NEXT_MINOR_VERSION_MINOR": "1",
         "$NEXT_MINOR_VERSION_PATCH": "0",
-        "$NEXT_PATCH_VERSION": "0.1.1",
+        "$NEXT_PATCH_VERSION": "0.0.1",
         "$NEXT_PATCH_VERSION_MAJOR": "0",
-        "$NEXT_PATCH_VERSION_MINOR": "1",
+        "$NEXT_PATCH_VERSION_MINOR": "0",
         "$NEXT_PATCH_VERSION_PATCH": "1",
-        "$NEXT_PRERELEASE_VERSION": "0.1.1-0",
+        "$NEXT_PRERELEASE_VERSION": "0.0.1-0",
         "$NEXT_PRERELEASE_VERSION_PRERELEASE": "-0",
         "$RESOLVED_VERSION": "0.1.0",
         "$RESOLVED_VERSION_MAJOR": "0",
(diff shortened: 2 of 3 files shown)


src/actions/drafter/lib/build-release-payload/get-version-info.ts
cchanche: The idea behind #1603 is that the default version (with no prior release) should be canonical and match semver's default behavior

Your suggested changes manually insert what you as an end user expects to see, which is not that bad, but I'd like to be more generic here, and fully rely on semver.

Original implementation focused on keeping v6 compatibility, which introduced some complexity.

I'd like if something like this could work : 

```ts
  let referenceVersion: VersionDescriptor
  if (versionFromInput.version) {
    _localIncrement = 'no_increment' // use that exact input version
    referenceVersion = versionFromInput
  } else if (versionFromLastRelease.version) {
    _localIncrement =
      _localIncrement?.startsWith('pre') &&
      versionFromLastRelease?.prerelease?.length
        ? 'prerelease' // simply increment the prerelease number (1.2.3-beta.6 -> 1.2.3-beta.7). TODO : issue a warning versionKeyIncrement is not honored
        : _localIncrement // otherwise, create the prerelease suffix and increment the version (prepatch, preminor, premajor)
    referenceVersion = versionFromLastRelease
  } else {
    // Keep _localIncrement intact (= versionKeyIncrement)
    // TODO : but make sure default stays "minor", such as first stable draft stays "0.1.0"
    referenceVersion = new VersionDescriptor(
      `0.0.0`,
      {
        preReleaseIdentifier: config['prerelease-identifier'],
        tagPrefix: config['tag-prefix'],
      },
    )
  }
```

This, to me, should both work and simplify things. Semver behaves like the following, and out code should too : 

```
> semverInc('0.0.0', 'prepatch', 'beta')
'0.0.1-beta.0'
> semverInc('0.0.0', 'preminor', 'beta')
'0.1.0-beta.0'
> semverInc('0.0.0', 'premajor', 'beta')
'1.0.0-beta.0'

> semverInc('0.0.0', 'prerelease', 'beta')
'0.0.1-beta.0'
> semverInc('0.0.1-beta.6', 'prerelease', 'beta')
'0.0.1-beta.7'
```

src/tests/drafter/get-version-info.test.ts
```diff
@@ -267,10 +268,62 @@ describe('versions', () => {
...
+      lastRelease: undefined,
+      input: {},
+      config: {
+        'version-template': '$MAJOR.$MINOR.$PATCH$PRERELEASE',
+        'prerelease-identifier': 'rc',
+      },
+      versionKeyIncrement: 'prepatch',
+    })
+
+    expect(versionInfo.$RESOLVED_VERSION).toEqual('0.1.0-rc.0')
```
cchanche: Shouldn't a prepatch be '0.0.1-rc.0' instead ?

```suggestion
      versionKeyIncrement: 'prepatch',
    })

    expect(versionInfo.$RESOLVED_VERSION).toEqual('0.0.1-rc.0')
```

cchanche (changes requested):
Thank you @spokodev for looking into this !

spokodev:
Reworked to rely on semver from a `0.0.0` baseline, per your suggestion — thanks for the clear direction.

The first-run branch now just anchors `referenceVersion` to `0.0.0` and applies the resolved increment, so the canonical results fall straight out of semver:

- `premajor` → `1.0.0-id.0`
- `preminor` → `0.1.0-id.0`
- `prepatch` → `0.0.1-id.0` (your line-323 point — updated the assertion and the test title)
- stable `major`/`minor` → `1.0.0` / `0.1.0`

For the TODO you left: a default stable bump (`patch`, the resolver's fallback) is promoted to `minor` in that branch, so a first stable draft stays `0.1.0` rather than `0.0.1`. Prerelease axes keep their increment, which is why prepatch stays patch-level.

One consequence worth flagging: the `$NEXT_MINOR`/`$NEXT_PATCH`/`$NEXT_PRERELEASE` outputs on first run now derive from `0.0.0` too (so `0.1.0` / `0.0.1` / `0.0.1-0` instead of `0.2.0` / `0.1.1` / `0.1.1-0`). That's a direct result of the single `0.0.0` reference, and reads as more canonical, but flagging in case you'd want it otherwise.

Tests updated accordingly, full suite green (394), dist rebuilt. The old per-axis hardcoding block is gone, so it's a net simplification.

src/tests/drafter/get-version-info.test.ts
spokodev: Done — prepatch now resolves to 0.0.1-rc.0, updated the assertion and title accordingly.

spokodev:
Reworked along the lines you suggested — thanks for the clear direction.

First run (no prior release, no input version) now anchors to a `0.0.0` baseline and lets semver apply the resolved increment, instead of hardcoding `0.1.0`:

| increment | `$RESOLVED_VERSION` |
|---|---|
| `premajor` | `1.0.0-rc.0` |
| `preminor` | `0.1.0-rc.0` |
| `prepatch` | `0.0.1-rc.0` |
| `prerelease` | `0.0.1-rc.0` |

The stable axes line up too (`major` → `1.0.0`, `minor` → `0.1.0`).

The only deviation from a pure passthrough is the `patch → minor` promotion you flagged in the TODO: the resolver's default stable increment is `patch`, so without it a first *stable* draft would be `0.0.1`. Promoting it keeps the long-standing "first release is `0.1.0`" convention, while prerelease increments keep their own axis (so an explicit `prepatch` still yields `0.0.1-rc.0`).

I also rebased onto `master` and rebuilt `dist/` with npm — the diff is now just the three relevant files; an earlier build had pulled in some unrelated pnpm-path churn.

jetersen:
:nose: :robot: :thought_balloon:

spokodev:
lol yeah, man — got carried away explaining. all of cchanche's points are in, prepatch's 0.0.1-rc.0 now ^_^

src/actions/drafter/lib/build-release-payload/get-version-info.ts
```diff
@@ -69,33 +69,25 @@ export const getVersionInfo = (params: {
...
+    // stable axes line up too (major → 1.0.0, minor → 0.1.0). Fixes #1603,
+    // where premajor on first run incorrectly produced 0.1.0-rc.0.
+    //
+    // A default stable bump (`patch`, the resolver's fallback) is promoted to
+    // `minor` so a first stable draft is 0.1.0 rather than 0.0.1, matching the
+    // long-standing "first release is 0.1.0" convention. Prerelease increments
+    // keep their axis, so an explicit prepatch yields 0.0.1-<id>.0.
+    if (_localIncrement === 'patch') {
+      _localIncrement = 'minor'
     }
```
cchanche, line 86: I feel like this smells

cchanche (changes requested):

src/actions/drafter/lib/build-release-payload/get-version-info.ts
spokodev: yeah, that block is the odd one out. thing is, it's wedged between your two asks: pure semver from `0.0.0` makes a default first release `0.0.1`, but we also wanted to keep the old "first release is `0.1.0`" convention, and the only way there is to nudge a default `patch` up to `minor`, which is exactly the manual insert you (rightly) don't love. so it's your call: keep the `0.1.0` convention (i'll pull the nudge into a named helper so it stops reading like a magic `if`), or go fully canonical and let a default first release land as `0.0.1`. small change either way, just say which behaviour you want and i'll push it.

PR #1631, opened 2026-06-06
fix: honor a major version-resolver label on a first release

### LINKED ISSUE #1630: First release does not honor a major (breaking) label — resolves to v0.1.0 instead of v1.0.0
### What happened

On a repository with **no prior release or tag**, the drafted version resolves to `v0.1.0` even when
one of the included pull requests carries a `version-resolver: major` label. A major-labeled PR should
produce a `v1.0.0` first release; instead the draft comes out as the minor bump.

### Config (trimmed)

```yaml
version-resolver:
  major:
    labels: ['breaking']
  minor:
    labels: ['enhancement']
  patch:
    labels: ['fix', 'refactor']
  default: patch
categories:
  - title: 💥 Breaking Changes
    labels: ['breaking']
  - title: 🚀 Features
    labels: ['enhancement']
```

### Steps to reproduce

1. A repo with no published releases and no tags.
2. Merge a PR labeled `enhancement` (minor) and a PR labeled `breaking` (major) into the default branch.
3. Release Drafter runs and drafts a release.

### Expected

`$RESOLVED_VERSION` = `1.0.0` — the highest-precedence bump among the included PRs is `major`.

### Actual

`$RESOLVED_VERSION` = `0.1.0` (the minor bump). The `breaking` PR is listed correctly under its
category, so the label is recognized for notes, but the major bump is not applied to the version.

### Notes

- Not a stale-draft artifact: deleting the draft and regenerating from scratch reproduces `v0.1.0`.
- The `breaking` label is present in both `version-resolver.major` and a `categories` entry.
- Environment: `release-drafter/release-drafter@v6`.
- Related but not the same: #1377 (closed) concerns repos with more than 1000 releases; this is the
  no-prior-release / first-release case.

### Question

Is a `major` first release supported, or is seeding an initial tag required? If it is supported, the
version resolver does not appear to apply the `major` bump on the first release.

Bugs5382:
## Summary

On a repository with no previously published release, a first release never honors a `major`
version-resolver label: a `breaking`-labeled PR drafts `v0.1.0` instead of `v1.0.0`, and on a fresh
repo the draft is empty ("No changes") even though merged PRs exist.

Fixes #1630.

## Root cause

Two compounding issues on the no-baseline (first release) path:

1. `findPullRequests` short-circuits when there is no published release/tag, returning no pull
   requests before any label is read — so the version-resolver increment always falls back to the
   default (`patch`) and the breaking PR never appears in the notes:

   ```ts
   if (!params.lastRelease?.tag_name) {
     core.warning('A previous (published) release is required to find changes')
     return { commits: [], pullRequests: [] }
   }
   ```

2. Even when the increment is `major`, `getVersionInfo` seeded the no-baseline reference at the
   hardcoded `0.1.0` and forced `no_increment`, discarding the resolved increment — so a first
   release always resolved to `0.1.0`.

## Fix

- **`get-version-info.ts`** — when there is no prior release, no version input, and the resolved
  increment is `major`, seed the reference at `0.0.0` and apply the increment so `$RESOLVED_VERSION`
  becomes `1.0.0`. The `0.1.0` minor/patch baseline and the prerelease-identifier first-run path are
  unchanged.
- **`find-pull-requests.ts`** — on the no-baseline path, instead of returning empty, scan the most
  recently merged pull requests so their labels drive categorization and version resolution. The
  warning banner is unchanged.
- **`find-recent-merged-pull-requests.ts`** — extracted the shared GraphQL query into a helper; the
  existing lagging-index recovery (`findRecentMergedPullRequests`) is behavior-identical; added
  `findRecentMergedPullRequestsWithoutBaseline`, which returns recent merged PRs without the
  commit-OID intersection (there are no comparison commits on a first release).

Result: a `breaking` PR on a fresh repo resolves to `major` -> `v1.0.0`, with the PR listed.

## Tests

- Updated the "no past releases" e2e tests (which previously asserted the short-circuit: no PR query,
  "No changes") to assert that recent merged PRs are now scanned.
- Added an e2e test: no baseline + a `major`-resolver-labeled merged PR -> drafted `v1.0.0` with the
  PR listed (new config + GraphQL fixtures).
- Added a `getVersionInfo` unit test for the major-first-release case.
- `npm run all` (lint, typecheck, full suite, schema, build) passes; the committed `dist/` is rebuilt.

## Notes

- The extra recent-merged-PR query runs **only** when there is no published baseline (bounded by
  `RECENT_PR_LOOKBACK`); steady-state runs are unaffected.
- The lookback scans the most recently updated merged PRs, so a large first-release backlog could
  exceed the window; this reuses the existing recovery constant rather than introducing a new one.
- On a first release there are no comparison commits, so path-based pre-include/pre-exclude
  conditions are inert (`matchedPaths` is empty); label-based categorization and version resolution
  are unaffected.

### DIFF
--- a/dist/actions/drafter/run.js
+++ b/dist/actions/drafter/run.js
@@ -4375,15 +4375,11 @@ var FindRecentMergedPullRequestsDocument = {
 //#region src/actions/drafter/lib/find-pull-requests/find-recent-merged-pull-requests.ts
 var RECENT_PR_LOOKBACK = 5;
 var findRecentMergedPullRequests = async (params) => {
-	const octokit = getOctokit();
 	const nameWithOwner = `${context.repo.owner}/${context.repo.repo}`;
-	const missingPRs = ((await executeGraphql(octokit.graphql, FindRecentMergedPullRequestsDocument, {
-		name: context.repo.repo,
-		owner: context.repo.owner,
+	const missingPRs = (await queryRecentMergedPullRequests({
 		baseRefName: params.baseRefName,
-		limit: RECENT_PR_LOOKBACK,
-		...params.fieldFlags
-	})).repository?.pullRequests.nodes ?? []).filter((pr) => {
+		fieldFlags: params.fieldFlags
+	})).filter((pr) => {
 		if (!pr?.mergeCommit?.oid) return false;
 		const prKey = `${nameWithOwner}#${pr.number}`;
 		return params.commitOids.has(pr.mergeCommit.oid) && !params.foundPrKeys.has(prKey);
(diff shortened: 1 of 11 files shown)


cchanche:
Closed from https://github.com/release-drafter/release-drafter/issues/1630#issuecomment-4761281533

Issue #1396, opened 2024-06-12
_extends seems to append autolabeler list instead of overriding it


ssbarnea:
I discovered that when using `_extends:`, if you define an `autolabeler` section in the child config, that one will also get all the entries from the parent one instead of overriding them.

That is problematic at it means that there is no way to override the behavior from the centralized file, all the other normal fields do work as overrides.

cchanche:
Here is how probot merges configs : 

https://github.com/probot/octokit-plugin-config/blob/14e9b829fd3154c40e2d5cf2e7da41cab223ffe3/src/compose-config-get.ts#L34

```ts
Object.assign({}, defaults, ...configs)
```

I am unsure how this would work out with `autolabeler` exactly, but this looks like a shallow merge to me (and your scenario is a bug indeed)

Do you still have some reproduction material at hand ?

PR #1642, opened 2026-07-06
feat: support per-key merge strategies in `_extends`

ReneWerner87:
## Problem

`_extends` currently composes configuration files with a shallow merge. That is the right default for scalar values, but it means a repository cannot add entries to an inherited list without copying and replacing the entire list.

This is especially limiting for shared `categories` configurations. For example, a repository extending an organization-wide config may only need to add a `pre-include` for one module or a `pre-exclude` for a `skip-changelog` label, but today it must duplicate every inherited category. Those copies can drift from the shared config over time.

## Solution

Add an opt-in, per-key merge strategy to a new mapping form of `_extends`:

```yml
# your-org/.github: .github/release-drafter.yml
categories:
  - title: '🚀 Features'
    when:
      label: feature
```

```yml
# your-repo: .github/release-drafter.yml
_extends:
  from: your-org/.github
  strategy:
    categories: append
categories:
  - type: pre-exclude
    when:
      label: skip-changelog
```

The composed `categories` list contains the inherited entries followed by the repository-specific entry.

The existing string form remains supported and keeps its current shallow-override behavior:

```yml
_extends: your-org/.github
```

## Merge behavior

- `override` is the default and preserves the existing behavior.
- `append` combines the inherited list followed by the extending file's list.
- `prepend` combines the extending file's list followed by the inherited list.
- Strategies work for any list-valued config key, including `categories`, `autolabeler`, and `replacers`.
- A strategy applies only while its declaring file is merged onto the configuration it extends; strategies are not inherited by the next file in a multi-level chain.
- `append` and `prepend` reject present non-list values with an error that identifies the key and config file. Missing and YAML-empty values are treated as empty lists.
- A strategy declared for a key that the same file does not set has no effect and emits a warning.
- `_extends` is removed from the composed result before action-specific config validation.

This remains explicit rather than changing `_extends` to deep-merge by default. Implicit list appending previously caused #1396, where users could not override inherited autolabeler entries, and changing the default would alter existing configurations.

## Parsing and validation

As reworked in `63d0a5f`, every raw YAML or JSON config file is parsed through a common, loose Zod config-file schema immediately after it is loaded, before `_extends` targets are resolved or the extension chain is composed.

That early parsing step:

- validates `_extends` in the context of the config file where it was declared and includes that file in validation errors;
- normalizes the string and mapping forms into the mapping shape used by config loading and composition;
- keeps all action-specific fields intact for composition and their existing later validation;
- accepts absent, `null`, and blank-string `_extends` values as backward-compatible no-ops;
- treats an empty YAML `strategy:` value as no strategies;
- rejects a missing or blank `from`, unknown mapping keys, non-mapping strategies, and strategy values other than `override`, `append`, or `prepend`.

The same Zod `_extends` schema is also used to generate the editor-facing JSON schemas. This makes the runtime parser and generated schemas share one source of truth instead of maintaining separate manual validation logic.

Older Release Drafter versions only understand the string form and fail loudly when given the mapping form, avoiding a silent fallback to override behavior.

## Implementation

- `mergeConfigChain` replaces the previous `Object.assign` composition and lives in a dedicated module.
- Config files are folded base-first so per-file strategies compose correctly across multi-level `_extends` chains.
- The generated Drafter and Autolabeler schemas expose both forms of the optional `_extends` key.
- Merge-strategy documentation lives in `docs/configuration-loading.md` alongside the existing `_extends` documentation.

## Testing

- 25 focused tests cover merge ordering, defaults, explicit override, multi-level chains, strategy locality, missing and empty values, normalization, validation failures, `_extends` removal, unused strategies, and keys that collide with `Object.prototype`.
- An end-to-end `composeConfigGet` test verifies list appending through a real `_extends` chain.
- JSON Schema tests verify that `_extends` is optional and that both its string and mapping forms are generated correctly.
- Generated schemas and `dist/` are rebuilt.


jetersen:
Seems like the wrong fix.
Simply fix it to be deep merge as autolabal is a non issue?
Or can consider a controllable merge strategy with override, append.

ReneWerner87:
Thanks for taking a look!

I would advise against a plain deep merge as the default: it flips the behavior that was reported as a bug in #1396 (autolabeler entries appending with no way to override), silently changes the composed result for every existing `_extends` setup, and would need a major release. The probot ecosystem went through this cycle already: the old `probot-config` deep-merged, and its successor `@probot/octokit-plugin-config` deliberately moved back to a shallow `Object.assign`, arguing no single deep-merge strategy is universally right.

The controllable merge strategy is the right generalization, and I am happy to rework the PR in that direction: a `_merge` key in the extending file that selects the strategy per list key, defaulting to the current override behavior so nothing changes for existing configs:

```yml
_extends: your-org/.github
_merge:
  categories: append
categories:
  - type: 'pre-exclude'
    when:
      label: 'skip-changelog'
```

Semantics I would implement:

- Applied in `composeConfigGet` itself (a fold replacing the single `Object.assign`), so it works for any list key of both actions (`categories`, `autolabeler`, `replacers`) instead of being categories-specific. It also composes correctly across multi-level `_extends` chains, which the current `categories-append` key cannot (as a shallow-merged key it collides at chain depth 2).
- Each file's `_merge` governs only its own merge step and is not inherited; strategies are `override` (default) and `append`.
- `_merge` is stripped before schema validation, same lifecycle as `_extends`; `append` on a present non-array value is a hard error naming the key and the file.
- One caveat to document: released versions strip unknown keys silently, so configs using `_merge` need a minimum action version note in the README.

Does this shape (per-key strategy under `_merge` in the extending file, default `override`) match what you had in mind, or would you prefer a single global switch?

ReneWerner87:
@jetersen WDYT about the last comment?

cchanche:
I do admit this `_extends` situation was not on my mind during #1558 , and I also feel like categories are precisely "_the most common sharing setup_" like you say.

About the original suggestion :

> Seems like the wrong fix.

I agree

> The probot ecosystem went through this cycle already: the old `probot-config` deep-merged, and its successor `@probot/octokit-plugin-config` deliberately moved back to a shallow `Object.assign`, arguing no single deep-merge strategy is universally right.

Let's keep `Object.assign` as a default, and not reinvent the wheel here.

> The controllable merge strategy is the right generalization

This sounds better to me !

About the syntax you suggest, I'd like better if everything stays under `_extends` (instead of introducing top-level `_merge`).

```yaml
# Overrides parent categories
_extends: your-org/.github
categories:
  - type: 'pre-exclude'
    when:
      label: 'skip-changelog'
```

```yaml
# appends parent categories to the current oned
_extends: 
  from: your-org/.github
  strategy:
    categories: append # the default being 'override'
categories:
  - type: 'pre-exclude'
    when:
      label: 'skip-changelog'
```

Keep in mind categories are evaluated in the order they are defined in (see [readme](https://github.com/release-drafter/release-drafter#categorize-changes)). We need to maintain a certain level of control about how the categories are merged together. Maybe `append` is enough, maybe not. I am unsure. So I though adding a `prepend` strategy could help without too much of an extra cost here.

ReneWerner87:
Reworked as suggested, the PR now implements exactly this shape:

```yml
_extends:
  from: your-org/.github
  strategy:
    categories: append # or prepend; the default is override
categories:
  - type: 'pre-exclude'
    when:
      label: 'skip-changelog'
```

- The plain string form of `_extends` keeps working unchanged, and `Object.assign`-style override stays the default for every key.
- `prepend` is included: since categories are evaluated in the order they are defined, the strategy decides where the extending file's entries land (`append` after the inherited ones, `prepend` before them, which also covers `exclusive` categories that must run first).
- The strategy is resolved in the composition fold itself, so it works for any list key of both actions (`categories`, `autolabeler`, `replacers`) and composes across multi-level `_extends` chains; a file's strategy governs only its own merge step and is not inherited.
- Validation is strict and names the offending file: unknown strategy values, unknown keys in the mapping form, a missing `from`, and `append`/`prepend` on non-list values are errors.
- A nice side effect of anchoring the syntax under `_extends`: older action versions fail loudly on the mapping form instead of silently falling back to override.

PR title, description, docs (`README` and `docs/configuration-loading.md`), and tests are updated accordingly; `schema.json` needed no changes since `_extends` is stripped before schema validation.

src/common/config/index.ts
cchanche, line 1: Please keep this `index.ts` file a literal index, and relocate `mergeConfigChain()` elsewhere like you did with `parse-extends.ts`

README.md
```diff
@@ -426,6 +426,54 @@ categories:
       label: "dependencies"
 ```
 
+### Extending categories from a shared config
+
+Configuration files pulled in via [`_extends`](docs/configuration-loading.md#extend-other-config-files-using-_extends)
+are merged shallowly: a `categories` key in the extending file replaces the
+inherited list entirely. To add categories on top of a shared list without
+repeating it, use the mapping form of `_extends` and opt into a merge
```
cchanche: Please keep everything about `_extends` in a single place (i.e. `configuration-loading.md`)

This is advanced stuff, and I feel It should not clutter the main README more than it already is.

cchanche (dismissed):
LGTM ! Two small refactoring requests

src/common/config/index.ts
ReneWerner87: Done in 70e22a8: moved `mergeConfigChain()` and its `toMergeableList` helper into a dedicated `merge-config-chain.ts`. `index.ts` now only holds `composeConfigGet`, which imports it.

README.md
ReneWerner87: Done in 70e22a8: removed the README section. The full `_extends` merge-strategy documentation now lives only in `docs/configuration-loading.md` under "Merge strategies for `_extends`".

ReneWerner87:
@cchanche fixed

jetersen:
I don't see any schema change for `_extends` :thinking: Maybe it should be included now that it is more than just a string.

I think `_extends` should be part of the generated JSON schema, since the schema is editor-facing for raw user YAML even though `_extends` is stripped before final config validation.

Could we add a raw/load schema for it, e.g.

```ts
const mergeStrategySchema = z.enum(MERGE_STRATEGIES)

const extendsDeclarationSchema = z.string().min(1).or(
  z.object({
    from: z.string().min(1),
    strategy: z.record(z.string(), mergeStrategySchema).optional(),
  }),
)
```

Then include `_extends: extendsDeclarationSchema.optional()` in the schema used for `schema.json` generation, while keeping it out of the final parsed `Config`.

Also, `parseStrategy` could avoid `value as Record<string, MergeStrategy>` by building a validated object:

```ts
type MergeStrategies = Partial<Record<string, MergeStrategy>>

const isMergeStrategy = (value: unknown): value is MergeStrategy =>
  MERGE_STRATEGIES.includes(value as MergeStrategy)
```

ReneWerner87:
Both addressed in 7e91b4b.

`_extends` is now part of the generated JSON schema for both actions (`schema.json`, `drafter/schema.json`, `autolabeler/schema.json`), optional and exposing both forms. The mapping form is strict (`additionalProperties: false`), so editors flag unknown keys the same way composition rejects them at runtime. I put the zod schema in a dedicated `src/common/config/extends.schema.ts` rather than next to the parser, so the runtime composition path does not pull in a schema it never uses; the parsed `Config` is unchanged since `_extends` is still stripped while the chain is composed. The mirror in `json-schema.test.ts` asserts all of this.

`parseStrategy` no longer casts: it uses the `isMergeStrategy` type guard and builds a validated `MergeStrategies` object. The lookup in `mergeConfigChain` follows the same type.

ReneWerner87:
@jetersen @cchanche

jetersen:
Refactored `_extends` handling to use the Zod schema as the single source of truth instead of maintaining separate manual parsing and JSON Schema definitions.

Each raw config file is now parsed through a loose Zod config schema before resolving the extension chain. This normalizes `_extends` while preserving action-specific fields for composition and later validation. It also keeps the existing compatibility behavior:

- `override` remains the default strategy.
- Missing or null strategies normalize to an empty strategy map.
- Empty or null `_extends` values remain no-ops.
- Invalid mapping keys and strategy values are rejected by Zod.

This removes `parse-extends.ts` and prevents runtime validation from drifting away from the generated editor-facing schema. The unrelated Zod registry `id` is also excluded from the generated autolabeler JSON Schema.

jetersen:
@ReneWerner87 If you disagree with my change let me know.

jetersen (approved):
LGTM

cchanche (approved):
Streamlining everything under zod is a nice catch

I get that you then had to shift schema validation up earlier into the get-config-file process, but that should not break anything 😄

cchanche:
@ReneWerner87 please reflect this shift (earlier validation) in your PR descrition

jetersen:
Updated the PR description

PR #1649, opened 2026-07-14
feat: resolve release commitish refs to commit SHAs

jetersen:
## Summary

- resolve fully qualified tag commitish refs to commit SHAs with one rev-parse-compatible GraphQL query
- resolve pull request head and merge refs through the typed PullRequest GraphQL fields
- preserve the existing warning and default-branch fallback when a ref cannot be resolved
- document the expanded commitish behavior and regenerate the bundled action

## Dependency changes

None.

## Related work

Builds on #1648, which normalized fully qualified branch refs for the Releases API.

## Testing

- npm run all
- 424 tests passed

## Reviewer context

Tag resolution appends ^{commit}, so GitHub recursively peels lightweight, annotated, and nested annotated tags in one query. Pull request /head refs use headRefOid; /merge refs prefer potentialMergeCommit for open PRs and fall back to mergeCommit after merge. Short tag names remain unchanged because they are ambiguous with branch names.

> [!NOTE]
> ### Resolve release commitish refs to commit SHAs via GitHub GraphQL
> - `parseCommitishForRelease` in [`src/common/parse-commitish.ts`](https://github.com/release-drafter/release-drafter/pull/1649/files#diff-5780b7712bd1a3ad377ef15333fdf8e1ee6836ade2cc37c296d9d1afadde9d05) is now async and resolves fully qualified `refs/tags/{tag}` and `refs/pull/{number}/{head|merge}` refs to commit SHAs using two new GraphQL queries.
> - `refs/heads/{branch}` refs are normalized to plain branch names; non-prefixed values are passed through unchanged.
> - When `commitish` is a pull request merge ref (`refs/pull/{number}/merge`), the action forces dry-run mode and disables publishing, since merge refs point to ephemeral commits. A warning is emitted unless dry-run was already explicitly set.
> - `buildReleasePayload` and related callers in [`src/actions/drafter/main.ts`](https://github.com/release-drafter/release-drafter/pull/1649/files#diff-2b5ce5512473bee19507fbf5343f0ec5168e0884819eeb24653e2eadb8fcc2df) are updated to await the async resolution.
> - Risk: if tag or pull request ref resolution fails, the util logs a warning and returns an empty string, which causes the release to fall back to the default branch commitish.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 7174220. 9 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>
>


src/common/parse-commitish.ts
```diff
@@ -1,33 +1,118 @@
...
+  if (commitish.startsWith('refs/pull/')) {
+    const pullRequestRef = /^refs\/pull\/(\d+)\/(head|merge)$/.exec(commitish)
+
+    if (pullRequestRef) {
+      const [, pullRequestNumber, refType] = pullRequestRef
+
+      return resolvePullRequestToCommitSha({
+        octokit: octokit ?? getOctokit(),
+        pullRequestNumber: Number(pullRequestNumber),
+        refType: refType as 'head' | 'merge',
```
cchanche, line 103: I am unsure we can accept refs/pull/<PR-ID>/merge that easily. From what I can recall, this is a some kind of virtual _to-be-created_ merge-commit for github's internal stuff.

```zsh
➜  release-drafter git:(t3code/pr-coauthors-graphql) git ls-remote | grep 1649 
From git@github.com:release-drafter/release-drafter.git
240689e62d267fad37c1ae52ea1c6352b02090e0        refs/pull/1649/head
02fad3d7cfe039eca96a3d94e63af52683fcaadb        refs/pull/1649/merge
```

`02fad3d7` (/merge) does not correspond to any known commit locally, wheras `240689e6` (/head) is the last commit of the PR's branch (which is probably what we want here


Sadly, [PR trigger events](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request) use _refs/pull/PULL_REQUEST_NUMBER/merge_ as their `GITHUB_REF`. So we would need to transform this ourselves somehow

cchanche (dismissed):

src/common/parse-commitish.ts
jetersen: @macroscope-app can you confirm?

cchanche: Ah maybe pull/merge ref (as-is) are supported by the release api, that would make releases targeting a merge commit I believe...

jetersen: Ya it would allow people to use release-drafter on a PR and then have it take the release-drafter output straight into the PR as changelog.md or something similar :exploding_head:

jetersen: @macroscope-app Could that workflow work with this change? release-drafter running in an action as part of opening a PR to a release branch that bumps repo's version file. We could based on that generate release notes to the potential merged commit?

jetersen: Which could be done via commitish :thinking: they can just choose to use the target branch. At least this workflow would not be used for publishing a github relase but rather as the drafter during a PR.

jetersen: Perhaps we should prevent the virtual case or in case make it point to the target branch during PR.

cchanche: Yeah, resolving to the pr's branch instead sounds good

jetersen: So ya:
```yaml
    id: release_drafter
    with:
      dry-run: true
```

Should still have these ouutput available:

```
  ${{ steps.release_drafter.outputs.body }}
  ${{ steps.release_drafter.outputs.tag_name }}
  ${{ steps.release_drafter.outputs.name }}
  ${{ steps.release_drafter.outputs.resolved_version }}
```

jetersen: I actually think PR virtual ref makes sense cause it allows you to basically lock in the commitish not mix it makes a difference between two drafts that might end up having different release numbers. Basically the version bump pr can be held from the releasing changelog and used a changelog.md PR drafting section.

Basically allows for clear PR hotfixes in the release draft.

jetersen:
I think the feature is good to go.
It will handle virtual merge header safely and only be in output mode so the intended usage if when your doing a release branch and are able to land hotfixes into your release notes generate branch.

I tested the usage over here: https://github.com/jetersen/release-drafter-pr-merge-e2e/pull/21

PR #1682, opened 2026-07-25
fix: preserve proxy-aware fetch in octokit client

jetersen:
## Problem

`getOctokit` passed `request: { fetch: global.fetch }` with a comment saying it "allows nock to intercept requests in tests".

`@octokit/core` **shallow-merges** the `request` option, so supplying it at all replaces the proxy-aware `fetch` and agent that `@actions/github` builds from `http_proxy`/`https_proxy`. Proxy support was silently disabled — nothing errors, requests just bypass the proxy. That breaks GitHub Enterprise Server and corporate-proxy setups.

## Fix

- Drop the `request` override from `src/common/get-octokit.ts` so production keeps `@actions/github`'s proxy-aware `fetch`.
- Move the test-only fetch swap into `src/tests/setup.ts`: a mock of `@actions/github` injects `globalThis.fetch`, which nock can intercept (nock cannot intercept the `undici`-backed fetch, and `nock.disableNetConnect()` does not catch it either).
- Add `src/tests/get-octokit-proxy.test.ts`, which `vi.unmock`s `@actions/github` to assert the real production wiring — that the fetch is not a bare `globalThis.fetch`, and that the agent is present when `https_proxy` is set. This pins the behaviour so the override cannot be reintroduced unnoticed.

## Verification

`npm run all` — 440 tests pass across 23 files, lint/tsc/schemas/build clean.

> [!NOTE]
> ### Fix `getOctokit` to preserve proxy-aware fetch from `@actions/github`
> - Removes the explicit `request.fetch = global.fetch` override in [`get-octokit.ts`](https://github.com/release-drafter/release-drafter/pull/1682/files#diff-15d574e2e6cb01e4eb4587a2bbdc686e28ed7d257e098b963de18e4a6b95286f), allowing Octokit to use the default fetch/agent provided by `@actions/github`, which respects `https_proxy`.
> - Adds a test mock in [`setup.ts`](https://github.com/release-drafter/release-drafter/pull/1682/files#diff-4a7b05cd6cc30011e9e0f101e6058851da233e0b3debb84839d39452c8095642) that injects `globalThis.fetch` only in test environments so nock interception continues to work.
> - Adds a new test suite in [`get-octokit-proxy.test.ts`](https://github.com/release-drafter/release-drafter/pull/1682/files#diff-dd3e599d73b55840d2649838157cd7bf438d3592f5198d2d201678ef619d2c24) that verifies `request.fetch` is not `globalThis.fetch` and that `request.agent` is set when `https_proxy` is configured.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6bb49ce.</sup>
> 
>

### DIFF
--- a/dist/chunks/ignore.js
+++ b/dist/chunks/ignore.js
@@ -32447,17 +32447,10 @@ retry.VERSION = VERSION;
 //#endregion
 //#region src/common/get-octokit.ts
 var getOctokit = () => {
-	return getOctokit$1(process$1.env.GITHUB_TOKEN || "", {
-		log: {
-			...core_exports,
-			warn: warning
-		},
-		request: { 
-		/**
-		* Allows nock to intercept requests in tests
-		*/
-fetch: global.fetch }
-	}, paginateGraphQL, retry);
+	return getOctokit$1(process$1.env.GITHUB_TOKEN || "", { log: {
+		...core_exports,
+		warn: warning
+	} }, paginateGraphQL, retry);
 };
 //#endregion
 //#region src/common/config/get-config-file-from-repo.ts
--- a/src/common/get-octokit.ts
+++ b/src/common/get-octokit.ts
@@ -8,16 +8,17 @@ import {
 import { type RetryPlugin, retry } from '@octokit/plugin-retry'
 
 export const getOctokit = () => {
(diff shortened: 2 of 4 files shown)


cchanche (approved):
Nice catch ! I'd like to release that with 7.7.0 if that's ok w/ you

PR #1684, opened 2026-07-26
feat: add local release drafter CLI

### LINKED ISSUE #148: CLI Tool?

### LINKED ISSUE #414: Publish official container
Release-drafter should publish a prebuild container that can run release-drafter while receiving environment variables with credentials and repository name.

I think others already started working in this direction at https://hub.docker.com/r/onenashev/jenkins-release-drafter

It is key to allow users to run it and not to rely on github to run it for you, which may be convenient but comes with serious strings attached.

### LINKED ISSUE #1146: Multiple drafts created due to GH workflow concurency

### LINKED ISSUE #1386: Make action compatible with Gitea
Really want to use this action in my Gitea instance, but it's not compatible and it's trying to use the github API when I look at the runner logs, see below.

It might be easily solved if the action would use the git cli if it is available on the runner machine and falls back to the API method if the git cli is not available. Would this be possible? I believe the `checkout` action also uses this approach.

```
::warning::MyOrg/MyRepo: Invalid config file%0A{%0A  name: 'HttpError',%0A  id: '118',%0A  status: 401,%0A  response: {%0A    url: 'https://api.github.com/repos/MyOrg/MyRepo/contents/.github%252F.gitea%252Frelease-drafter.yml',%0A    status: 401,%0A    headers: {%0A      'access-control-allow-origin': '*',%0A      'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',%0A      connection: 'close',%0A      'content-length': '80',%0A      'content-security-policy': "default-src 'none'",%0A      'content-type': 'application/json; charset=utf-8',%0A      date: 'Wed, 24 Apr 2024 09:06:33 GMT',%0A      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',%0A      server: 'GitHub.com',%0A      'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',%0A      vary: 'Accept-Encoding, Accept, X-Requested-With',%0A      'x-content-type-options': 'nosniff',%0A      'x-frame-options': 'deny',%0A      'x-github-media-type': 'github.v3; param=raw',%0A      'x-github-request-id': 'F232:0F8A:31015FC:3148B4A:6628CB99',%0A      'x-ratelimit-limit': '60',%0A      'x-ratelimit-remaining': '59',%0A      'x-ratelimit-reset': '1713953193',%0A      'x-ratelimit-resource': 'core',%0A      'x-ratelimit-used': '1',%0A      'x-xss-protection': '0'%0A    },%0A    data: {%0A      message: 'Bad credentials',%0A      documentation_url: 'https://docs.github.com/rest'%0A    }%0A  },%0A  request: {%0A    method: 'GET',%0A    url: 'https://api.github.com/repos/MyOrg/MyRepo/contents/.github%252F.gitea%252Frelease-drafter.yml',%0A    headers: {%0A      accept: 'application/vnd.github.v3.raw',%0A      'user-agent': 'probot/12.2.5 octokit-core.js/3.5.1 Node.js/20.12.2 (win32; x64)',%0A      authorization: 'token [REDACTED]'%0A    },%0A    request: {}%0A  },%0A  stack: 'HttpError: Bad credentials\n' +%0A    '    at C:\\Users\\bvandevliet\\.cache\\act\\62485edbd9157925\\act\\actions\\release-drafter-release-drafter@v6.0.0\\dist\\index.js:8462:21\n' +%0A    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n' +%0A    '    at async Job.doExecute (C:\\Users\\bvandevliet\\.cache\\act\\62485edbd9157925\\act\\actions\\release-drafter-release-drafter@v6.0.0\\dist\\index.js:30793:18)',%0A  type: 'Error'%0A}
```

### LINKED ISSUE #1531: chore: discard CJS dependencies and produce side-effect-free bundle
See [discussion](https://github.com/release-drafter/release-drafter/pull/1475#discussion_r2902620016)

jetersen:
## Summary

- add a CAC and Consola-based CLI for drafting, updating, and publishing releases locally, including prerelease/latest controls and machine-readable `--json` output with logs preserved on stderr
- support SHA, branch, lightweight-tag, and annotated-tag ranges with optional `--from`, repository-default `--to`, automatic version resolution, shared config fallback, and dry runs
- expose a typed native-ESM API through `import { draftRelease } from 'release-drafter'` with explicit Octokit, GitHub context, logger, GHES, and publication options
- isolate public CLI/library and GitHub Actions graphs with one Vite 8 multi-environment build while retaining Toolkit proxy behavior in thin Action adapters
- use shared semantic emoji lifecycle logging and report the created or updated release URL
- replace `regex-parser` with a compatibility-tested local parser and assert public bundles contain no Actions runtime or CommonJS loader
- restore publishable npm metadata, generated NodeNext declarations, the TypeScript package marker, and OIDC `npm stage publish` through the protected `npm` environment; npm-only artifacts are built during publication while only Action bundles remain tracked
- mark generated distribution artifacts for GitHub Linguist without disabling diffs

## Testing

- `npm run all` (487 tests)
- `npm run coverage` (90.52%)
- isolated packaged NodeNext type/runtime consumer and `npx` CLI smoke tests
- public native-ESM and self-contained Action bundle scans
- manual dry runs for SHA and annotated-tag ranges, including an exact `v7.6.0` historical replay and captured Action/CLI output parity
- verified automatic version resolution remains independent from an explicit comparison base

Fixes #148
Fixes #414
Fixes #1146
Fixes #1386
Fixes #1531

> [!NOTE]
> ### Add local CLI and programmatic API for drafting GitHub releases outside of Actions
> - Adds a `release-drafter` CLI binary ([src/cli/run.ts](https://github.com/release-drafter/release-drafter/pull/1684/files#diff-dfa9baf011a9c38263c91af5d92745dd686bcc12a41d4974257c98da689589df)) with flags for `--from`, `--to`, `--config`, `--dry-run`, `--publish`, `--prerelease`, `--json`, etc., backed by a `draftRelease` programmatic API ([src/drafter.ts](https://github.com/release-drafter/release-drafter/pull/1684/files#diff-9d4d8cbab337ecd3bdb11746a26df7d8fba1ecca60cba25aa3aa9d513addd752)) that can be imported as an ESM module.
> - Introduces REST-only mode (auto-detected from API URL or `GITHUB_GRAPHQL_URL`) so the tool works on Gitea/Forgejo and other forges without a GraphQL endpoint; adds REST equivalents for commit comparison, PR lookup, and new-contributor detection.
> - Replaces all implicit `@actions/github` / `@actions/core` global state with an injected `GitHubContext` type carrying `octokit`, `repo`, `ref`, `serverUrl`, `logger`, and `restOnly` throughout the codebase.
> - Expands action outputs to include `target_commitish`, `previous_commitish`, `draft`, `prerelease`, `latest`, and `dry_run`; prereleases are now always emitted with `make_latest=false` regardless of config.
> - Adds container-based integration tests against real Gitea and Forgejo instances ([src/tests/drafter/forge-compat.container.test.ts](https://github.com/release-drafter/release-drafter/pull/1684/files#diff-e7657faf529c785460262e0a57f792cda8001a999a1b86c6c8a3320f7f3060e7)) and a separate CI job to run them.
> - The release workflow gains an npm publish job that stages the package to npm (using the `next` tag for prereleases).
> - Risk: `getPullRequestsChangedFiles` now requires an explicit Octokit parameter; `sharedInputSchema` no longer fails on a missing token at parse time (validation moved to `getOctokit`).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 03a520f.</sup>
> 
>
>


src/actions/drafter/lib/find-pull-requests/graphql/find-commits-in-comparison.gql
```diff
@@ -22,38 +23,64 @@ query findCommitsInComparison(
...
+    base: object(expression: $baseCommitish) @include(if: $useCommitishes) {
+      __typename
+      ... on Commit {
+        oid
+      }
+    }
+    head: object(expression: $headCommitish) @include(if: $useCommitishes) {
+      __typename
+      ... on Commit {
+        history(first: $historyLimit, after: $cursor) {
```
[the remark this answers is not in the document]
jetersen: Should be fixed in the last commit

src/drafter.ts
```diff
@@ -0,0 +1,91 @@
...
+    commits,
+    config,
+    input,
+    lastRelease,
+    previousCommitish: options.previousCommitish,
+    newContributorLogins,
+    pullRequests,
+    github,
+  })
+  const upsertedRelease = await upsertRelease({
```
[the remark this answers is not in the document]
jetersen: Fixed in 80ecc3c by routing the programmatic CLI through the shared drafter main flow. Pull request merge refs now force dry-run mode and disable publishing exactly as the action does, and the CLI reports the effective dry-run result.

cchanche:
This looks great ! What about addressing #1531 and restore publishing to npmjs at the same time ?

jetersen:
CJS is gonna be hard while depending on `@actions/github` and `@actions/core` cause they both pull in undici but we can make so the action is not included in the publishing.

cchanche:
Thx for looking it up !

jetersen:
:robot::runner::dash:

.github/workflows/release.yml
```diff
@@ -55,3 +50,24 @@ jobs:
...
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version-file: .node-version
+
+      - name: Install npm with staged publishing support
+        run: npm install --global npm@12.0.1
+
+      - name: Stage npm package
+        run: npm stage publish
```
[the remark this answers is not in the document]
jetersen: Fixed in 522fffd. The release step now has an `id`, exports its effective `prerelease` state through the build job, and the npm job stages prereleases with `npm stage publish --tag next`. The pushed tag is passed directly to Release Drafter, so downstream tag, major-version, and prerelease decisions use the Action outputs rather than a duplicate shell parser.

jetersen:
Went and updated the schema store as well https://github.com/SchemaStore/schemastore/pull/6135

jetersen:
@cchanche thought I tackle the `gitea` and `forgejo` support since GitHub isn't what it used to be and self hosting is gaining.

cchanche:
Hell yes !!!!

I thought about it recently, and thought gitlab support may be better to start with

I believe companies usually go for gitlab when they self host (including mine 👉👈)

Gitea however, should be easier to support since it's supposed to have similar apis (similar to GitHub) I recall.

In the end, choosing may be down to whatever we will be able to test with. Do you have a gitea instance at hand ?

And also, it might be wise to mirror the repo to public gitea/gitlab if we are to support them

jetersen:
The setup uses testcontainers, and we are testing it during CI! :see_no_evil:

cchanche:
Oh sorry yeah didn't take the time to read your entier changes yet, I had missed that (and the npmjs publishing)

I still have to understand the details, but does that mean gitea and forgejo have Github-compatible APIs ?

I would simply take the time to test the action in a real instance (meaning outside a test container), but otherwise, this looks like a great _low-hanging fruit_ situation here !

cchanche:
GitLab on the other hand will probably need a whole compatibility layer

jetersen:
Nah I have done gitlab testcontainers before!
But ya we would need a separate rest api although I happen to know a great gitlab library:
https://github.com/jdalrymple/gitbeaker I haven't used gitlab in years but I used to contribute to it.

jetersen:
Lets do the gitlab as a separate PR. Cause I think we will need to split the code further to allow different forge vehicles.

cchanche:
Sure !

README.md
```diff
@@ -39,6 +39,59 @@ jobs:
           config-name: release-drafter.yml # the default, loads '.github/release-drafter.yml'
 ```
 
+### ✍️ Local CLI
```
cchanche, line 43: Would avoid emojis 

```suggestion
### Local CLI
```

and also locate that section a bit lower

README.md
```diff
@@ -39,6 +39,59 @@ jobs:
...
+
+By default, the CLI creates or updates a draft. Use `--publish` to publish it,
+optionally with `--prerelease` or `--latest`:
+
+```sh
+npx release-drafter owner/repository --publish --prerelease
+npx release-drafter owner/repository --publish --latest
+```
+
+Each release-mode flag accepts an optional `true` or `false` value, so
```
cchanche, line 83:
```suggestion
The `--publish`, `--prerelease`, and `--latest` flags each accept an optional `true` or `false` value, so
```

vite.config.ts
cchanche, line 1: Vite is powerfull enough to accomodate what you are doing here, but I wonder if we should not try to pivot to monorepo with proper packages instead.

maybe something like :
- @release-drafter/core (the actual code)
- @release-drafter/autolabeler
- @release-drafter/cli (runtime wrapper for CLI)
- @release-drafter/gh-actions (runtime wrapper for Github Actions)
- @release-drafter/github-adapter (remote adapter for github)
- @release-drafter/gitea-adapter (remote adapter for gitea)
- @release-drafter/forejo-adapter (remote adapter for forgejo)

I have yet to read how you implemented logging in this new cli wrapper, but I already see how a wrapper would (for example) create a logger instance for @release-drafter/core to use.

There should be similar instances for the octokit client and stuff i believe.

Let me know what you think !

Also, similar discussion here : https://github.com/release-drafter/release-drafter/pull/1685#issuecomment-5115998229 (for a potential @release-drafter/pr-title-check)

vite.config.ts
```diff
@@ -21,43 +26,159 @@ function syncDrafterActionYml(): Plugin {
...
+        if (output.type !== 'chunk') continue
+
+        const actionModule = output.moduleIds.find((id) =>
+          id.includes('/node_modules/@actions/'),
+        )
+        if (actionModule) {
+          this.error(
+            `Public bundle ${output.fileName} contains an Actions runtime module: ${actionModule}`,
+          )
+        }
```
cchanche, line 55: It feels weird we have to do this... See comment about monorepo

vite.config.ts
```diff
@@ -21,43 +26,159 @@ function syncDrafterActionYml(): Plugin {
...
+        let relative = path.relative(path.dirname(file), target)
+        if (!relative.startsWith('.')) relative = `./${relative}`
+
+        return `${quote}${relative.split(path.sep).join('/')}.js${quote}`
+      },
+    )
+
+    if (/(['"])(?:#src\/|\.{1,2}\/)[^'"]+\.ts\1/.test(rewritten)) {
+      throw new Error(`Failed to rewrite declaration imports in ${file}`)
+    }
```
cchanche, line 104: I am a bit uncomfortable with this tbh 😅 rewriting imports and declarations feels like we are reinventing the wheel somehow.

You probably wrote more library stuff than I did and have more exeprience, maybe it is a common pattern i don't know

.github/workflows/release.yml
```diff
@@ -31,27 +33,68 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version-file: .node-version
+          cache: npm
+
+      - name: Install npm with staged publishing support
+        run: npm install --global npm@12.0.1
```
cchanche, line 70: I feel like this should be included in the step above (actions/setup-node)

jetersen:
README is conflicted :sweat:

cchanche:
I'll look into it, probably my formatter thing

Since biome does not support markdown yet, i gave readme.md a one-off formatting pass for v7.7.0

jetersen:
Setting up formatter again for prettier to look at markdowns

src/actions/drafter/lib/find-pull-requests/find-commits-in-comparison.ts
```diff
@@ -1,25 +1,130 @@
...
     octokit.graphql,
     FindCommitsInComparisonDocument,
-    params,
+    {
+      ...comparisonParams,
+      useCommitishes,
+      headCommitish: useCommitishes
+        ? commitishToCommitExpression(params.headCommitish)
+        : params.headCommitish,
+    },
```
cchanche, line 84: I am very confused about what happens here. Even with opencode trying to explain it to me 😅 

From what I understand, there's a workaround using REST for people using the CLI's `--from` arguement, which is kind of arbitratry.

What about using the same codepath for the regular action runtime ? Wouldn't that simplify this function a bit ?

src/cli/auth.ts
cchanche, line 1: that's nice !

src/cli/run.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (value === true || value === 'true') return true
+  if (value === false || value === 'false') return false
+  throw new Error(`--${name} must be true or false`)
+}
+
+const cli = cac('release-drafter')
+
+cli
+  .command('<repository>', '✍️ Create, update, or publish a GitHub release')
+  .option('-f, --from <commitish>', 'Override the previous release')
```
cchanche, line 36: cac's syntax feels good !

src/common/string-to-regex.ts
cchanche, line 1: nice catch

src/drafter.ts
cchanche, line 1: How is this file different from src/actions/drafter/runner.ts ?

If it is indeed different or not, I'd like it to be in the src/actions/drafter (as long as we don't do monorepo)

cchanche (changes requested):
I think gitea and forgejo support should have been implemented in separate PRs, this is a lot

If we end up rewriting this somewhere else following a monorepo rewrite, maybe it'll be the opportunity to split concerns better

Otherwise, this looks very good to me ! Exploring REST endpoints for compatibility seem to show they sometimes are more stable and reliable than GraphQL equivalents. Maybe progressively going for REST instead of Gql is not that bad even for github runtimes

vite.config.ts
cchanche: Ah, slight edit : cli / gh-actions are runtimes, and github/gitea/forgejo are remotes.

Corresponding packages should be runtime wrappers and remote adapters

jetersen:
Heads up @cchanche — I force-pushed a rebase onto `main` (`2d41729` → `b1c1367`), which replaces your merge commit. You'll need `git fetch && git reset --hard origin/feat/local-cli` on your clone. Sorry for the collision; we were integrating at the same time.

I diffed the two integrations before rewriting, and the trees were identical apart from `README.md` and one line of `package.json` — no `src/`, `dist/`, or test differences, so no code resolution was lost. Two things the rebase recovers:

- the `no-new-contributor-template` row from #1687, which the merge's README resolution dropped (the code for it was present either way, only the docs row went missing)
- the `commitish` row still said "e.g. `master` when configured to run on pushes to `master`"; `main` now

The `package.json` difference was `all` ordering — `build` before or after `format:md:write`. Functionally equivalent since the Markdown step touches only `.md`; both keep `build` ahead of `tsc:check`. I kept `check && format:md:write && build && …`.

### One check will be red until #1689 lands

The README's `js` example gets rewritten to double quotes and semicolons by `npm run all`, because #1688 left Prettier formatting fenced code with *its own* defaults instead of Biome's. Your merge committed the rewritten form, which is why it was green here. #1689 fixes the root cause by mirroring `javascript.formatter` from `biome.jsonc`; this branch keeps the example in the repo's style, so `check:clean` fails until that merges. Once it does, I'll rebase and this goes green with the example intact.

jetersen:
Think I want to work on getting the split done over the weekend

cchanche:
Sure ! I took the liberty of going through your changes a bit early then. I can do it again when you feel ready (and if you think it's useful). Feel free to close non-relevant threads

src/actions/drafter/lib/find-pull-requests/find-commits-in-comparison.ts
jetersen: Ya I thought about moving the from to the action input as well.

.github/workflows/release.yml
jetersen: actions/setup-node does not guarantee a given npm version from what I recall and you cannot pick a version from setup-node.

PR #1689, opened 2026-07-29
style: adapt prettier markdown formatting

jetersen:
## Summary

Follow-up to #1688. That PR left Prettier formatting code inside fenced blocks, which is desirable — but Prettier formats it with **its own defaults**, not Biome's. Nothing on `main` exposed the gap, because every fence here is YAML. A JavaScript fence does:

```diff
-import { draftRelease } from 'release-drafter'
+import { draftRelease } from "release-drafter";
```

Double quotes and semicolons, contradicting `biome.jsonc` (`quoteStyle: single`, `semicolons: asNeeded`) — the style of the very code the example documents. This surfaced while rebasing #1684, which adds the first `js` fence to the README.

## Fix

Mirror `javascript.formatter` from `biome.jsonc` in `.prettierrc.yml`, with a comment noting the two must stay in sync.

## Why the YAML examples changed

Prettier applies `singleQuote` to YAML too, and it cannot scope quote style per embedded language — so this normalizes the config examples to single quotes. That is a real 168-line diff, and worth reviewing deliberately, but it moves the docs *toward* the repository's own convention:

| | single-quoted | double-quoted |
|---|---|---|
| repo `.yml`/`.yaml` files | 260 | 38 |
| README examples (before) | 4 | 58 |
| README examples (after) | 61 | 1 |

`.github/release-drafter.yml` — the canonical instance of the config the README documents — is single-quoted throughout, so the examples now match what maintainers actually write.

## Quoting that carries meaning is preserved

In YAML, single quotes do not process escapes, so a blind conversion would corrupt values. Prettier is value-aware and does not do that:

```console
$ npx prettier --config .prettierrc.yml sample.md
change-authors-separator: "\n"   # left alone — escape would break
plain: 'hello'                   # converted
```

`README.md:322` keeps `change-authors-separator: "\n    "` double-quoted for exactly this reason. No line containing a backslash appears anywhere in the diff.

## Alternative considered

`embeddedLanguageFormatting: 'off'` would fix the JS problem with zero churn, but gives up machine-formatted YAML examples entirely. Mirroring the style keeps the examples enforced and consistent with the codebase.

## Verification

`npm run all` passes, `npm run check:clean` exits 0, and `npm run format:md:check` is idempotent.

> [!NOTE]
> ### Align fenced code block formatting with repo code style via Prettier config
> Adds a [.prettierrc.yml](https://github.com/release-drafter/release-drafter/pull/1689/files#diff-c1167bf7ca27b627e9d890eddb7e6f595f1327c3346050b6d1a7bec85c3aae2d) config to enforce single quotes, no semicolons, trailing commas, and consistent bracket/arrow formatting across JS/TS. Updates code examples in [README.md](https://github.com/release-drafter/release-drafter/pull/1689/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) and [docs/configuration-loading.md](https://github.com/release-drafter/release-drafter/pull/1689/files#diff-1a69b77af86ae318c29cc4b3ace72a9c9108a42da83a04ed2c6f9607ae5ed08b) to match the new style.
>
> 
> #### Changes since #1689 opened
>
> - Configured Prettier to process only Markdown files repository-wide [88df273]
> - Modified `.prettierignore` to simplify ignore patterns by removing explicit directory and file exclusions while maintaining global ignore with selective re-inclusion of markdown files [9f0555b]
> - Reformatted the `last-not-found.md` markdown template to include blank lines after blockquote markers and reflowed long lines into multiple wrapped lines [9f0555b]
> - Changed Prettier configuration to preserve authored line breaks in prose [1256f9a]
> - Reformatted `last-not-found.md` Markdown template and updated corresponding test snapshots [1256f9a]
> - Removed trailing spaces after Markdown admonition markers [edbd46e]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c83b0ad.</sup>
> 
>
>

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,9 +1,13 @@
-# Generated or vendored output.
+# Markdown only; Biome owns every other language.
+# `!*/` is required: gitignore cannot re-include a file under an excluded directory.
+*
+!*/
+!**/*.md
+
 dist
 coverage
 .licenses
 logs
 
-# Release note content shipped to users, not documentation. Reflowing these
-# templates would rewrite the text that ends up in drafted releases.
+# Release note templates, shipped to users as content.
 src/actions/drafter/lib/build-release-payload/static
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,13 +1,4 @@
-# Markdown only; Biome owns every other language.
-# `!*/` is required: gitignore cannot re-include a file under an excluded directory.
+# Markdown only; `!*/` lets the negation reach files below the root.
 *
 !*/
-!**/*.md
-
-dist
-coverage
-.licenses
-logs
-
-# Release note templates, shipped to users as content.
-src/actions/drafter/lib/build-release-payload/static
+!*.md
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -11,3 +11,16 @@ proseWrap: always
 tabWidth: 2
 useTabs: false
 endOfLine: lf
+
+# Prettier formats code inside fenced blocks, and it does so with its own
+# defaults rather than Biome's. Left alone it rewrites JavaScript examples with
+# double quotes and semicolons, contradicting the style of the code they
+# document. These mirror `javascript.formatter` in biome.jsonc and must stay in
+# sync with it.
+singleQuote: true
+semi: false
+quoteProps: as-needed
+trailingComma: all
+bracketSpacing: true
+bracketSameLine: false
+arrowParens: always
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,10 +1,4 @@
-# See: https://prettier.io/docs/en/configuration
-#
-# Prettier is scoped to Markdown only. Biome owns every other language in this
-# repository, but it has no Markdown formatter, so `.md` files were silently
-# skipped by `biome check` (unknown file types are ignored). The `format:md:*`
-# scripts and the ignore rules in .prettierignore keep the two tools from
-# formatting the same files.
+# Markdown only; Biome owns every other language but cannot format Markdown.
 
 printWidth: 80
 proseWrap: always
@@ -12,11 +6,8 @@ tabWidth: 2
 useTabs: false
 endOfLine: lf
 
-# Prettier formats code inside fenced blocks, and it does so with its own
-# defaults rather than Biome's. Left alone it rewrites JavaScript examples with
-# double quotes and semicolons, contradicting the style of the code they
-# document. These mirror `javascript.formatter` in biome.jsonc and must stay in
-# sync with it.
+# Applied to code in fenced blocks. Mirrors `javascript.formatter` in
+# biome.jsonc; keep in sync.
 singleQuote: true
 semi: false
 quoteProps: as-needed
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,7 +1,9 @@
 # Markdown only; Biome owns every other language but cannot format Markdown.
 
 printWidth: 80
-proseWrap: always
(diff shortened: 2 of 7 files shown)


cchanche (changes requested):
Why not scope prettier to .md files only ?

`.prettierignore`
```text
# Ignore everything
*

# ...except markdown files:
!**/*.md
```

cchanche (approved):

jetersen:
Followed up on the `static/` exclusion rather than leaving it in — pushed as 9f0555b.

### Why it was excluded, concretely

The template is two GitHub alerts whose markers sat directly above their prose, which is exactly what `proseWrap: always` joins. Formatting it destroyed both alerts, confirmed against GitHub's own renderer:

| | alerts rendered |
|---|---|
| before | 2 — `markdown-alert-warning`, `markdown-alert-important` |
| formatted, unfixed | **0** — plain `<blockquote>` |

Users would have seen the literal marker text in their release notes. This is release content rather than documentation: `build-release-payload.ts` imports it with `?raw` and splices it into the body whenever there is no previous release.

### Fix

Separating the markers with a blank quote line makes the file stable under Prettier, so it no longer needs excluding. Verified through `gh api /markdown`:

- both alerts still render, with titles
- rendered text is **identical** before and after, including after `$OWNER/$REPOSITORY` substitution

The four affected e2e inline snapshots were regenerated; the diff is confined to the template region.

### Ignore file simplified

Down to three lines:

```
# Markdown only; `!*/` lets the negation reach files below the root.
*
!*/
!*.md
```

`!*.md` matches at any depth, so `!**/*.md` was redundant, and the explicit `dist` / `coverage` / `.licenses` / `logs` entries were too — none contain Markdown, and the `*` baseline already excludes everything not re-included. Verified all 9 Markdown files are picked up and every non-Markdown file is ignored.

On the question of a config-level way to express "Markdown only": there isn't one. Prettier's full option list from `--support-info` has no include/scope key, so CLI globs and the ignore file are the only mechanisms.

jetersen:
Switched to `proseWrap: preserve` in 1256f9a, which supersedes the template work in 9f0555b.

### Why no width setting fixes it

Prettier breaks *before* a long link because it cannot break inside a URL. The "open an issue" line in the release note template is 167 characters, so no sensible width keeps it intact:

| printWidth | result |
|---|---|
| 80 | 2 lines |
| 120 | 2 lines |
| 160 | 2 lines |
| 200 | 1 line |

Raising the width is also not free in general: measured at roughly 550 reflowed lines across the docs, and since `printWidth` governs fenced code as well as prose, it would desync code examples from Biome's `lineWidth: 80` — the very thing c83b0ad aligned. Verified: a 78-character `js` statement in a fence stays one line at width 100 and breaks at 80.

### What `preserve` gives us

Prettier still normalizes structure while leaving authored line breaks alone:

- misaligned tables → still padded and aligned
- a 120-character prose line → left exactly as written
- `> [!note]` above its prose → **no longer joined**, so the alert-breaking hazard disappears at the root

Every Markdown file in the repo is already byte-identical under it, so there is zero churn.

### Consequences

Because reflowing is what broke the release note template, 9f0555b is reverted: the template, the four e2e snapshots asserting its text, and the bundle are back to their previous bytes. The template no longer needs excluding either — `preserve` cannot join the marker into the prose. Net change against 88df273 is now just the two config files.

The tradeoff, stated plainly: wrap width is no longer machine-enforced for prose, so a long line stays long until someone wraps it. The existing docs are all within 80 already.

jetersen:
I ended up fixing the configs and ignore. Hopefully this aligns better with what you asked for @cchanche

cchanche:
Thank you !

PR #1698, opened 2026-08-02
build: establish Node 24 workspace and CI tooling

jetersen:
## Summary

- Convert the repository root into a private npm workspace coordinator with ten Node 24 packages.
- Add shared Vite 8 builds, TypeScript declarations through `rolldown-plugin-dts`, package guards, dependency checks, and workspace boundary checks.
- Consolidate pull request CI into title validation, Action smoke tests, and the main test/build job. Remove CodeQL, cancel superseded runs, and enforce the existing coverage threshold without a second install.
- Keep root `dist/` tracked, ignore workspace package output, and classify generated bundles, schemas, lockfiles, and license data for review.

## Dependency changes

- Add workspace development tooling, including `vite@^8.2.1` and `rolldown-plugin-dts@^0.28.2`.
- Require Node `^24.11.0 || >=26.0.0` and npm `>=12.0.1` for root development tooling.
- Do not add a published runtime dependency or enable npm publication.

## Validation

- `npm run ci`: 24 test files and 454 tests passed.
- `npm run check:clean`
- Coverage: 90.99%, above the 90% threshold.
- Package, dependency, boundary, declaration, and clean-generation checks passed.

## Stack

This is the bottom layer of the monorepo stack. It establishes the workspace and CI foundation without moving the existing Action behavior.

Closes #1692.

## Review size

- Reviewable changes: `+1,516 -201` across 57 files
- Generated artifacts: `+3,408 -965` across 4 files

GitHub's headline total includes both rows. `.gitattributes` marks generated artifacts and collapses their diffs by default.

Issue #1693, opened 2026-08-02
Core: extract release orchestration and forge contracts


jetersen:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Add a description

Parent RFC: #1691

## Scope

Extract forge-independent release behavior into private
`@release-drafter/core` and define the runtime/adapter contracts consumed by
Actions and the public facade.

Move configuration schemas and merge behavior, normalized domain types,
category evaluation, changelog rendering, version calculation, release
orchestration, dry-run safeguards, and explicit `Logger`/`ForgeAdapter` ports
without changing existing behavior. Extract reusable autolabeling configuration
and pure matching into `@release-drafter/autolabeler`.

Then place existing GitHub and GHES API behavior behind
`@release-drafter/github-adapter`, preserving the optimized GraphQL path and
reviewed REST compatibility fixes from #1684. Temporary compatibility exports
may keep the existing Actions working while the stack lands.

Suggested foundation branches:

1. `monorepo/core-contract`
2. `monorepo/github-adapter`

## Acceptance criteria

- [x] Core imports no `@actions/*`, Octokit, GitBeaker, environment-specific
      runtime globals, or forge response types.
- [x] Named forge-neutral DTOs replace types inferred from API response shapes.
- [x] One pure category evaluator covers conventional rules, labels, paths,
      pre-inclusion/exclusion, exclusivity, version rules, and fallback status.
- [x] Release generation and autolabeling retain their current observable
      behavior.
- [x] The GitHub adapter owns client construction, GraphQL/REST calls, config
      retrieval, ref comparison, contributor lookup, and release writes.
- [x] GitHub.com, GHES endpoints, proxies, pagination, lightweight/annotated
      tags, and REST/GraphQL parity remain covered.
- [x] Existing Action smoke tests remain green through compatibility wrappers.
- [x] Generated schemas, GraphQL types, and Action bundles are committed with
      the PR that changes them.
- [x] Each PR finishes with `npm run all` and `npm run check:clean`.

## Out of scope

- Public npm publication.
- CLI argument parsing and packaging.
- Gitea, Forgejo, or GitLab adapters.
- Moving GitHub Actions entrypoints.
- Testcontainers infrastructure.
- Downstream monorepo behavior from #469.

## Dependencies and parallelism

Stacks on the workspace foundation. The core contract lands before the GitHub
[...]

PR #1699, opened 2026-08-02
feat: extract forge-neutral core contract

jetersen:
## Summary

- Move release configuration, category evaluation, rendering, versioning, planning, and execution into private `@release-drafter/core`.
- Define forge-neutral DTOs plus explicit `Logger` and authenticated `ForgeAdapter` contracts.
- Move runtime-neutral Autolabeler configuration and matching into private `@release-drafter/autolabeler`.
- Add `ReleaseInput.from` as an optional comparison baseline without changing release selection or version resolution.
- Keep the existing Actions compatible through thin wrappers and GraphQL-to-core normalization.

## Dependency changes

- Register the private core and Autolabeler workspaces as root workspace dependencies.
- Add no external runtime dependency to the core contract.

## Validation

- `npm run ci`
- `npm run check:clean`
- Core, compatibility, Autolabeler, schema, package-boundary, and generated-bundle checks passed.
- Coverage remained above the 90% threshold.

## Stack

Stacked on #1698. The GitHub transport implementation follows in #1701.

Part of #1693.

## Review size

- Reviewable changes: `+4,286 -2,087` across 66 files
- Generated artifacts: `+2,113 -1,989` across 4 files


tsconfig.json
```diff
@@ -25,7 +25,11 @@
     "noEmit": true,
     "useDefineForClassFields": true,
     "noFallthroughCasesInSwitch": true,
-    "paths": { "#src/*": ["./src/*"] },
+    "paths": {
+      "#src/*": ["./src/*"],
+      "@release-drafter/autolabeler": ["./packages/autolabeler/src/index.ts"],
+      "@release-drafter/core": ["./packages/core/src/index.ts"]
```
cchanche: So I get how importing stuff from core makes sense for the drafter action.

Drafter still resides in src/actions/drafter, but most of it's logic was moved to @release-drafter/core. Remaining code will be split between facades (actual package & CLI) and adapters.

Autloabeler however seem to follow a different pattern here. Ex: drafter's config schema is in @release-drafter/core, but autolabeler's is in @release-drafter/autolabeler.

What's the plan here ?

cchanche: If that stuff has been adressed in another stacked PR down the line, simply resolve @jetersen 

I am unsure how rebasing the rest of the stack would go if we were to add commits here.

cchanche (changes requested):

tsconfig.json
jetersen: gh stack rebase should just work, so sure I can take a look :)

jetersen: The split is intentional. `@release-drafter/core` contains Release Drafter's configuration and orchestration. Autolabeler is separate, so its schema and pure matching logic belong in the peer `@release-drafter/autolabeler` package, not in core.

Some of the inconsistency visible in this PR is temporary:

- #1702 adds the public `release-drafter` facade over core.
- #1705 moves the Drafter and Autolabeler Action composition from `src/actions/*` to `@release-drafter/gh-actions`.
- #1712 removes the last obsolete `#src/types/*` migration reference.

Once the full stack lands, `src/` will contain only the private root project's repository tooling: repo-wide scripts, integration and E2E tests, fixtures, mocks, and test helpers. The remaining `#src/*` imports are limited to that tooling and its tests. None of the workspace packages import through the root alias anymore.

We could rename that area to root-level `tooling/` and `tests/`, but I'd keep that as a separate cleanup. I also wouldn't make it a `packages/tooling` workspace. These scripts operate across and test the whole workspace; they aren't a reusable package. Keeping them in the private root project's `src/` doesn't leave the package split half-finished.

cchanche: > @release-drafter/core is the domain engine for the release-drafter, so it owns configuration and orchestration. Autolabeler is a separate product/domain, so its schema and pure matching logic live in the peer @release-drafter/autolabeler package rather than in core.

LGTM !

cchanche (approved):

PR #1701, opened 2026-08-02
feat: add GitHub adapter

jetersen:
## Summary

- Add private `@release-drafter/github-adapter` as the GitHub implementation of the core `ForgeAdapter` contract.
- Move Octokit construction, GitHub.com and GHES endpoints, proxy handling, pagination, configuration loading, refs, tags, contributors, and release operations behind the adapter.
- Use GraphQL for hydration, changed files, contributors, and refs. Keep REST for exact comparisons, raw configuration, release fields, and release writes.
- Limit recent-pull-request recovery to the five most recent merged pull requests after direct association lookup.
- Keep the existing Actions working through thin wrappers and regenerate their tracked bundles.

## Dependency changes

- Move Octokit transport plugins and `undici` ownership to the GitHub adapter workspace.
- Add `@graphql-typed-document-node/core` for generated GraphQL documents.
- Publish no package in this layer.

## Validation

- `npm run ci`
- `npm run check:clean`
- Adapter coverage includes GHES, proxies, retries, tags, comparisons, bounded pull request recovery, pagination, and release operations.
- Type checks, package boundaries, generated GraphQL output, and tracked Action bundles passed.

## Stack

Stacked on #1699. The public programmatic API follows in #1702.

Closes #1693.

## Review size

- Reviewable changes: `+2,275 -39,398` across 51 files
- Generated artifacts: `+37,438 -5,821` across 5 files

GitHub's headline total includes both rows. `.gitattributes` marks generated artifacts and collapses their diffs by default.

Issue #1694, opened 2026-08-02
Package: expose the CLI and programmatic API facade


jetersen:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Add a description

Parent RFC: #1691

## Scope

Create private `@release-drafter/cli` and the sole initially publishable
`release-drafter` facade. Selectively port the reviewed CLI, injected logging,
and typed programmatic API from #1684 instead of adopting that PR's monolithic
build.

Suggested stack:

1. `public/programmatic-api`
2. `public/cli-and-package`

The facade must bundle private workspace implementation with Vite. Importing
its root export must not execute the CLI.

## Acceptance criteria

- [x] `release-drafter` exports a typed `draftRelease()` API and forge-neutral
      option/result types.
- [x] Programmatic consumers can inject a `ForgeAdapter` and `Logger`.
- [x] The package declares an executable `release-drafter` binary so
      `npx release-drafter` works.
- [x] The CLI initially supports the reviewed repository, range, config,
      release, dry-run, publication, JSON, and server/API options from #1684.
- [x] Forge selection is explicit and testable; convenience detection never
      makes Gitea/Forgejo identity claims from an ambiguous `/api/v1` URL.
- [x] JSON mode writes exactly one JSON document to stdout; logs and diagnostics
      go to stderr.
- [x] Native ESM import and NodeNext declarations work in an isolated consumer.
- [x] An isolated packed-package test runs the binary and imports the API.
- [x] Public JavaScript and declarations contain no unresolved private
      workspace imports, `@actions/*`, GitBeaker types, or CommonJS loader
      shims.
- [x] All scoped workspaces remain private.
- [x] Each PR finishes with `npm run all` and `npm run check:clean`.

## Out of scope

- Publishing the package to npm.
- Production npm credentials or permissions.
- Moving the Action entrypoints.
- Implementing alternate forge adapters in this stack.
- Porting unrelated structure or unresolved defects from #1684.
- Downstream monorepo behavior from #469.

## Dependencies and parallelism

Depends on the workspace foundation, core contract, and GitHub adapter. It may
run in parallel with the Actions and alternate-forge stacks after those
foundations land.

jetersen:
Implementation status: all acceptance criteria are implemented across #1702, #1703, and #1704, with final multiforge and packed `npx` verification in #1709 and #1711. This issue remains open until #1704 merges.

PR #1702, opened 2026-08-02
feat: expose programmatic API

jetersen:
## Summary

- Expose a typed, forge-neutral `draftRelease(options)` API from the public `release-drafter` package.
- Bundle the core implementation into native ESM runtime output with self-contained TypeScript declarations.
- Add build-artifact, packed NodeNext consumer, runtime smoke, and workspace-boundary coverage.
- Document adapter injection, parsed configuration, dry-run behavior, and the default no-op logger.

## Dependency changes

- Add private `@release-drafter/core` as a workspace build dependency of the public package.
- Bundle the core implementation so the packed package has no runtime workspace dependency.

## Validation

- `npm run ci`
- `npm run check:clean`
- Pack, install, type-check, import, and invoke the package from an isolated NodeNext consumer.
- Scan runtime and declaration output for private workspace, forge, Actions Toolkit, CommonJS loader, and `require()` leakage.

## Stack

Stacked on #1701. CLI and executable packaging follow in #1704 after the dependency cleanup layers.

Refs #1694.

## Review size

- Reviewable changes: `+919 -4` across 10 files
- Generated artifacts: `+4 -0` across 1 file


vite.workspace.config.ts
```diff
@@ -6,7 +6,41 @@ const packageJson = process.env.npm_package_json
 if (!packageJson)
   throw new Error('npm_package_json is required to build a workspace')
 const workspaceRoot = dirname(packageJson)
+const packageName = process.env.npm_package_name
+if (!packageName) throw new Error('npm_package_name is required')
+const convertSemverModuleToEsm = (source: string) => {
+  let importIndex = 0
+  let converted = source.replace(/^['"]use strict['"];?\s*/u, '')
+  converted = converted.replace(
+    /const\s+(\{[\s\S]*?\}|[$\w]+)\s*=\s*require\((['"])([^'"]+)\2\)/gu,
```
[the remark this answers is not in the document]
jetersen: Verified against current head d3eebf38927b1b8468a3176f4c362c7d8ad87d99: the regex at line 15 has no `^` anchor and already uses the global `g` flag, so it converts every matching `require()` declaration. The generated-facade tests also scan for `require(`, `__require`, and CommonJS shims, and the packed NodeNext runtime invocation passes. This finding describes a form that is not present in the PR head, so no code change is needed.

packages/release-drafter/src/index.ts
cchanche, line 1: Are all these types and interfaces specific to the facade ? If so, I believe they should be refactored either into a dedicated file or hoisted into core

Having index.ts be like : 

```ts
// index.ts
import * as releaseDrafterCore from '@release-drafter/core'
import type { some_shared_types } from '@release-drafter/core'
import type { some_local_types } from './types.ts'


/**
 * Calculates and optionally creates or updates a release through an injected
 * forge adapter.
 */
export const draftRelease = async (
  options: DraftReleaseOptions,
): Promise<DraftReleaseResult> => {
  const result = (await releaseDrafterCore.draftRelease({
    adapter: options.adapter,
    config: options.config,
    input: options.input,
    logger: options.logger ?? defaultLogger,
    repository: options.repository,
  })) as DraftReleaseResult

  return {
    plan: result.plan,
    release: result.release,
    releasePayload: result.releasePayload,
  }
}
```

... would make things a lot easier to read & understand for contributors

cchanche (changes requested):

packages/release-drafter/src/index.ts
jetersen: Should be resolved

PR #1706, opened 2026-08-03
feat: add config-driven check-pr validation

### LINKED ISSUE #1695: Actions: consolidate drafter, autolabeler, and check-pr-title entrypoints
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Add a description

Parent RFC: #1691

## Scope

Create one private `@release-drafter/gh-actions` runtime workspace with three
separate Vite entrypoints and root distribution bundles:

- release drafter;
- autolabeler; and
- `check-pr-title`.

Suggested stack:

1. `actions/gh-actions-workspace`
2. `actions/check-pr-title`

The workspace groups shared Actions Toolkit/runtime composition; it must not
combine the products into one entry bundle. Pure release, autolabeling, and
category logic remain in their domain workspaces.

`check-pr-title` validates configured `when.conventional` conditions. Labels
and changed paths are considered only when part of the same condition. A change
excluded by configuration passes as skipped; matching only an unconditional
fallback category fails. It must not mutate the pull request.

## Acceptance criteria

- [x] Drafter, Autolabeler, and `check-pr-title` have distinct source and
      `dist/actions/*/run.js` entrypoints.
- [x] Existing root, `/drafter`, and `/autolabeler` Action references remain
      compatible.
- [x] Existing metadata, inputs, outputs, schemas, GHES behavior, and proxy
      behavior remain covered.
- [x] All Action metadata uses Node 24.
- [x] Drafter and Autolabeler retain their existing behavior after moving.
- [x] `check-pr-title` loads the normal Release Drafter configuration and reuses
      core's pure category evaluator.
- [x] Title-check tests cover valid, invalid, fallback-only, excluded,
      path/label-sensitive, edited-title, `_extends`, and malformed-event cases.
- [x] The title Action requests read-only permissions and performs no label or
      release mutation.
- [x] Action bundle scans prevent accidental dependencies on other Action
      entrypoints or public-facade-only code.
- [x] Generated metadata and bundles are committed with the change that creates
      them.
- [x] Each PR finishes with `npm run all` and `npm run check:clean`.

## Out of scope

- Publishing `@release-drafter/gh-actions` to npm.
- Independently versioning the three Actions.
- A second list of allowed conventional types or scopes.
- Alternate forge support for `check-pr-title`.
- Replacing repository CI with the new Action before parity is demonstrated.
- Downstream monorepo behavior from #469.

## Dependencies and parallelism

Depends on the foundation, core evaluator, and GitHub adapter. It can proceed
in its own worktree and stack alongside public packaging and alternate forge
work after those contracts land. Dogfooding belongs in a later tooling-only PR.

jetersen:
## Summary

- Add a read-only Node 24 `check-pr` Action that validates the current pull request with the normal Release Drafter configuration.
- Keep validation in `@release-drafter/core` instead of creating another package.
- Add `release-drafter check-pr <repository> <number>` so the CLI uses the same policy.
- Add a forge-neutral `PullRequestReader` contract and implement it for GitHub.
- Add `check-pr/action.yml` and generate its inputs through the Action metadata tooling from #1705.

## Validation behavior

- Conditions with `conventional` validate the title.
- Conditions with configured labels validate the current labels.
- A condition that contains both requires both to match.
- Path predicates do not take part in this check. Autolabeler remains responsible for path-based labeling.
- Pull requests excluded by title or label pre-categories pass as skipped.
- Pull requests that select no category, or only an unconditional fallback, fail.

The Action reads the title, labels, and base branch from the pull request event. It does not fetch changed files or modify the pull request, so its example workflow needs only `contents: read`.

The CLI fetches one pull request, loads configuration from its base branch, and returns 0 for valid or skipped, 1 for invalid, and 2 for command or runtime errors.

## Dependency changes

No new package or external runtime dependency. Gitea, Forgejo, and GitLab readers follow in #1707 and #1708.

## Validation

- `npm run ci`
- `npm run check:clean`
- Action, CLI, configuration inheritance, title, label, exclusion, fallback, metadata, and bundle tests passed.

## Stack

Stacked on #1720. Alternate forge readers follow in #1707 and #1708.

Closes #1695.

## Review size

- Reviewable changes: `+1,152 -45` across 39 files
- Generated artifacts: `+1,725 -1,567` across 5 files


jetersen:
@cchanche this is currently only a github action but I wonder if we should introduce it in CLI or support the different forge adapters.

So something like this:

```
  @release-drafter/core
  └── configuration, category primitives, normalized PR types

  @release-drafter/check-pr-title
  ├── title-validation policy
  ├── PullRequestReader port
  └── forge-neutral validation orchestration

  Forge adapters
  └── implement single PR/MR retrieval

  @release-drafter/gh-actions
  └── current GitHub event → validator

  @release-drafter/cli
  └── selected forge and PRs → validator
```

cchanche:
Yeah this sounds good to me.
> selected forge and PRs → validator

> current GitHub event → validator

Would `validator` be something like 

```ts
import {PRTitleValidator} from @release-drafter/check-pr-title

...

const isValid = PRTitleValidator.validate(title)
```

... ?


Also, I like to think about _changes_ in a more generic way (instead of simple _PRs_). Maybe this could be a higher level validator for any type of change. We'd omit a body and a footer for PR titles, and include them (sometimes) for commits. On top of doing the PR categorization stuff, we would prepare for whenever we want to implement commits into what we consider as possible changes.

jetersen:
Well ya thinking about it we could prepare it for semantic commit and semantic pr title support :+1:

cchanche:
but then about the _gh-action_ side of things, it would only make sense for the PR-titles I guess. So maybe two pacakges then ? One for the actual action `@release-drafter/check-pr-title`, and one for the lib `@release-drafter/change-validator` ? Or maybe `@release-drafter/change-validator`'s content should simply be in `@release-drafter/core`

jetersen:
I think change-validator could live in core cause `cli`, `gh-actions`/`check-pr-title` and forge adapters may need pr fetching title support

jetersen:
Updated the stack around the design discussed here:

- the validation policy now lives in `@release-drafter/core`; there is no separate validator package
- the Action is now `check-pr`, not `check-pr-title`
- validation projects the configured categories to title (`conventional`) and label predicates only
- paths are deliberately ignored because Autolabeler owns path-based label assignment
- the CLI now exposes `release-drafter check-pr <repository> <number>` and uses the same core policy
- #1707 and #1708 add single-PR or MR readers for Gitea, Forgejo, and GitLab, while #1709 wires those readers into the CLI

The Action uses the GitHub event payload directly, so it remains small and needs no forge request or separate package.

PR #1716, opened 2026-08-15
test: speed up and stabilize forge conformance

jetersen:
## Summary

- seed the randomized access token during initial database setup instead of booting Rails again through `gitlab-rails runner`
- disable unused GitLab KAS, OpenSSH, logrotate, monitoring, NGINX, and Sidekiq metrics for the conformance-only instance
- expose Workhorse directly over TCP, keep the required Puma and Sidekiq services constrained, and skip clearing an empty Redis cache
- keep disposable PostgreSQL data on tmpfs to reduce initialization I/O
- use Testcontainers' immediate default stop behavior and collect complete container logs only during failure teardown for every real-forge job
- recover group and project creation when GitLab commits a POST but closes the warm-up response before the client receives it
- stabilize Gitea and Forgejo merges by delaying base-branch PR rechecks, avoiding an upstream race that can overwrite the persisted merge base while a merge is saved
- replace the fixed Gitea/Forgejo setup delay with bounded API-state polling; no queue-flush command or unconditional sleep is required
- remove low-signal source-text, static-capability, and generated merge-timestamp assertions while retaining server-backed protocol coverage
- reduce cached-image GitLab wall clock from 185.8 seconds to 73.1–79.3 seconds (about 57–61%)

## Testing

- repeated `mise exec -- npm run test:conformance:gitlab` benchmarks (7 tests pass; final configuration 73.05s, 73.51s, 75.56s, and 79.27s wall clock)
- 20 consecutive four-core `mise exec -- npm run test:conformance:gitea-forgejo` runs passed with both containers running concurrently (12 behavioral tests, usually 15–18 seconds)
- a no-flush/no-delay-control experiment reproduced the Gitea merge-base corruption by combined run 5; `repository.pull-request.DELAY_CHECK_FOR_INACTIVE_DAYS = 0` then passed all 20 final stress runs without queue flushing
- forced failure teardown produced a non-empty Gitea container log, while successful runs left no log artifact
- `mise exec -- npm run all` (50 files and 628 tests pass, plus formatting, type checks, schemas, builds, dependency checks, and workspace guards)
- full GitLab container logs showed Cinc falling from 63 seconds to about 52 seconds after disabling NGINX and moving PostgreSQL to tmpfs
- GitHub-hosted GitLab conformance passed in 3m18s with a 179.25s Vitest duration, down from 4m54–5m03 and 266.47–273.29s on the previous official-image revision
- a schema-only preinitialized PostgreSQL snapshot passed all tests in 68.27 seconds, but was rejected because the extra 25–40 MiB artifact, matching GitLab secrets, and exact image/PostgreSQL coupling only saved about 5 seconds beyond tmpfs
- disabling Sidekiq was rejected because merge requests never became mergeable; bypassing Workhorse with direct Puma was rejected because repository commit creation failed
- broad GitLab tmpfs, `synchronous_commit = off`, PostgreSQL version-change restart configuration, and explicit PostgreSQL defaults produced no additional gain
- Gitea/Forgejo `/data` and SQLite-directory tmpfs experiments were rejected after exposing branch visibility and persisted merge-base corruption
- rejected `ghcr.io/feskol/gitlab` after its equivalent pinned image took 5m39s hosted and showed no local gain

## Checklist

- [x] My PR title uses the conventional format required by CI.
- [x] I ran `npm run all` locally.
- [x] I updated tests or docs where needed.


jetersen:
Potential follow-up: I tested a locally built, preinitialized GitLab CE 19.1.3 image to avoid repeating the full Omnibus reconfigure on every conformance run.

- Cold local runs took roughly 72–79 seconds.
- Three warm-image runs took 34.10, 34.24, and 34.49 seconds.
- The prototype completed a normal first boot once, captured the required Omnibus state, then started with GITLAB_PRE_RECONFIGURE_SCRIPT and GITLAB_SKIP_RECONFIGURE.

That is roughly a 50% wall-clock reduction, but it is intentionally not included in this PR because the image would bake GitLab/Omnibus state and test credentials into an artifact. Making this suitable for CI would need a controlled, reproducible image build and a clear maintenance/security story. It may be worth revisiting if GitLab startup time becomes a larger bottleneck.

jetersen:
Folded the forge fixes into #1710 and the cleanup-specific changes into #1714, so this layer is now obsolete.

Issue #706, opened 2020-11-12
Group dependencies update in changelog


slawekjaranowski:
Currently we have changelog items like:

```
Bump mockito-core from 3.5.11 to 3.5.13 #165
Bump mockito-core from 3.5.13 to 3.5.15 #177
Bump mockito-core from 3.5.15 to 3.6.0 #184
```

All those PR was made by dependabot and have similar description:

Bump `<component-name>` from `<old-version>` to `<new-version>`

It will be fine to possibility to generate one chenagelog line like:

`Bump mockito-core from 3.5.11 to 3.6.0 #165, #177, #184`

lucacome:
any updates on this?

jetersen:
PRs are welcome :)

robbinjanssen:
I will take a swing at this when I have some spare time! 👍

PR #1371, opened 2023-11-21
Add group pull request capability for dependabot pull requests

netomi:
This fixes #1370 .

This is a POC PR to gather feedback if such a functionality would be deemed useful.

Please be nice to me, thats my first PR for a javascript project, I am happy to receive feedback and improve my code.

I generated the test data from a repo (https://github.com/eclipse-cbi/macos-notarization-service) which served as the starting point for this PR as we have the problem there that regular dependabot updates pollute the changelog generated by release-drafter. We ended up to clean up the release notes manually but this is super tedious.

If this PR is something that you want to include the fixture generation should be added as well.

PR #1725, opened 2026-09-13
feat: merge repeated changes with group-changes

slawekjaranowski:
## Summary

Adds a `group-changes` option that merges pull requests updating the same thing
into a single changelog entry, which is what #706 asks for. A dependency that a
bot bumps several times between releases currently takes one line per bump.

```yml
change-template: '* $TITLE ($NUMBERS) $AUTHORS'

group-changes:
  - pattern: '/^Bump (?<group>.+?) from (?<from>\S+) to (?<to>\S+)$/'
    title-template: 'Bump $GROUP from $FIRST_FROM to $LAST_TO'
```

```
* Bump lib from 1.0.0 to 1.3.0 (#41, #42, #44) @dependabot[bot]
* Bump other-lib from 2.0.0 to 2.1.0 (#43) @dependabot[bot]
```

- `pattern` is matched against the pull request title and has to be a regular
  expression literal with a `group` capture group. Every further `group_<name>`
  capture group extends the grouped value, so a bump of the same dependency in
  another submodule stays a change of its own.
- `title-template` builds `$TITLE` of a merged entry from `$GROUP`,
  `$GROUP_<NAME>`, `$FIRST_<NAME>` and `$LAST_<NAME>`.
- `$NUMBERS` is a new change template variable listing every pull request number
  of an entry, separated with `, `. It renders the single number of an entry that
  was not merged, so one `change-template` fits both.
- `$NUMBER`, `$AUTHOR`, `$BODY` and `$URL` keep coming from the newest pull
  request of an entry, `$AUTHORS` lists the authors of all of them.
- Changes are grouped within every category separately, an entry keeps the
  position of its newest pull request, and `collapse-after` counts entries.
- Without `group-changes` in the configuration the changelog is unchanged.

No dependency changes.

## Testing

- `npm run ci`
- New unit tests in `packages/core/src/config/parse-group-changes.test.ts`,
  `packages/core/src/release/group-changes.test.ts` and
  `packages/core/src/release/pull-request-to-string.test.ts`, plus an end-to-end
  case in `src/tests/drafter/drafter.e2e.test.ts`.
- Running in the s4u organization, where it drafts and publishes real release
  notes. In
  [sign-maven-plugin v1.2.0](https://github.com/s4u/sign-maven-plugin/releases/tag/v1.2.0)
  52 dependency pull requests are rendered as 10 entries, for example 14 bumps of
  `org.junit:junit-bom` as one line with the full version range.

Refs #706

## AI assistance

This change was prepared with the help of AI: the code, the tests and the
documentation were written with Claude Code, and the commit carries a
`Co-Authored-By` trailer for it. I reviewed the result and verified the behavior
on real repositories before opening this pull request.

## Checklist

- [x] My pull request title uses the conventional format required by CI.
- [x] I ran `npm run ci` locally.
- [x] I updated tests and documentation where needed.


cchanche (commented):
hey @slawekjaranowski , categories couldn't do trick ? This feels like an escape hatch to the categories mechanism to me

However, using categories, we'd probably have to implement something around regex and regex groups for category titles, but that should not be this hard

jetersen:
Related to #1371

slawekjaranowski:
@cchanche Thanks for looking at it. I don't think categories can cover this, because the two
work on different axes: a category decides **in which section** a change is rendered,
`group-changes` decides **how many lines a set of changes becomes** inside a section.
They compose — grouping is applied per category, so an `⬆️ Dependencies` category still
renders one section, just with one line per dependency, and `collapse-after` counts the
merged entries.

What #706 asks for is a single line like

```
* Bump mockito-core from 3.5.11 to 3.6.0 (#165, #177, #184) @dependabot[bot]
```

That needs values from several pull requests at once: `from` of the oldest, `to` of the
newest, all numbers and all authors. A category title is rendered once per section and
the changes below it are still one bullet per pull request, so regex groups in category
titles would give a *section per dependency* — for the repository I tested on, 10
headings with 1–14 bullets each instead of 10 lines, and every bullet would still show a
single version step instead of the range.

It would also be the bigger change of the two. Categories are a static, ordered list that
`exclusive`, `collapse-after`, `semver-increment` and the `pre-include` / `pre-exclude` /
`version-resolver` types rely on, while grouping keys only exist at runtime; and `when`
has no title predicate today, so this would add regex matching to the path every
configuration goes through. `group-changes` is opt-in and inert: without it in the
configuration the changelog is unchanged.

slawekjaranowski:
@jetersen Yes — this is that idea rebuilt on the TypeScript codebase, as you suggested when closing
#1371. Compared to that POC it is opt-in and configuration driven instead of a hardcoded
`Bump X from Y to Z`, so it is not dependabot specific; it needs no `compare-versions`
dependency (members are ordered by merge date, the title takes captures from the oldest
and the newest one); the numbers are a regular `$NUMBERS` template variable instead of a
hack in the template engine; and grouping happens per category, so it composes with
categories and `collapse-after` — which is complementary to this: it hides a long list,
this shortens it.

cchanche (commented):
I've run an automated review of this PR and everything looks solid. The code is well-structured, the grouping logic handles edge cases properly, and the graceful degradation for bad regexes is exactly how it should be done.

Good use of standard library sorting over external dependencies for ordering. The fact that this seamlessly composes with existing features like `collapse-after` and categorisation shows good architectural fit.

Approving from my end! 🚀

cchanche:
> I've run an automated review of this PR and everything looks solid.

Dang, I still have to learn how to use claude code properly 😅 Didn't intend for it to comment and approve automatically

I'm going to actually look at the code real quick !

README.md
```diff
@@ -552,6 +554,68 @@ exclude-contributors:
...
+```
+
+Release Drafter matches `pattern` against the pull request title. Write it as a
+regular expression literal, such as `/…/i`, because a plain string is matched
+literally and cannot hold capture groups. A `group` capture group is required
+and holds the value that changes are grouped by. Release Drafter skips a rule it
+cannot use and reports the reason in the action log.
+
+Add a `group_<name>` capture group for every further value that has to match
+before two pull requests are merged. A bump of the same dependency in another
```
cchanche:
```suggestion
before two pull requests are groupped. A bump of the same dependency in another
```

README.md
```diff
@@ -552,6 +554,68 @@ exclude-contributors:
...
+```yml
+group-changes:
+  - pattern: '/^Bump (?<group>.+?) from (?<from>\S+) to (?<to>\S+)(?<group_in> in .+)?$/'
+    title-template: 'Bump $GROUP from $FIRST_FROM to $LAST_TO$GROUP_IN'
+```
+
+```
+* Bump lib from 1.0.0 to 1.2.0 in /module-a (#41, #43) @dependabot[bot]
+* Bump lib from 1.0.0 to 1.1.0 in /module-b (#42) @dependabot[bot]
+```
```
cchanche, line 596: Nice touch adding this example for additional `group_<name>` capture groups !

README.md
```diff
@@ -552,6 +554,68 @@ exclude-contributors:
...
+
+`title-template` builds `$TITLE` of a merged entry:
+
+| Variable        | Description                                                                      |
+| --------------- | -------------------------------------------------------------------------------- |
+| `$GROUP`        | The value of the `group` capture group.                                          |
+| `$GROUP_<NAME>` | The value of the `group_<name>` capture group.                                   |
```
cchanche:
```suggestion
| `$GROUP_<NAME>` | The value of a corresponding `group_<name>` capture group.                                   |
```

README.md
cchanche: So it's basically grouping by unique tuples of `group` and `group_<name>` matches ?

ex: one groupped entry for each : `(lib, module-a)`, `(lib, module-b)`, `(other-lib, module-a)`, `(other-lib, module-b)`, etc....

README.md
```diff
@@ -552,6 +554,68 @@ exclude-contributors:
...
+category separately, and orders the pull requests of an entry by merge date. An
+entry keeps the position of its newest pull request, which is also the source of
+`$NUMBER`, `$AUTHOR`, `$BODY`, and `$URL`. `$AUTHORS` lists the authors of every
+pull request of the entry. A capture group that did not participate in the match
+holds an empty value, and a group that matches a single pull request keeps its
+original title.
+
+`collapse-after` counts entries, so merged changes count as one. Because
+`replacers` run on the finished changelog body, write `pattern` against the
+original pull request titles.
```
cchanche: Doesn't the _"finished changelog entry"_ contains the groupped entries already ? Whether it does or not, this docs section feels confusing

```suggestion
`collapse-after` counts entries, so merged changes count as one. Because
`replacers` run on the finished changelog body, write `pattern` against the
group `title-template`.
```

packages/core/src/config/parse-group-changes.ts
```diff
@@ -0,0 +1,83 @@
...
+import type { ParsedGroupChange } from '../types.ts'
+import type { Config } from './config.schema.ts'
+
+/** Capture group names that `renderTemplate` can expand as `$FIRST_<NAME>`/`$LAST_<NAME>`. */
+const templatableName = /^[A-Za-z_]+$/
+
+/** Capture group names that build the grouping key: `group` and every `group_<name>`. */
+const isGroupName = (name: string) =>
+  name === 'group' || name.startsWith('group_')
+
```
cchanche, line 11: `templatableName` allows capitalized group names, but `isGroupName` doesn't. Is this intended ?

packages/core/src/config/parse-group-changes.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Converts the configured `group-changes` patterns into regular expressions and
+ * collects the capture group names their `title-template` can reference.
+ *
+ * Rules that cannot be used are dropped with a warning so that a single bad
+ * pattern never fails the whole release, matching how `replacers` are handled.
```
cchanche, line 17: That's also a nice touch

packages/core/src/release/generate-changelog.ts
```diff
@@ -31,12 +41,20 @@ export const generateChangeLog = (params: {
 
   if (totalPullRequestsInChangelog === 0) return config['no-changes-template']
   const changeLog: string[] = []
+  // Grouping is applied per bucket so that a pull request matching several
+  // categories is merged with the other changes of each category separately.
+  const toChanges = (categoryPullRequests: PullRequest[]): ChangeGroup[] =>
```
cchanche:
```suggestion
  const toGrouppedChanges = (categoryPullRequests: PullRequest[]): ChangeGroup[] =>
```

README.md
```diff
@@ -320,7 +321,8 @@ Use these variables in `change-template`:
...
 | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `$NUMBER`        | The pull request number. Example: `42`.                                                                                                                                                                                                                                                    |
+| `$NUMBER`        | The pull request number. Example: `42`. For a [merged entry](#group-changes), the number of the newest pull request.                                                                                                                                                                       |
```
cchanche:
```suggestion
| `$NUMBER`        | The pull request number. Example: `42`. For a [groupped entry](#group-changes), the number of the newest pull request.                                                                                                                                                                       |
```

README.md
```diff
@@ -320,7 +321,8 @@ Use these variables in `change-template`:
...
+| `$NUMBERS`       | Every pull request number of the entry, prefixed with `#` and separated with `, `, oldest first. Example: `#42`, or `#308, #310, #316` for a [merged entry](#group-changes).                                                                                                               |
```
cchanche:
```suggestion
| `$NUMBERS`       | Every pull request number of the entry, prefixed with `#` and separated with `, `, oldest first. Example: `#42`, or `#308, #310, #316` for a [groupped entry](#group-changes).                                                                                                               |
```

cchanche (changes requested):
Mostly small documentation rephrases to avoid confusion between merging PRs in github and merging changes in the release body (grouping)

LGTM !

README.md
slawekjaranowski: Applied — with `grouped`, one p: the suggestions spell it "groupped", which reads like a typo, so I took the intent and used `grouped` everywhere. Same for the `$NUMBER` / `$NUMBERS` rows and for the `toChanges` rename, which is now `toGroupedChanges`.

I also went through the rest of the feature and replaced "merge" with "group" in the option table, the `group-changes` schema description and the JSDoc of `group-changes.ts` (`mergeTitle` is now `groupTitle`), so "merge" only ever refers to merging a pull request. The one place it stays is `by merge date` / `byMergeOrder`, where it really is the GitHub merge.

README.md
slawekjaranowski: Yes, exactly that: the key is the combination of the `group` value and every `group_<name>` value, so `(lib, /module-a)`, `(lib, /module-b)`, `(other-lib, /module-a)` are three entries. A pull request that matches no rule stays an entry of its own.

The README did not say it outright, so it does now, right after the `group_in` example:

> An entry is one combination of the `group` value and every `group_<name>` value, so `lib` in `/module-a` and `lib` in `/module-b` stay two entries.

README.md
slawekjaranowski: You are right that the paragraph was confusing, so I rewrote it — but not quite as suggested, because `pattern` there belongs to `group-changes`, not to `replacers`, and it never sees the changelog body: it is matched against the original pull request titles, before any entry is rendered. A `replacer` is the one that runs last, on the finished body, and therefore sees the title that `title-template` built.

The paragraph mixed those two subjects in one sentence, which is where the confusion came from. It now separates them:

> `collapse-after` counts entries, so grouped changes count as one. Release Drafter matches `pattern` against the original pull request titles, before the changelog body exists. `replacers` run last, on the finished body, so their `search` sees the title that `title-template` built, not the original ones.

And yes — a finished entry is already the grouped one, which is exactly why `collapse-after` counts it as a single change.

packages/core/src/config/parse-group-changes.ts
slawekjaranowski: Not intended, good catch. `groupTitle` uppercases every capture name when it builds the template variables, so `(?<Group>…)` would have been rendered as `$GROUP` while silently not building the grouping key — it would have landed in `captureNames` as `$FIRST_GROUP` instead. The two checks now agree:

```ts
const isGroupName = (name: string) => /^group(_|$)/i.test(name)
```

Covered by a new test, `accepts group capture names in any case`, which asserts that `(?<Group>…)` and `(?<GROUP_IN>…)` end up in `groupNames`. The documented spelling stays lowercase; this only stops a differently-cased name from failing quietly.

README.md
cchanche: Oh yes that's a typo, thx for the follow up

cchanche (approved):

